Issue #15, opened 2021-08-05
Create a `TypeCache` for resolving types


desaikd:
Implement a struct called `TypeCache` which can be used by `Resolver` for resolving the `Type`. 
```rust
type TypeId = usize;

struct TypeCache {
    by_name: HashMap<String, TypeId>,
    by_id: Vec<TypeDefinition>,
}

impl TypeCache {
    fn new() -> Self;
    fn get_type_by_name(&self, name: &str) -> Option<&TypeDefinition>;
    fn get_type_by_id(&self, id: TypeId) -> Option<&TypeDefinition>;
    fn add_named_type(&mut self, name: &str, type_def: TypeDefinition) -> TypeId;
    fn add_anonymous_type(&mut self, type_def: TypeDefinition) -> TypeId;
}
```

desaikd:
`TypeCache` is renamed with a more appropriate name `TypeStore` in the implementation.

PR #19, opened 2021-08-27
adds implementation of TypeCache

desaikd:
*Issue #15*

*Description of changes:*
This PR works on implementation of  `TypeCache`.

*Change:*
- creates a struct `TypeCache` which helps with the cycling issue as described in #15.
- creates a type `TypeId` which represents the index/id for Types stores in `TypeCache`.
- modifies the `Constraint`s to store `Vec<TypeId>` instead of `Vec<Type>`.
- modifies `Scehma` struct to store the `TypeCache` that is created while loading the schema instead of a `HashMap<String, Type>`. (This `TypeCache` will help in resolving the `TypeId`s while validation)
- modifies get_types() in Schema as it needs to return an Iterator and not the `HashMap`/ `TypeCache` stored inside `Schema`.(I will work on this on next PR)

### DIFF
--- a/src/constraint.rs
+++ b/src/constraint.rs
@@ -1,6 +1,6 @@
 use crate::result::{invalid_schema_error_raw, IonSchemaResult};
-use crate::system::SharedTypeCache;
-use crate::types::{Type, TypeRef};
+use crate::system::{SharedTypeCache, TypeId};
+use crate::types::TypeRef;
 use crate::violation::Violations;
 use ion_rs::value::owned::OwnedElement;
 use ion_rs::value::{Element, Sequence};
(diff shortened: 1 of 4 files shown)


zslayton:
Thanks for your feedback, @popematt!

Per our offline discussion, `TypeId` is a hidden implementation detail. Passing around indexes into a `Vec` is a common Rust idiom for representing cyclic graph structures in memory (see [the `petgraph` crate](https://docs.rs/petgraph/0.6.0/petgraph/index.html) for a popular library leveraging this approach).

The types in this PR are intended to constitute the normalized, in-memory representation of a schema, its imports, and their types. We have not yet designed the API for constructing types/schemas programmatically; I expect it to use distinct public-facing types that get normalized into this PR's types via a `SchemaSystem`. This design work is our next priority.

Issue #18, opened 2021-08-26
`Schema#get_types()` should return an Iterator of `Type`s


desaikd:
As per #5, `Schema#get_types()` returns the `HashMap` which is used under the hood for storing `Type`s for a `Schema`. Returning this `HashMap` exposes the implementation detail of `Schema`, we should rather return an `Iterator<Item=Type>` for this.

PR #21, opened 2021-08-31
adds SchemaTypeIterator implementation

desaikd:
*Issue #18*

*Description of changes:*
This PR works on implementation of `SchemaTypeIterator` which can be used by `Schema` to return associated `Type`s.

*Changes:*
- adds a new struct `SchemaTypeIterator` and its `Iterator` implementation.
- adds a new method in `TypeStore` to return all the `Type`s stored inside it.
- changed `get_types` signature to return `SchemaTypeIterator` instead of `HashMap`.

### DIFF
--- a/src/schema.rs
+++ b/src/schema.rs
@@ -1,8 +1,8 @@
 use crate::import::Import;
 use crate::system::TypeStore;
 use crate::types::Type;
-use std::collections::HashMap;
 use std::rc::Rc;
+use std::slice::Iter;
 
 /// A Schema is a collection of zero or more [Type]s.
 ///
@@ -50,10 +50,11 @@ impl Schema {
         self.types.get_type_by_name(name.as_ref())
     }
 
-    //TODO: change get_types() return type to SchemaTypeIterator and define SchemaTypeIterator
     /// Returns an iterator over the types in this schema.
-    pub(crate) fn get_types(&self) -> &HashMap<String, Type> {
-        todo!()
+    pub(crate) fn get_types(&self) -> SchemaTypeIterator {
+        SchemaTypeIterator {
+            types: self.types.get_types().iter(),
+        }
     }
 
     /// Returns a new [Schema] instance containing all the types of this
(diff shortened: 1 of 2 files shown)


src/schema.rs
```diff
@@ -64,3 +65,16 @@ impl Schema {
         todo!()
     }
 }
+
+/// Provides an Iterator which returns [Type]s inside a [Schema]
+pub struct SchemaTypeIterator<'a> {
```
popematt, line 70: I have a couple of questions: 
- This looks like `SchemaTypeIterator` is just a wrapper around an `Iter`. Can you explain to me why we need to have a wrapper?
- Is it considered idiomatic to use non-descriptive lifetime identifiers like `'a`?

desaikd: We are just trying to hide the usage of `TypeStore` into a `SchemaTypeIterator` rather than returning an `Iter` and yes, we can use non-descriptive lifetime here which is used with `Iter` as well as `Type` reference. Does this answer your question?

popematt (approved):

src/schema.rs
```diff
@@ -50,10 +50,11 @@ impl Schema {
         self.types.get_type_by_name(name.as_ref())
     }
 
-    //TODO: change get_types() return type to SchemaTypeIterator and define SchemaTypeIterator
     /// Returns an iterator over the types in this schema.
-    pub(crate) fn get_types(&self) -> &HashMap<String, Type> {
-        todo!()
+    pub(crate) fn get_types(&self) -> SchemaTypeIterator {
+        SchemaTypeIterator {
+            types: self.types.get_types().iter(),
```
zslayton, line 56: I think this is ok for now since this isn't the final implementation, but I'm realizing that this will iterate over _all_ the types in the typestore. Since this is a method on `Schema`, this should only expose the types that are part of this schema.

zslayton (approved):
One item below that we need to track going forward.

src/schema.rs
desaikd: Reference issue: https://github.com/amzn/ion-schema-rust/issues/22

PR #24, opened 2021-09-07
adds changes for constructing types/constraints programmatically

### LINKED ISSUE #14: Implement `TypeDefinition` and `Type`
Current implementation (#2), has a struct `Type` which consists of an optional name and zero or more constraints. This should rather be renamed as `TypeDefinition` and create another struct `Type` which is defined as below:
```rust
/// Represents a [Type] with `TypeId` and reference to [Resolver]
/// Provides a way to get the corresponding [TypeDefinition] during validation
pub struct Type {
    id: TypeId, // TypeId is a usize to represent the id/index for a type
    resolver: Rc<Resolver>,
}
```
This `Type` struct will help in validation to get the `TypeDefinition` from `Resolver`. 
Reference issue for `Resolver` definition: #7

### LINKED ISSUE #23: Implementation of public facing API for constructing types/constraints programmatically
The issue is current implementation (https://github.com/amzn/ion-schema-rust/pull/19) of `ion-schema-rust` does not have any API designed for constructing types/schemas programmatically. Hence there is a need to create a public facing `TypeDefinition`, `Constraint`, `TypeRef` which then can be converted to a normalized, in-memory representation of `TypeDefinition` and `Constraints` (which uses the hidden implementation of `TypeStore`)

*Approach:*
For loading a schema into schema system there can be two phases in implementation as below:

1. Create all `IslType`,  `IslConstraint`, `IslTypeRef` from the given ion content (All "Isl" prefixed implementation)
2. load all types into Schema
    1. Convert all `IslType` → `TypeDefinition`, `IslConstraint` → `Constraint`, `IslTypeRef` → `TypeDefinition`
    2. While doing (i) store all `TypeDefinition` in the `TypeStore`(which could help returning resolved types in a schema) and store generated `TypeId` in the `Constraint`.
    3. For self referencing types `TypeStore` now will have a different parameter called parent which is a tuple storing the name and `TypeId` of a parent. This can be used when an alias type is not there in  the `TypeStore` but it’s name matches with the name of the parent (meaning a self referencing type) hence the parent’s `TypeId` is saved for the self referencing type.
 
*The implementation would look something like below:*
```rust
/// Represents a schema type which can be converted to a solid TypeRef using TypeStore
struct IslType {
    name: String,
    constraints: Vec<IslConstraint> 
}

/// Represents Constraint which stores IslTypeRef
enum IslConstraint { // this represents an AllOf constraint
    AllOf(Vec<TypeRef>),
    Type(TypeRef)
}

enum IslTypeRef {
   Alias(&str),
   Anonymous(IslType),
   CoreType(IonType)
}

/// Provides a public facing schema type which has a reference to TypeStore
/// to return an underlying TypeDefinition from TypeStore
struct TypeRef {
    id: TypeId,
    type_store: Rc<TypeStore>
}
```

desaikd:
*Description of changes:*
This PR works on changes for implementation of a public facing API for constructing types/constraints programmatically.

*Changes:*
- Adds implementation of new public facing API `IslType`, `IslConstraint`, `IslTypeRef` (stored in `isl` module)
- Renames `Type` to `TypeDefinition`
- Renames current `TypeRef` to `IslTypeRef`
- Creates a `TypeRef` which has a `TypeId` and reference to `TypeStore`
- Bug fix for self referencing types using parent information stored in `TypeStore`
- `TypeDefinition` field name is changed to be an `Option` to store both anonymous and named types
- doc comment changes

*Fixes #23*
*Fixes #14*

### DIFF
--- a/src/constraint.rs
+++ b/src/constraint.rs
@@ -1,6 +1,6 @@
 use crate::isl::IslTypeRef;
 use crate::result::IonSchemaResult;
-use crate::system::{SharedTypeStore, TypeId};
+use crate::system::{SharedContext, SharedTypeStore, TypeId};
 use crate::violation::Violations;
 use ion_rs::value::owned::OwnedElement;
 
@@ -13,7 +13,7 @@ pub trait ConstraintValidator {
 }
 
 /// Defines schema Constraints
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 // TODO: add other constraints
 pub enum Constraint {
     AllOf(AllOfConstraint),
@@ -36,10 +36,11 @@ impl AllOfConstraint {
     pub fn resolve_from_isl_constraint(
         type_references: &[IslTypeRef],
         type_store: &SharedTypeStore,
+        context: &SharedContext,
     ) -> IonSchemaResult<Self> {
         let resolved_types: Vec<TypeId> = type_references
             .iter()
-            .map(|t| IslTypeRef::resolve_type_reference(t, type_store))
+            .map(|t| IslTypeRef::resolve_type_reference(t, type_store, context))
             .collect::<IonSchemaResult<Vec<TypeId>>>()?;
         Ok(AllOfConstraint::new(resolved_types))
     }
@@ -48,6 +49,12 @@ impl AllOfConstraint {
 impl ConstraintValidator for AllOfConstraint {
     fn validate(&self, value: OwnedElement, issues: &mut Violations) {
         todo!()
+    }
+}
+
+impl PartialEq for AllOfConstraint {
+    fn eq(&self, other: &Self) -> bool {
+        self.type_ids == other.type_ids
     }
 }
 
@@ -67,8 +74,9 @@ impl TypeConstraint {
     pub fn resolve_from_isl_constraint(
         type_reference: &IslTypeRef,
         type_store: &SharedTypeStore,
+        context: &SharedContext,
     ) -> IonSchemaResult<Self> {
-        let type_id = IslTypeRef::resolve_type_reference(type_reference, type_store)?;
+        let type_id = IslTypeRef::resolve_type_reference(type_reference, type_store, context)?;
         Ok(TypeConstraint::new(type_id))
     }
 }
@@ -76,5 +84,11 @@ impl TypeConstraint {
 impl ConstraintValidator for TypeConstraint {
     fn validate(&self, value: OwnedElement, issues: &mut Violations) {
         todo!()
+    }
+}
+
+impl PartialEq for TypeConstraint {
+    fn eq(&self, other: &Self) -> bool {
+        self.type_id == other.type_id
     }
 }
--- a/src/constraint.rs
+++ b/src/constraint.rs
@@ -1,6 +1,6 @@
 use crate::isl::IslTypeRef;
 use crate::result::IonSchemaResult;
-use crate::system::{SharedContext, SharedTypeStore, TypeId};
+use crate::system::{SharedPendingTypes, SharedTypeStore, TypeId};
 use crate::violation::Violations;
 use ion_rs::value::owned::OwnedElement;
 
@@ -36,7 +36,7 @@ impl AllOfConstraint {
     pub fn resolve_from_isl_constraint(
         type_references: &[IslTypeRef],
         type_store: &SharedTypeStore,
-        context: &SharedContext,
+        context: &SharedPendingTypes,
     ) -> IonSchemaResult<Self> {
         let resolved_types: Vec<TypeId> = type_references
             .iter()
@@ -74,7 +74,7 @@ impl TypeConstraint {
     pub fn resolve_from_isl_constraint(
         type_reference: &IslTypeRef,
         type_store: &SharedTypeStore,
-        context: &SharedContext,
+        context: &SharedPendingTypes,
     ) -> IonSchemaResult<Self> {
         let type_id = IslTypeRef::resolve_type_reference(type_reference, type_store, context)?;
         Ok(TypeConstraint::new(type_id))
--- a/src/constraint.rs
+++ b/src/constraint.rs
@@ -33,7 +33,7 @@ impl AllOfConstraint {
     }
 
     /// Tries to create an [AllOf] constraint from the given OwnedElement
-    pub fn parse_from_isl_constraint(
+    pub fn resolve_from_isl_constraint(
         type_references: &[IslTypeRef],
         type_store: &SharedTypeStore,
     ) -> IonSchemaResult<Self> {
@@ -41,7 +41,7 @@ impl AllOfConstraint {
             .iter()
             .map(|t| IslTypeRef::resolve_type_reference(t, type_store))
             .collect::<IonSchemaResult<Vec<TypeId>>>()?;
-        Ok(AllOfConstraint::new(resolved_types.to_owned()))
+        Ok(AllOfConstraint::new(resolved_types))
     }
 }
 
@@ -64,7 +64,7 @@ impl TypeConstraint {
     }
 
     /// Tries to create a [Type] constraint from the given OwnedElement
-    pub fn parse_from_isl_constraint(
+    pub fn resolve_from_isl_constraint(
         type_reference: &IslTypeRef,
         type_store: &SharedTypeStore,
     ) -> IonSchemaResult<Self> {
--- a/src/constraint.rs
+++ b/src/constraint.rs
@@ -1,10 +1,8 @@
-use crate::result::{invalid_schema_error_raw, IonSchemaResult};
+use crate::isl::IslTypeRef;
+use crate::result::IonSchemaResult;
 use crate::system::{SharedTypeStore, TypeId};
-use crate::types::TypeRef;
 use crate::violation::Violations;
 use ion_rs::value::owned::OwnedElement;
-use ion_rs::value::{Element, Sequence};
-use ion_rs::IonType;
 
 /// Provides validation for schema Constraint
 pub trait ConstraintValidator {
@@ -35,26 +33,13 @@ impl AllOfConstraint {
     }
 
     /// Tries to create an [AllOf] constraint from the given OwnedElement
-    pub fn parse_from_ion_element(
-        ion: &OwnedElement,
+    pub fn parse_from_isl_constraint(
+        type_references: &[IslTypeRef],
         type_store: &SharedTypeStore,
     ) -> IonSchemaResult<Self> {
-        if ion.ion_type() != IonType::List {
-            return Err(invalid_schema_error_raw(format!(
-                "all_of constraint was a {:?} instead of a list",
-                ion.ion_type()
-            )));
-        }
-        let types: Vec<TypeRef> = ion
-            .as_sequence()
-            .unwrap()
+        let resolved_types: Vec<TypeId> = type_references
             .iter()
-            .map(|e| TypeRef::parse_from_ion_element(e, type_store))
-            .collect::<IonSchemaResult<Vec<TypeRef>>>()?;
-
-        let resolved_types: Vec<TypeId> = types
-            .iter()
-            .map(|t| TypeRef::resolve_type_reference(t, type_store))
+            .map(|t| IslTypeRef::resolve_type_reference(t, type_store))
             .collect::<IonSchemaResult<Vec<TypeId>>>()?;
         Ok(AllOfConstraint::new(resolved_types.to_owned()))
     }
@@ -79,18 +64,11 @@ impl TypeConstraint {
     }
 
     /// Tries to create a [Type] constraint from the given OwnedElement
-    pub fn parse_from_ion_element(
-        ion: &OwnedElement,
+    pub fn parse_from_isl_constraint(
+        type_reference: &IslTypeRef,
         type_store: &SharedTypeStore,
     ) -> IonSchemaResult<Self> {
-        if ion.ion_type() != IonType::Symbol && ion.ion_type() != IonType::Struct {
-            return Err(invalid_schema_error_raw(format!(
-                "type constraint was a {:?} instead of a symbol/struct",
-                ion.ion_type()
-            )));
-        }
-        let type_reference: TypeRef = TypeRef::parse_from_ion_element(ion, type_store)?;
-        let type_id = TypeRef::resolve_type_reference(&type_reference, type_store)?;
+        let type_id = IslTypeRef::resolve_type_reference(type_reference, type_store)?;
         Ok(TypeConstraint::new(type_id))
     }
 }
--- a/src/isl.rs
+++ b/src/isl.rs
@@ -1,9 +1,10 @@
 use crate::result::{
-    invalid_schema_error_raw, unresolvable_schema_error, IonSchemaError, IonSchemaResult,
+    invalid_schema_error, invalid_schema_error_raw, unresolvable_schema_error, IonSchemaError,
+    IonSchemaResult,
 };
-use crate::system::{SharedContext, SharedTypeStore, TypeId};
+use crate::system::{SharedPendingTypes, SharedTypeStore, TypeId};
 use crate::types::TypeDefinition;
-use ion_rs::value::owned::{OwnedElement, OwnedStruct};
+use ion_rs::value::owned::OwnedElement;
 use ion_rs::value::{Element, Sequence, Struct, SymbolToken};
 use ion_rs::IonType;
 use std::convert::{TryFrom, TryInto};
@@ -46,11 +47,13 @@ impl PartialEq for IslType {
 }
 
 /// Parse constraints inside an [OwnedStruct] to an [IslType]
-impl TryFrom<&OwnedStruct> for IslType {
+impl TryFrom<&OwnedElement> for IslType {
     type Error = IonSchemaError;
 
-    fn try_from(ion_struct: &OwnedStruct) -> Result<Self, Self::Error> {
+    fn try_from(ion: &OwnedElement) -> Result<Self, Self::Error> {
         let mut constraints = vec![];
+
+        let ion_struct = try_to!(ion.as_struct());
 
         // parses the name of the type specified by schema
         let type_name: Option<String> = match ion_struct.get("name") {
@@ -169,7 +172,7 @@ impl IslTypeRef {
                     })
             }
             IonType::Struct =>
-                Ok(IslTypeRef::AnonymousType(value.as_struct().unwrap().try_into()?)),
+                Ok(IslTypeRef::AnonymousType(value.try_into()?)),
             _ => Err(invalid_schema_error_raw(
                 "type reference can either be a symbol(For base/alias type reference) or a struct (for anonymous type reference)",
             )),
@@ -181,7 +184,7 @@ impl IslTypeRef {
     pub fn resolve_type_reference(
         type_reference: &IslTypeRef,
         type_store: &SharedTypeStore,
-        context: &SharedContext,
+        context: &SharedPendingTypes,
     ) -> IonSchemaResult<TypeId> {
         match type_reference {
             IslTypeRef::CoreIslType(ion_type) => {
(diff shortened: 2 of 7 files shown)


src/isl.rs
```diff
@@ -0,0 +1,211 @@
...
+    type Error = IonSchemaError;
+
+    fn try_from(ion_struct: &OwnedStruct) -> Result<Self, Self::Error> {
+        let mut constraints = vec![];
+
+        // parses the name of the type specified by schema
+        let type_name: Option<String>=  match ion_struct.get("name") {
+            Some(name_element) => match name_element.as_str() {
+                Some(name) => Some(name.to_owned()),
+                None => { return Err(invalid_schema_error_raw("A type name is not string/symbol, if the value is any null, or the text of the symbol is not defined.")) }
```
zslayton, line 50: I'd simplify this to `type names must be a string or a symbol with defined text`. It would be even friendlier to break this into different error cases for `wrong type`, `null`, and `unknown text`, but we can worry about that later.

src/isl.rs
```diff
@@ -0,0 +1,211 @@
...
+            let constraint_name = match field_name.text() {
+                Some("name") => continue, // if the field_name is "name" then it's the type name not a constraint
+                Some(name) => name,
+                None => {
+                    return Err(invalid_schema_error_raw(
+                        "A type name symbol token does not have any text",
+                    ))
+                }
+            };
+            // TODO: add more constraints to match below
```
zslayton, line 66: For a future PR: I think we'll want to break these `match` arms into helper methods as we add more. Even with just two constraint types it's pretty long.

src/isl.rs
```diff
@@ -0,0 +1,211 @@
...
+                            + &type_name.unwrap()
+                            + " can not be built as constraint: "
+                            + constraint_name
+                            + " does not exist",
+                    ))
+                }
+            };
+            constraints.push(constraint);
+        }
+        Ok(IslType::new(type_name.to_owned(), constraints))
```
zslayton, line 105: I think this `to_owned()` is doing an unnecessary clone?

src/isl.rs
```diff
@@ -0,0 +1,211 @@
...
+    }
+}
+
+/// Provides an internal representation of schema type reference.
+/// Type reference grammar is defined in [Ion Schema Spec]
+/// [Ion Schema spec]: https://amzn.github.io/ion-schema/docs/spec.html#grammar
```
zslayton, line 111: Nice, thanks for adding the link. 👍 

```suggestion
/// Provides an internal representation of a schema type reference.
/// The type reference grammar is defined in the [Ion Schema Spec]
/// [Ion Schema spec]: https://amzn.github.io/ion-schema/docs/spec.html#grammar
```

src/isl.rs
```diff
@@ -0,0 +1,211 @@
...
+#[derive(Debug, Clone)]
+pub enum IslTypeRef {
+    /// represents core ion type reference
```
zslayton, line 114:
```suggestion
    /// Represents a reference to one of ISL's built-in core types:
    /// https://amzn.github.io/ion-schema/docs/spec.html#core-types
```

src/isl.rs
```diff
@@ -0,0 +1,211 @@
...
+    IslCoreType(IonType),
+    /// represents a type reference which represents a type imported from another schema
```
zslayton, line 116: Consider changing this to `NamedType` since these won't always be an alias, per se.

```suggestion
    /// represents a reference to a named type (including aliases)
```

src/isl.rs
```diff
@@ -0,0 +1,211 @@
...
+        match value.ion_type() {
+            IonType::Symbol => {
+                value.as_sym().unwrap()
+                    .text()
+                    .ok_or_else(|| {
+                        invalid_schema_error_raw(
+                            "a base or alias type reference symbol doesn't have text",
+                        )
+                    })
+                    .and_then(|type_reference| {
```
zslayton, line 137: We're in the process of creating a type reference, but I think this particular variable is a type name.
```suggestion
                    .and_then(|type_name| {
```

src/isl.rs
```diff
@@ -0,0 +1,211 @@
...
+    /// Resolves a type_reference into a [TypeId] using the type_store
+    pub fn resolve_type_reference(
+        type_reference: &IslTypeRef,
+        type_store: &SharedTypeStore,
+    ) -> IonSchemaResult<TypeId> {
+        match type_reference {
+            IslTypeRef::IslCoreType(ion_type) => {
+                // TODO: create CoreType struct for storing ISLCoreType type definition instead of Type
+                // inserts ISLCoreType as a Type into type_store
+                Ok(type_store.borrow_mut().add_named_type(
```
zslayton, line 174: TODO for the future: `TypeStore` should be responsible for calling `borrow_mut()` internally; making users call it is a leaked implementation detail.

src/isl.rs
```diff
@@ -0,0 +1,211 @@
...
+                let borrow_type_store = type_store.borrow_mut();
+                // verify if the AliasType actually exists in the type_store or throw an error
+                match borrow_type_store.get_type_id_by_name(alias) {
+                    Some(type_id) => Ok(type_id.to_owned()),
+                    None => match borrow_type_store.get_parent() {
+                        Some(parent) => {
+                            // if it is a self referencing type resolve it using parent information from type_store
+                            if parent.0.eq(alias) {
+                                Ok(parent.1)
+                            } else {
```
zslayton, line 189: TODO: We'll need to generalize this to work at any level of nesting, not just one.

src/schema.rs
```diff
@@ -46,7 +46,7 @@ impl Schema {
 
     /// Returns the requested type, if present in this schema;
     /// otherwise returns None.
-    fn get_type<A: AsRef<str>>(&self, name: A) -> Option<&Type> {
+    fn get_type<A: AsRef<str>>(&self, name: A) -> Option<&TypeDefinition> {
```
zslayton, line 49: Is there a good reason to hand out a `&TypeDefinition` instead of a `TypeRef`? (Same question for all of the below method signatures in `Schema` too.)

src/types.rs
```diff
@@ -18,21 +18,28 @@ pub trait TypeValidator {
     fn validate(&self, value: &OwnedElement, issues: &mut Violations);
 }
 
+/// Provides a public facing schema type which has a reference to TypeStore
+/// to get the underlying TypeDefinition from TypeStore
```
zslayton, line 22: This is the public-facing documentation, so it shouldn't describe implementation details. This would make a good developer comment, though. (`//` vs `///`)

src/types.rs
```diff
@@ -18,21 +18,28 @@ pub trait TypeValidator {
...
+    type_store: Rc<TypeStore>,
+}
+
 /// A Type consists of an optional name and zero or more constraints.
 ///
 /// Unless otherwise specified, the constraint `type: any` is automatically applied.
 #[derive(Debug, Clone)]
-pub struct Type {
-    name: String,
+pub struct TypeDefinition {
```
zslayton, line 32: This should be either private or, if necessary, `pub(crate)`.

zslayton (commented):
Looking good! Lots of small cleanup items and TODOs for future PRs below.

src/isl.rs
```diff
@@ -0,0 +1,211 @@
...
+
+/// Represents a public facing schema type [IslType] which can be converted to a solid TypeRef using TypeStore
+#[derive(Debug, Clone)]
+pub struct IslType {
+    name: Option<String>,
+    constraints: Vec<IslConstraint>,
+}
+
+impl IslType {
+    pub fn new(name: Option<String>, constraints: Vec<IslConstraint>) -> Self {
```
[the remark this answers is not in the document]
popematt: +1

...but we may even want to consider breaking named and anonymous IslTypes into different (Rust) types. (Or should I say different structs?) The reason is because that way we can use the Rust type system to help enforce some schema invariants—for example, a schema cannot have a top-level anonymous type, and a nested/inline type cannot have a name. It seems like that would make it easier for users because they would get a compile time error rather than a runtime error if they use the wrong one.

Just to be clear, this is a "have you considered?" comment, not a "you must" sort of comment.

src/isl.rs
```diff
@@ -0,0 +1,211 @@
...
+    }
+}
+
+/// Provides an internal representation of schema type reference.
+/// Type reference grammar is defined in [Ion Schema Spec]
+/// [Ion Schema spec]: https://amzn.github.io/ion-schema/docs/spec.html#grammar
+#[derive(Debug, Clone)]
+pub enum IslTypeRef {
+    /// represents core ion type reference
+    IslCoreType(IonType),
```
popematt, line 115: question – Are we prefixing all (Rust) type names for ISL with `Isl`, or just the top level ones? In other words, should the prefix be here and/or should it be added to the other members of this enum?

src/isl.rs
```diff
@@ -0,0 +1,211 @@
...
+            };
+            // TODO: add more constraints to match below
+            let constraint = match constraint_name {
+                "all_of" => {
+                    if value.ion_type() != IonType::List {
+                        return Err(invalid_schema_error_raw(format!(
+                            "all_of constraint was a {:?} instead of a list",
+                            value.ion_type()
+                        )));
+                    }
```
popematt, line 74: question (non-blocking) – Is there a good Rust idiom to DRY some of this IonType checking? I know this only happens a few times in this file right now, but as you implement more of these, you'll find it popping up a lot.

Could we come up with something like this maybe?
```rust
assert_constraint_param_type!("all_of", value, IonType::List)
```

src/isl.rs
```diff
@@ -0,0 +1,211 @@
...
+#[derive(Debug, Clone)]
+pub enum IslTypeRef {
+    /// represents core ion type reference
+    IslCoreType(IonType),
+    /// represents a type reference which represents a type imported from another schema
+    AliasType(String),
+    /// represents a type reference defined as an inlined import type from another schema
+    // TODO: add ImportType(Import) where ImportType could either point to a schema represented by an id with all the types or a single type from inside it
+    /// represents an unnamed type definition reference
+    AnonymousType(IslType),
```
popematt, line 121: Okay, now I see that you do have an AnonymousType.

question – Rather than using the `AnonymousType` as a wrapper for `IslType` (which has an optional name), do you think it would make sense to make `AnonymousType` just have its own list of constraints, and use a trait or something for what `AnonymousType` and `IslType` have in common (the getter for the constraints, I guess)?

popematt (commented):
TODO for later (but not too much later). Can you document the naming conventions for the Isl models, the user-facing models, and the internal-only models and commit it somewhere in the repo? I'm already starting to forget which is which. (I'm sorry!)

src/isl.rs
```diff
@@ -0,0 +1,211 @@
...
+use std::convert::{TryFrom, TryInto};
+
+/// Represents a public facing API for schema constraints [IslConstraint] which stores IslTypeRef
+#[derive(Debug, Clone)]
+pub enum IslConstraint {
+    AllOf(Vec<IslTypeRef>),
+    Type(IslTypeRef),
+}
+
+/// Represents a public facing schema type [IslType] which can be converted to a solid TypeRef using TypeStore
```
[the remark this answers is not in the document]
desaikd: Reference issue: #26

src/isl.rs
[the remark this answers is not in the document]
desaikd: I feel like there needs to be a common representation for all types(whether named or anonymous). Although it maybe required to divide these two types into different structs to enforce some of their specific schema rules. I have created an issue(#25) for this, currently I am going ahead with the suggestion to create different methods for them and later if this is required I can create a common `trait` `TypeDefinition` and its two implementations(structs) for named and anonymous types or even core types.

src/isl.rs
```diff
@@ -0,0 +1,211 @@
+use crate::result::{
+    invalid_schema_error_raw, unresolvable_schema_error, IonSchemaError, IonSchemaResult,
+};
+use crate::system::{SharedTypeStore, TypeId};
+use crate::types::TypeDefinition;
+use ion_rs::value::owned::{OwnedElement, OwnedStruct};
+use ion_rs::value::{Element, Sequence, Struct, SymbolToken};
+use ion_rs::IonType;
+use std::convert::{TryFrom, TryInto};
+
```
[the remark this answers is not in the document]
desaikd: Reference issue #26

src/isl.rs
desaikd: We are only adding `Isl` prefix to the top-level meaning to the given enum here not the members, unfortunately the naming I had for isl core types(not the ion type`$int` but `int`) confuses with the `IslTypeRef`, how about I rename it to `CoreType`?

src/isl.rs
desaikd: Did you mean if we could have  reusable implementation/method for these type checks? I can add a TODO for that and then as I go further with implementation I can use that everywhere.

src/isl.rs
popematt: > the naming I had for isl core types(not the ion type$int but int)

Thanks for the explanation. You've actually convinced me that having `Isl` in the name could be helpful. We could use `CoreIslType` (and later, `CoreIonType`) if you think that the distinction needs to be made, but `CoreType` is also fine.

src/isl.rs
popematt: Yes. Sorry that was unclear. (DRY == Don't Repeat Yourself; contrast with WET == Write Everything Twice)

A TODO is just fine.

src/isl.rs
desaikd: I think this is similar to having two different implementation of `TypeDefinition` but rather for the upper layer `IslType`. I will add it in the issue: #25.

src/isl.rs
desaikd: As per offline, current implementation will work fine as the named type can be defined only as a top level type, it should be sufficient to just have a parent which stores the named type's name and `TypeId` rather than storing a `Vec` of parent types for any level of nesting. The named type is going to be the only parent with a name and hence any self reference type can refer to this named type parent to resolve the type. I will also add more doc comments around the usage of parent in `TypeStore`.

zslayton:
As discussed offline, we should add some unit tests for the following conversions:
* `OwnedElement` to `Isl*` types
* `Isl*` types to a `Schema` (and `TypeRef`)
* `OwnedElement` to `Schema`/`TypeRef` (doesn't need to be exhaustive as its the sum of the above parts, but it would be good to demonstrate the API.)

src/isl.rs
```diff
@@ -201,11 +218,79 @@ impl IslTypeRef {
...
+    #[rstest(
+    isl_type1,isl_type2,
+    case::type_constraint_with_anonymous_type(
+        /* For a schema with single anonymous type as below: 
+            { type: int }
+         */
+        (&OwnedStruct::from_iter(vec![("type", OwnedElement::from(text_token("int")))].into_iter())).try_into().unwrap(),
+        IslType::new(None, vec![IslConstraint::Type(IslTypeRef::CoreIslType(IonType::Integer))])
+    ),
```
zslayton, line 248: These `OwnedElement` construction snippets get pretty dense. I suggest using `ElementReader` and writing out the Ion text instead.

```rust
use crate::value::reader::{element_reader, ElementReader};
// ...

// new helper function
fn load(text: &str) -> OwnedElement {
    element_reader()
        .read_one(text.as_bytes())
        .expect("parsing failed unexpectedly")
}

// ...
case::type_constraint_with_anonymous_type(
    load(r#" // Raw string, no escaping of special characters required
        {type: int}
    "#),
    IslType::new(None, vec![IslConstraint::Type(IslTypeRef::CoreIslType(IonType::Integer))])
),
```

src/schema.rs
```diff
@@ -52,9 +51,14 @@ impl Schema {
...
-            types: self.types.get_types().iter(),
-        }
+        SchemaTypeIterator::new(
+            self.types
+                .get_types()
+                .iter()
+                .enumerate()
+                .map(|(i, t)| TypeRef::new(i, Rc::clone(&self.types)))
+                .collect(),
+        )
```
zslayton, line 61: Iterators are usually lazy so the caller only pays for what they use. Allocating a `Vec` and building all of the `TypeRef`s when `get_types()` is called would be pretty expensive if the caller only wanted the first type in the schema.

Your original implementation was very close to what I had in mind; `SchemaTypeIterator` should have a lifetime linked to `&self` and, instead of passing out `&TypeDefinition`s, would construct a single `TypeRef` each time `next()` was called.

src/schema.rs
```diff
@@ -67,14 +71,115 @@ impl Schema {
...
+    use rstest::*;
+    use std::cell::RefCell;
+
+    #[rstest(
+    owned_elements, total_types,
+    case::type_constraint_with_anonymous_type(
+        /* For a schema with single anonymous type as below: 
+            type:: { type: int }
+         */
+        vec![OwnedElement::new_struct(vec![("type", OwnedElement::from(text_token("int")))].into_iter()).with_annotations(vec![text_token("type")])].into_iter(),
```
zslayton, line 112: A `load()` helper method would help streamline these tests too.

src/system.rs
```diff
@@ -13,6 +14,174 @@ use std::convert::TryInto;
 use std::io::ErrorKind;
 use std::rc::Rc;
 
+pub type SharedContext = Rc<RefCell<Context>>;
+
+/// Defines a [Context] struct which stores information on parent types
+/// while loading a [Schema] with types which is used to resolve self referencing schema types
+#[derive(Debug, Clone)]
+pub struct Context {
```
zslayton, line 22: `Context` is a reasonable name, but I think we can be more specific: `PendingTypes` or `ParentTypes` give a clearer sense of its role, IMO.

src/system.rs
```diff
@@ -13,6 +14,174 @@ use std::convert::TryInto;
 use std::io::ErrorKind;
 use std::rc::Rc;
 
+pub type SharedContext = Rc<RefCell<Context>>;
```
zslayton, line 17: IIRC, we originally needed to pass an `Rc<RefCell<Resolver>>` as a method parameter because the `Resolver` contained the list of authorities and calls to `resolve(...)` could recurse.

* Do we still need `Rc<RefCell<_>>` to pass the `Context` as a parameter? I don't think it suffers from this issue.
* Do we still need `Rc<RefCell<TypeStore>>` when passing the `TypeStore` (not `Resolver`) as a parameter in methods like `resolve_type_reference`?

These don't need to block the PR, but we should look into them.

src/system.rs
```diff
@@ -13,6 +14,174 @@ use std::convert::TryInto;
...
+
+/// Defines a [Context] struct which stores information on parent types
+/// while loading a [Schema] with types which is used to resolve self referencing schema types
+#[derive(Debug, Clone)]
+pub struct Context {
+    ids_by_name: HashMap<String, TypeId>,
+    parent: Option<(String, TypeId)>,
+    types_by_id: Vec<Option<TypeDefinition>>, // a None in this vector represents unresolved type
```
zslayton, line 25:
```suggestion
    types_by_id: Vec<Option<TypeDefinition>>, // a None in this vector represents a not-yet-resolved type
```

src/system.rs
```diff
@@ -13,6 +14,174 @@ use std::convert::TryInto;
...
+                    None => None,
+                },
+            })
+    }
+
+    /// Provides the [TypeId] associated with given name if it exists in the [TypeStore] or [Context]  
+    /// Otherwise returns None
+    pub fn get_type_id_by_name(&self, name: &str, type_store: &SharedTypeStore) -> Option<TypeId> {
+        match self.ids_by_name.get(name) {
+            Some(id) => Some(id.to_owned()),
```
zslayton, line 84: There are a lot of `to_owned()`s in this file that I think are unnecessary? `TypeId` is just a `usize`, so it can be freely copied everywhere.

src/system.rs
```diff
@@ -13,6 +14,174 @@ use std::convert::TryInto;
...
+    /// or in the [Context] Otherwise returns None
+    pub fn get_type_by_id(
+        &self,
+        id: TypeId,
+        type_store: &SharedTypeStore,
+    ) -> Option<TypeDefinition> {
+        match self.types_by_id.get(id) {
+            Some(t) => Some(t.to_owned().unwrap()),
+            None => match type_store.borrow().get_type_by_id(id) {
+                Some(t) => Some(t.to_owned()),
```
zslayton, line 102: I see a lot of `TypeDefinition` cloning happening in this file that I think we can eliminate in another PR.

I think making `Context` responsible for the fallback-to-`TypeStore` logic is muddying the waters a bit. IMO, `get_type_by_id` should only return the type if the `Context` has it. Callers can instead do:

```rust
  self.context
    .get_type_by_id(type_id)
    .or_else(|| type_store.borrow().get_type_by_id(type_id))
```

src/types.rs
```diff
@@ -103,3 +139,80 @@ impl TypeValidator for TypeDefinition {
...
+            { type: int }
+         */
+        IslType::new(None, vec![IslConstraint::Type(IslTypeRef::CoreIslType(IonType::Integer))]),
+        TypeDefinition::new(None, vec![Constraint::Type(TypeConstraint::new(1))])
+    ),
+    case::type_constraint_with_named_type(
+        /* For a schema with named type as below: 
+            { name: my_int, type: int }
+         */
+        IslType::new(Some("my_int".to_owned()), vec![IslConstraint::Type(IslTypeRef::CoreIslType(IonType::Integer))]),
```
zslayton, line 166: Same thought in these tests RE: `load()`

src/system.rs
desaikd: Sure. I think `PendingTypes` sounds like a good name for this.

src/schema.rs
desaikd: The issue here was when trying to convert `TypeDefinition` into `TypeRef` instead of collecting into vector, I canuse `.iter()` but that creates a temporary value created in this function hence can not return it. So I had to change it to rather be `Vec<TypeRef>`.  Did you have something particular in mind for this?

src/system.rs
desaikd: I have been using `Rc<RefCell<Context>` everywhere as I have to modify it while adding types and `Rc<Context>` won't be enough here. For `Rc<RefCell<TypeStore>>`, we maybe able to change it to `Rc<TypeStore>`. (I had used it in methods like `resolve_type_reference` just so that I don't have to unwrap it into `TypeStore`/` Rc<TypeStore>` in between rather that is done after everything is loaded into schema while passing to `Schema` constructor)

src/system.rs
desaikd: The issue here is that `Context` stores `TypeDefintion` in terms of `Option` i.e. `types_by_id: Vec<Option<TypeDefinition>>`, where as TypeStore stores TypeDefinition directly i.e. `types_by_id: Vec<TypeDefinition>`. Because of this I would have to return `Option<TypeDefintion>` here rather than `Option<&TypeDefinition>`.  I could make it concise by using `or_else` and `and_then` but `to_owned()` still needs to be used. Also, if I try to do something like below:
```rust
 self.context
    .get_type_by_id(type_id).and_then(|t| &t.unwrap())   // <---- this &__t.unwrap()__ creates a temporary value that can not be returned from this fucntion.
    .or_else(|| type_store.borrow().get_type_by_id(type_id)
```

src/system.rs
zslayton: But can't `Rc<RefCell<Context>>` just be `&mut Context`?

src/system.rs
zslayton: Why does the `Context` ever need to _give out_ a `TypeDefinition`? I thought it just accumulated `TypeDefinition`s and then moved them all into the `TypeStore` at the end of the resolution process.

desaikd: I am using the `Context` while resolving a type inside `resolve_type_reference` where I refer to the `types_by_id` vector inside the `Context` to check if a resolved `TypeDefinition` already exists.

desaikd: Upon having another look at the method and from offline discussion the method isn't really required to be here inside `Context` struct hence will be removing this method.

src/system.rs
desaikd: added a TODO for this in the implementation

src/schema.rs
desaikd: Changes `SchemaTypeIterator` to store `Rc<TypeStore> `and create `TypeRef` when `.next()` is called.

src/constraint.rs
```diff
@@ -36,7 +36,7 @@ impl AllOfConstraint {
     pub fn resolve_from_isl_constraint(
         type_references: &[IslTypeRef],
         type_store: &SharedTypeStore,
-        context: &SharedContext,
+        context: &SharedPendingTypes,
```
zslayton, line 39: Per our conversation you're planning to simplify this to `&mut PendingTypes` in a future PR. If you don't already have one, turn this comment into an issue to track it.

src/isl.rs
```diff
@@ -232,61 +235,61 @@ impl IslTypeRef {
...
     #[rstest(
     isl_type1,isl_type2,
     case::type_constraint_with_anonymous_type(
-        /* For a schema with single anonymous type as below: 
-            { type: int }
-         */
-        (&OwnedStruct::from_iter(vec![("type", OwnedElement::from(text_token("int")))].into_iter())).try_into().unwrap(),
+        (&load(r#" // For a schema with single anonymous type
+            {type: int}
+        "#)).try_into().unwrap(),
```
zslayton, line 255: You can make another helper method that also takes care of the `try_into().unwrap()` on each of these.

zslayton (approved):

src/isl.rs
```diff
@@ -0,0 +1,299 @@
...
+    isl_type1,isl_type2,
+    case::type_constraint_with_anonymous_type(
+        (&load(r#" // For a schema with single anonymous type
+            {type: int}
+        "#)).try_into().unwrap(),
+        IslType::new(None, vec![IslConstraint::Type(IslTypeRef::CoreIslType(IonType::Integer))])
+    ),
+    case::type_constraint_with_named_type(
+        (&load(r#" // For a schema with named type 
+            { name: my_int, type: int }
```
popematt, line 260: Minor nitpick: according to the spec, a named type is required to be annotated with `type::`. (Without that annotation, it's an inline type, and the `name` field is just "open content".) I don't know if it makes any functional difference in this test, but it's a good idea to make sure that the test inputs conform to the ISL spec (unless we're testing an unhappy case, of course).

src/schema.rs
```diff
@@ -46,35 +45,135 @@ impl Schema {
...
+    ),
+    case::type_constraint_with_named_type(
+        load(r#" For a schema with named type as below: 
+            type:: { name: my_int, type: int }
+        "#).into_iter(),
+        2 // this includes the core type int and the anonymous type
+    ),
+    case::type_constraint_with_self_reference_type(
+        load(r#" For a schema with self reference type as below: 
+            type:: { name: my_int, type: my_int }
```
popematt, line 126: Thought/nitpick—I've just realized that this particular case should actually be invalid ISL. It's logically impossible to resolve a type that defines itself in terms of `not`, `type`, `one_of`, `all_of`, or `any_of` itself. Self-referencing only makes sense in the scenarios where a thing is composed of other instances of the same type (eg. `element`, `ordered_element`, `fields`).

We can make a new issue to follow up on this later.

popematt (approved):

src/isl.rs
desaikd: Yes, there needs to be type:: annotation check for named type as this has been already checked in `schema_from_elements` method defined in `system` I haven't added a check here. Also, currently both anonymous/named types are defined by same struct `TypeDefinition`. But in a separate PR with solution #25 I will fix this by adding a check for named types specifically.

src/schema.rs
desaikd: I have opened an issue #27 for this which could be handled once the implementation of `element`, `fields`, `ordered_elements` is available.

src/constraint.rs
desaikd: Reference issue: #28

PR #29, opened 2021-09-17
adds changes to type aliases SharedPendingTypes and SharedTypeStore

### LINKED ISSUE #20: Replace usage of `SharedTypeStore` with `TypeStore`
As per https://github.com/amzn/ion-schema-rust/pull/19#discussion_r697560916, to remove usage of `SharedTypeStore` we will need to load a `Schema` into 2 phases: get all the owned elements using `Authority` and then convert them to `Type`s.

> `type_cache.borrow().to_owned()`
> 
> is cloning the entire TypeCache. However, I think we want to just move the existing TypeCache into the Schema we're creating. That means we should change this method to take TypeCache instead of &SharedTypeCache

Hence replacing `SharedTypeStore` to `TypeStore` would resolve the above mentioned issue of cloning.

### LINKED ISSUE #28: Change `SharedPedningTypes` to be defined as `&mut PendingTypes`
Currently the `SharedPendingTypes` is `Rc<RefCell<PedningTypes>>` which isn't really required. This can  now be changed to a more appropriate type `&mut PedningTypes`.

Reference: https://github.com/amzn/ion-schema-rust/pull/24#discussion_r710479435

desaikd:
*Issues:*
Fixes #28
Fixes #20

*Description of changes:*
Adds changes for  using `&mut PendingTypes` for `SharedPendingTypes `and `&mut TypeStore` for  `SharedTypeStore.` 

*Changes:*
- removes type alias SharedTypeStore and SharedPendingTypes
- changes all methods loading a schema to use &mut PendingTypes and &mut TypeStore instead.
- renames variables to use the name `pending_types` instead of `context`

### DIFF
--- a/src/system.rs
+++ b/src/system.rs
@@ -15,7 +15,7 @@ use std::rc::Rc;
 
 /// Stores information about types that are in the process of being defined.
 ///
-/// An ISL type definition can include types that are not yet fully defined. 
+/// An ISL type definition can include types that are not yet fully defined.
 /// For example, an ISL type definition might include:
 /// * A reference to itself. This could happen in a recursive structure like a
 ///   linked list or binary tree.
@@ -23,7 +23,7 @@ use std::rc::Rc;
 /// Because the [SchemaSystem] does not yet know the complete definition
 /// of these types, it cannot find them in the [TypeStore].
 /// An instance of [PendingTypes] is used to track information about types
-/// that we do not have a complete definition for yet. When the 
+/// that we do not have a complete definition for yet. When the
 /// [SchemaSystem] finishes loading these types, the type definitions in
 /// [PendingTypes] can be promoted the [TypeStore].
 #[derive(Debug, Clone)]
--- a/src/system.rs
+++ b/src/system.rs
@@ -13,8 +13,19 @@ use std::convert::TryInto;
 use std::io::ErrorKind;
 use std::rc::Rc;
 
-/// Defines a [PendingTypes] struct which stores information on parent types/ not-yet-resolved types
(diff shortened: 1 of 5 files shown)


src/system.rs
```diff
@@ -8,16 +8,12 @@ use crate::schema::Schema;
...
 use std::convert::TryInto;
 use std::io::ErrorKind;
 use std::rc::Rc;
 
-// TODO: this could be replaced with &mut Context
-pub type SharedPendingTypes = Rc<RefCell<PendingTypes>>;
-
-/// Defines a [Context] struct which stores information on parent types/ not-yet-resolved types
+/// Defines a [PendingTypes] struct which stores information on parent types/ not-yet-resolved types
 /// while loading a [Schema] with types which is used to resolve self referencing schema types
```
zslayton, line 17: This concept may not be obvious to new readers, so a more verbose
explanation might be worthwhile.

```suggestion
/// Stores information about types that are in the process of being defined.
///
/// An ISL type definition can include types that are not yet fully defined. 
/// For example, an ISL type definition might include:
/// * A reference to itself. This could happen in a recursive structure like a
///   linked list or binary tree.
/// * A nested anonymous type definition.
/// Because the [SchemaSystem] does not yet know the complete definition
/// of these types, it cannot find them in the [TypeStore].
/// An instance of [PendingTypes] is used to track information about types
/// that we do not have a complete definition for yet. When the 
/// [SchemaSystem] finishes loading these types, the type definitions in
/// [PendingTypes] can be promoted the [TypeStore].
```

zslayton (approved):
This simplified a lot! 🚀 One doc comment suggestion below.

PR #31, opened 2021-09-17
adds changes to create a module for `Isl*` implementations

### LINKED ISSUE #26: Create a module for all `Isl*` implementations to construct types/constraints
As per #24, there is a file isl which has all `Isl*` implementations of constructing types/constraints but instead it could be in a separate module and all these implementations can then be divided into files.
Also, add a module-level documentation to it explaining the role of these `Isl* `structs/enum in how these types helps constructing types/constraints programmatically and point to relevant implementations.

desaikd:
*Issue:*
Fixes #26

*Description of changes:*
This PR works on creating a separate module for all `Isl*` implementations

*Changes:*
- adds doc comment to define the role of `Isl*` implementations.
- adds a module called Isl for `IslType`, `IslConstraint` and `IslTypeRef` implementation.


zslayton (approved):
Looks good!

PR #32, opened 2021-09-22
adds changes for anonymous and named `TypeDefinition`s and `Isl*` implementations

### LINKED ISSUE #25: Create different implementations of `TypeDefinition` for named and unnamed types
As per #24, there is only one implementation of `TypeDefinition` which handles both named and anonymous types. According to the grammar defined in [Ion schema specification](https://amzn.github.io/ion-schema/docs/spec.html#grammar), the rules for defining a named and anonymous types may differ in where they are allowed.
For this reason it might be necessary to have two different implementations for both of these which would look like:
```rust
trait TypeDefiniton {
    fn constraints(&self) -> &[Constraint];
    fn parse_from_isl_type_and_update_type_store(
        isl_type: &IslType,
        type_store: &SharedTypeStore,
    ) -> IonSchemaResult<Self>;
}

struct NamedTypeDefinition {
      name: Option<String>,
      constraints: Vec<Constraint>
}

struct AnonymousTypeDefinition {
      constraints: Vec<Constraint>
}
```
(there could be more implementations of `TypeDefinition` like `IslCoreTypeDefintiions`)

Also, add a separate check for `type::` annotation for named types. (reference discussion: https://github.com/amzn/ion-schema-rust/pull/24#discussion_r710480716)

desaikd:
*Issue:*
Fixes #25 

*Description of changes:*
- adds `IslType` enum with two different implementations for anonymous and named ISL types
- adds `TypeDefinition` enum with two implementations for anonymous and named type definitions
- Fixes a bug in `PendingType`s to store correct `TypeId` in `Constraint`s (`type_id + type_store.types_by_id.len()`)

*Tests:*
adds changes for using corresponding anonymous/named types in unit tests.

### DIFF
--- a/src/isl/isl_constraint.rs
+++ b/src/isl/isl_constraint.rs
@@ -1,8 +1,70 @@
 use crate::isl::isl_type_reference::IslTypeRef;
+use crate::result::{invalid_schema_error_raw, IonSchemaResult};
+use ion_rs::value::owned::OwnedElement;
+use ion_rs::value::{Element, Sequence};
+use ion_rs::IonType;
 
 /// Represents a public facing API for schema constraints [IslConstraint] which stores IslTypeRef
 #[derive(Debug, Clone, PartialEq)]
 pub enum IslConstraint {
     AllOf(Vec<IslTypeRef>),
     Type(IslTypeRef),
+}
+
+impl IslConstraint {
+    /// Creates a [IslConstraint::Type] using the [IslTypeRef] referenced inside it
+    pub fn type_constraint(isl_type: IslTypeRef) -> IslConstraint {
+        IslConstraint::Type(isl_type)
+    }
+
+    /// Creates a [IslConstraint::AllOf] using the [IslTypeRef] referenced inside it
+    pub fn all_of(isl_types: Vec<IslTypeRef>) -> IslConstraint {
+        IslConstraint::AllOf(isl_types)
+    }
+
+    /// Parse constraints inside an [OwnedElement] to an [IslConstraint]
+    pub fn parse_from_ion_element(
+        constraint_name: &str,
+        value: &OwnedElement,
+        type_name: &str,
+    ) -> IonSchemaResult<IslConstraint> {
+        // TODO: add more constraints to match below
+        match constraint_name {
+            "all_of" => {
+                //TODO: create a method/macro for this ion type check which can be reused
+                if value.ion_type() != IonType::List {
+                    return Err(invalid_schema_error_raw(format!(
+                        "all_of constraint was a {:?} instead of a list",
+                        value.ion_type()
+                    )));
+                }
+                let types: Vec<IslTypeRef> = value
+                    .as_sequence()
+                    .unwrap()
+                    .iter()
+                    .map(|e| IslTypeRef::parse_from_ion_element(e))
+                    .collect::<IonSchemaResult<Vec<IslTypeRef>>>()?;
+                Ok(IslConstraint::AllOf(types))
+            }
+            "type" => {
+                if value.ion_type() != IonType::Symbol && value.ion_type() != IonType::Struct {
+                    return Err(invalid_schema_error_raw(format!(
+                        "type constraint was a {:?} instead of a symbol/struct",
+                        value.ion_type()
+                    )));
+                }
+                let type_reference: IslTypeRef = IslTypeRef::parse_from_ion_element(value)?;
+                Ok(IslConstraint::Type(type_reference))
+            }
+            _ => {
+                return Err(invalid_schema_error_raw(
+                    "Type: ".to_owned()
+                        + type_name
+                        + " can not be built as constraint: "
+                        + constraint_name
+                        + " does not exist",
+                ))
+            }
+        }
+    }
 }
--- a/src/isl/isl_constraint.rs
+++ b/src/isl/isl_constraint.rs
@@ -13,17 +13,18 @@ pub enum IslConstraint {
 
 impl IslConstraint {
     /// Creates a [IslConstraint::Type] using the [IslTypeRef] referenced inside it
+    // type is rust keyword hence this method is named type_constraint unlike other ISL constraint methods
     pub fn type_constraint(isl_type: IslTypeRef) -> IslConstraint {
         IslConstraint::Type(isl_type)
     }
 
     /// Creates a [IslConstraint::AllOf] using the [IslTypeRef] referenced inside it
-    pub fn all_of(isl_types: Vec<IslTypeRef>) -> IslConstraint {
-        IslConstraint::AllOf(isl_types)
+    pub fn all_of<A: Into<Vec<IslTypeRef>>>(isl_types: A) -> IslConstraint {
+        IslConstraint::AllOf(isl_types.into())
     }
 
     /// Parse constraints inside an [OwnedElement] to an [IslConstraint]
-    pub fn parse_from_ion_element(
+    pub fn from_ion_element(
         constraint_name: &str,
         value: &OwnedElement,
         type_name: &str,
--- a/src/isl/isl_type.rs
+++ b/src/isl/isl_type.rs
@@ -12,13 +12,13 @@ pub enum IslType {
 
 impl IslType {
     /// Creates a [IslType::Named] using the [IslConstraint] defined within it
-    pub fn named(name: String, constraints: Vec<IslConstraint>) -> IslType {
-        IslType::Named(IslTypeImpl::new(Some(name), constraints))
+    pub fn named<A: Into<String>, B: Into<Vec<IslConstraint>>>(name: A, constraints: B) -> IslType {
+        IslType::Named(IslTypeImpl::new(Some(name.into()), constraints.into()))
     }
 
     /// Creates a [IslType::Anonymous] using the [IslConstraint] defined within it
-    pub fn anonymous(constraints: Vec<IslConstraint>) -> IslType {
-        IslType::Anonymous(IslTypeImpl::new(None, constraints))
+    pub fn anonymous<A: Into<Vec<IslConstraint>>>(constraints: A) -> IslType {
+        IslType::Anonymous(IslTypeImpl::new(None, constraints.into()))
     }
 
     /// Provides the underlying constraints of [IslTypeImpl]
@@ -30,9 +30,9 @@ impl IslType {
     }
 }
 
-/// Represents both named and anonymous [IslType] which can be converted to a solid [TypeDefinition] using TypeStore
-/// named ISL type grammar: `type:: { name: <NAME>, <CONSTRAINT>...}`
-/// anonymous ISL type grammar: `{ <CONSTRAINT>... }`
+/// Represents both named and anonymous [IslType]s and can be converted to a solid [TypeDefinition] using TypeStore
+/// Named ISL type grammar: `type:: { name: <NAME>, <CONSTRAINT>...}`
+/// Anonymous ISL type grammar: `{ <CONSTRAINT>... }`
 #[derive(Debug, Clone, PartialEq)]
 pub struct IslTypeImpl {
     name: Option<String>,
@@ -75,7 +75,7 @@ impl IslTypeImpl {
         };
 
         if contains_annotations && type_name.is_none() {
-            // If a named type doesn't have name field throw an error
+            // If a named type doesn't have name field return an error
             return Err(invalid_schema_error_raw(
                 "Top level types must have a name field in its definition",
             ));
@@ -85,6 +85,12 @@ impl IslTypeImpl {
                 "Top level types must have `type::` annotation in their definition",
             ));
         }
+
+        // set the isl type name for any error that is thrown while parsing its constraints
+        let isl_type_name = match type_name.to_owned() {
+            Some(name) => name,
+            None => format!("{:?}", ion_struct),
+        };
 
         // parses all the constraints inside a Type
         for (field_name, value) in ion_struct.iter() {
@@ -98,14 +104,8 @@ impl IslTypeImpl {
                 }
             };
 
-            // set the isl type name for any error that is thrown while parsing its constraints
-            let mut isl_type_name = format!("{:?}", ion_struct);
-            if type_name.is_some() {
-                isl_type_name = type_name.to_owned().unwrap();
-            }
-
             let constraint =
-                IslConstraint::parse_from_ion_element(constraint_name, value, &isl_type_name)?;
+                IslConstraint::from_ion_element(constraint_name, value, &isl_type_name)?;
             constraints.push(constraint);
         }
         Ok(IslTypeImpl::new(type_name, constraints))
--- a/src/isl/isl_type.rs
+++ b/src/isl/isl_type.rs
@@ -80,13 +80,13 @@ impl IslTypeImpl {
                 "Top level types must have a name field in its definition",
             ));
         } else if !contains_annotations && type_name.is_some() {
-            // For named types if it does not have the `type::` annotation throw an error
+            // For named types if it does not have the `type::` annotation return an error
             return Err(invalid_schema_error_raw(
                 "Top level types must have `type::` annotation in their definition",
             ));
         }
 
-        // set the isl type name for any error that is thrown while parsing its constraints
+        // set the isl type name for any error that is returned while parsing its constraints
         let isl_type_name = match type_name.to_owned() {
             Some(name) => name,
             None => format!("{:?}", ion_struct),
--- a/src/isl/isl_type.rs
+++ b/src/isl/isl_type.rs
@@ -1,115 +1,50 @@
 use crate::isl::isl_constraint::IslConstraint;
-use crate::isl::isl_type_reference::IslTypeRef;
 use crate::result::{invalid_schema_error, invalid_schema_error_raw, IonSchemaResult};
 use ion_rs::value::owned::{text_token, OwnedElement, OwnedSymbolToken};
-use ion_rs::value::{Element, Sequence, Struct, SymbolToken};
-use ion_rs::IonType;
+use ion_rs::value::{Element, Struct, SymbolToken};
 
-/// Represents the public facing schema [IslType] which converts given ion content in the schema file
-/// into an internal representation of ISL types(not-yet-resolved types).
+/// Represents a type in an ISL schema.
 #[derive(Debug, Clone, PartialEq)]
 pub enum IslType {
-    NamedIslType(NamedIslType),
-    AnonymousIslType(AnonymousIslType),
+    Named(IslTypeImpl),
+    Anonymous(IslTypeImpl),
 }
 
-/// Represents a public facing schema type [AnonymousIslType] which can be converted to a solid [AnonymousTypeDefinition] using TypeStore
-/// anonymous ISL type grammar: `{ <CONSTRAINT>... }`
-#[derive(Debug, Clone)]
-pub struct AnonymousIslType {
-    constraints: Vec<IslConstraint>,
-}
-
-impl AnonymousIslType {
-    pub fn new(constraints: Vec<IslConstraint>) -> Self {
-        Self { constraints }
+impl IslType {
+    /// Creates a [IslType::Named] using the [IslConstraint] defined within it
+    pub fn named(name: String, constraints: Vec<IslConstraint>) -> IslType {
+        IslType::Named(IslTypeImpl::new(Some(name), constraints))
     }
 
+    /// Creates a [IslType::Anonymous] using the [IslConstraint] defined within it
+    pub fn anonymous(constraints: Vec<IslConstraint>) -> IslType {
+        IslType::Anonymous(IslTypeImpl::new(None, constraints))
+    }
+
+    /// Provides the underlying constraints of [IslTypeImpl]
     pub fn constraints(&self) -> &[IslConstraint] {
-        &self.constraints
-    }
-
-    /// Parse constraints inside an [OwnedElement] to an [AnonymousIslType]
-    pub fn parse_from_owned_element(ion: &OwnedElement) -> IonSchemaResult<Self> {
-        let mut constraints = vec![];
-
-        let ion_struct = try_to!(ion.as_struct());
-
-        // parses all the constraints inside a Type
-        for (field_name, value) in ion_struct.iter() {
-            let constraint_name = match field_name.text() {
-                Some("name") => continue, // if the field_name is "name" then it's the type name not a constraint
-                Some(name) => name,
-                None => {
-                    return Err(invalid_schema_error_raw(
-                        "A type name symbol token does not have any text",
-                    ))
-                }
-            };
-            // TODO: add more constraints to match below
-            let constraint = match constraint_name {
-                "all_of" => {
-                    //TODO: create a method/macro for this ion type check which can be reused
-                    if value.ion_type() != IonType::List {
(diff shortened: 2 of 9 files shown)


src/isl/isl_type.rs
```diff
@@ -1,61 +1,151 @@
...
+
+    /// Parse constraints inside an [OwnedElement] to an [AnonymousIslType]
+    pub fn parse_from_owned_element(ion: &OwnedElement) -> IonSchemaResult<Self> {
+        let mut constraints = vec![];
+
+        let ion_struct = try_to!(ion.as_struct());
+
+        // parses all the constraints inside a Type
+        for (field_name, value) in ion_struct.iter() {
+            let constraint_name = match field_name.text() {
```
zslayton, line 40: Could/should this parsing logic live on the `Constraint` type?

src/isl/isl_type.rs
```diff
@@ -109,6 +199,13 @@ impl TryFrom<&OwnedElement> for IslType {
...
         }
-        Ok(IslType::new(type_name, constraints))
+        Ok(NamedIslType::new(type_name, constraints))
+    }
+}
+
+/// Provides an implementation of PartialEq to compare two IslTypes
+impl PartialEq for NamedIslType {
+    fn eq(&self, other: &Self) -> bool {
+        self.name == other.name && self.constraints.eq(&other.constraints)
```
zslayton, line 209: This looks like it's the same `PartialEq` implementation that we'd get from `#[derive(PartialEq)]`. If that's true, consider deriving it instead. Same thought for `AnonymousIslType`'s `PartialEq` impl.

src/isl/mod.rs
```diff
@@ -30,66 +30,80 @@ pub mod isl_type_reference;
...
     isl_type1,isl_type2,
     case::type_constraint_with_anonymous_type(
-    load(r#" // For a schema with single anonymous type
-            {type: int}
-        "#),
-    IslType::new(None, vec![IslConstraint::Type(IslTypeRef::CoreIslType(IonType::Integer))])
+        load_anonymous_type(r#" // For a schema with single anonymous type
+                {type: int}
+            "#),
+        IslType::AnonymousIslType(AnonymousIslType::new(vec![IslConstraint::Type(IslTypeRef::CoreIslType(IonType::Integer))]))
```
zslayton, line 70: Following other refactorings suggested above, this would look something like:
```rust
IslType::anonymous(
    vec![
        IslConstraint::type(
            IslTypeRef::core(IonType::Integer)
        )
    ]
)
```
(Spaced out more here for readability.)

src/system.rs
```diff
@@ -52,9 +51,15 @@ impl PendingTypes {
                 .ok_or(unresolvable_schema_error_raw(
                     "Unable to load schema due to unresolvable type",
                 ))?;
-            match type_def.to_owned().name() {
-                Some(name) => type_store.add_named_type(name, type_def),
-                None => type_store.add_anonymous_type(type_def),
+            match type_def.to_owned() {
+                // Some(name) => type_store.add_named_type(name, type_def),
+                // None => type_store.add_anonymous_type(type_def),
```
zslayton, line 56:
```suggestion
```

src/types.rs
```diff
@@ -33,21 +33,101 @@ impl TypeRef {
...
 
-/// A Type consists of an optional name and zero or more constraints.
-///
-/// Unless otherwise specified, the constraint `type: any` is automatically applied.
+/// Represents a [TypeDefinition] which stores a resolved ISL type using [TypeStore]
+#[derive(Debug, Clone, PartialEq)]
+pub enum TypeDefinition {
+    NamedTypeDefinition(NamedTypeDefinition),
+    AnonymousTypeDefinition(AnonymousTypeDefinition),
+}
```
zslayton, line 41:
```suggestion
pub enum TypeDefinition {
    Named(NamedTypeDefinition),
    Anonymous(AnonymousTypeDefinition),
}
```

src/types.rs
```diff
@@ -33,21 +33,101 @@ impl TypeRef {
...
+
+/// A [AnonymousTypeDefinition] consists of a name and zero or more constraints.
```
zslayton, line 43:
```suggestion
/// An [AnonymousTypeDefinition] consists zero or more constraints.
```

src/types.rs
```diff
@@ -33,21 +33,101 @@ impl TypeRef {
...
+        let type_def = AnonymousTypeDefinition::new(constraints);
+
+        // update with this resolved type_def to context for type_id
+        pending_types.update_anonymous_type(type_id, type_def.to_owned(), type_store);
+
+        Ok(type_def)
+    }
+}
+
+impl PartialEq for AnonymousTypeDefinition {
```
zslayton, line 102: I think this can be `#[derive()]`ed.

src/types.rs
```diff
@@ -33,21 +33,101 @@ impl TypeRef {
...
+        type_store: &mut TypeStore,
+        pending_types: &mut PendingTypes,
+    ) -> IonSchemaResult<Self> {
+        let mut constraints = vec![];
+
+        // add this unresolved type to context for type_id
+        let type_id = pending_types.add_type(type_store);
+
+        // convert IslConstraint to Constraint
+        for isl_constraint in isl_type.constraints() {
```
zslayton, line 70: I think this logic should live in `Constraint`?

src/types.rs
```diff
@@ -33,21 +33,101 @@ impl TypeRef {
...
+    }
+
+    fn validate(&self, value: &OwnedElement, issues: &mut Violations) {
+        todo!()
+    }
+}
+
+/// A [NamedTypeDefinition] consists of a name and zero or more constraints.
+#[derive(Debug, Clone)]
+pub struct NamedTypeDefinition {
```
zslayton, line 120: `NamedTypeDefinition` and `AnonymousTypeDefinition` are so closely related that having two separate structs with nearly identical method implementations feels like overkill to me. Here are some alternatives to consider.

--------

### Option 1

We could restructure the `TypeDefinition` enum to hold the name:

```rust
pub enum TypeDefinition {
    Named(String, TypeDefinition), // This variant has a `String` for the name
    Anonymous(TypeDefinition),
}
```

With this option, we can still distinguish between the named and anonymous cases via the enum but we eliminate a largely redundant type.

--------

### Option 2

This requires a bit more effort but would still simplify things quite a bit; we could redefine `NamedTypeDefinition` as:
```rust
pub struct NamedTypeDefinition {
    name: String,
    type_definition: TypeDefinition
}
```
We still have a second full type (`NamedTypeDefinition`), but by re-using `TypeDefinition` via composition, we allow all of the methods/functionality to go on living on the `TypeDefinition` type itself.

--------

My personal preference is option 1.

src/isl/mod.rs
```diff
@@ -30,66 +30,80 @@ pub mod isl_type_reference;
...
+    // helper function to create AnonymousIslType for isl tests
+    fn load_anonymous_type(text: &str) -> IslType {
+        IslType::AnonymousIslType(
+            AnonymousIslType::parse_from_owned_element(
+                &element_reader()
+                    .read_one(text.as_bytes())
+                    .expect("parsing failed unexpectedly"),
+            )
+            .unwrap(),
+        )
```
popematt, line 61: question—the Kotlin implementation of `Schema` has a function [`newType(String)`](https://github.com/amzn/ion-schema-kotlin/blob/master/src/com/amazon/ionschema/Schema.kt#L78) that is similar to these ones. Would it be worth putting these functions in some other module where they could be reused by [`Schema`](https://github.com/amzn/ion-schema-rust/blob/main/src/schema.rs)?

src/schema.rs
```diff
@@ -58,7 +58,7 @@ impl Schema {
     /// instance plus the provided type.  Note that the added type
     /// in the returned instance will hide a type of the same name
     /// from this instance.
-    fn plus_type(&self, schema_type: TypeDefinition) -> Self {
+    fn plus_type(&self, schema_type: NamedTypeDefinition) -> Self {
```
popematt, line 61: question—apologies if I'm getting the models confused, but should this be `IslType` (the model for constructing new types)?

src/isl/isl_type.rs
```diff
@@ -1,61 +1,151 @@
...
+    NamedIslType(NamedIslType),
+    AnonymousIslType(AnonymousIslType),
+}
+
+/// Represents a public facing schema type [AnonymousIslType] which can be converted to a solid [AnonymousTypeDefinition] using TypeStore
+/// anonymous ISL type grammar: `{ <CONSTRAINT>... }`
 #[derive(Debug, Clone)]
-pub struct IslType {
-    name: Option<String>,
+pub struct AnonymousIslType {
```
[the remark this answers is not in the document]
popematt: @zslayton, I like the spacing. I think the spacing makes it easy to see how the Rust functions correspond to different parts of the ISL syntax (which is kind of like having some bonus documentation 🥳 ).

src/isl/isl_type.rs
```diff
@@ -1,61 +1,151 @@
...
+    /// Parse constraints inside an [OwnedElement] to an [AnonymousIslType]
+    pub fn parse_from_owned_element(ion: &OwnedElement) -> IonSchemaResult<Self> {
+        let mut constraints = vec![];
+
+        let ion_struct = try_to!(ion.as_struct());
+
+        // parses all the constraints inside a Type
+        for (field_name, value) in ion_struct.iter() {
+            let constraint_name = match field_name.text() {
+                Some("name") => continue, // if the field_name is "name" then it's the type name not a constraint
```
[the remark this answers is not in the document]
popematt: ISL allows for open content, which means that you can stick a "name" field anywhere, but it will be ignored if it's not in a top-level type. Skipping it silently is what the spec requires.

src/isl/isl_type.rs
```diff
@@ -1,61 +1,151 @@
...
-use ion_rs::value::owned::OwnedElement;
+use crate::result::{invalid_schema_error, invalid_schema_error_raw, IonSchemaResult};
+use ion_rs::value::owned::{text_token, OwnedElement, OwnedSymbolToken};
 use ion_rs::value::{Element, Sequence, Struct, SymbolToken};
 use ion_rs::IonType;
-use std::convert::TryFrom;
 
-/// Represents a public facing schema type [IslType] which can be converted to a solid TypeRef using TypeStore
+/// Represents the public facing schema [IslType] which converts given ion content in the schema file
+/// into an internal representation of ISL types(not-yet-resolved types).
```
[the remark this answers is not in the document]
desaikd: I had added some developer comments over here: https://github.com/amzn/ion-schema-rust/blob/1f25a9038533b6902669ebe9b25b6cee58d40a3b/src/isl/mod.rs#L5-L28 Let me know if that is something you were looking for or if there's any changes there as well.

src/schema.rs
desaikd: Yes. That is correct. I will change this to have `IslType` in the method parameter.

src/isl/isl_type.rs
```diff
@@ -1,115 +1,50 @@
...
-/// Provides an implementation of PartialEq to compare two IslTypes
-impl PartialEq for AnonymousIslType {
-    fn eq(&self, other: &Self) -> bool {
-        self.constraints.eq(&other.constraints)
     }
 }
 
-/// Represents a public facing schema type [NamedIslType] which can be converted to a solid [NamedTypeDefinition] using TypeStore
+/// Represents both named and anonymous [IslType] which can be converted to a solid [TypeDefinition] using TypeStore
 /// named ISL type grammar: `type:: { name: <NAME>, <CONSTRAINT>...}`
```
zslayton:
```suggestion
/// Represents both named and anonymous [IslType]s and can be converted to a solid [TypeDefinition] using [TypeStore].
/// Named ISL type grammar: `type:: { name: <NAME>, <CONSTRAINT>...}`
```
As I was reading this my brain turned the two lines into one big sentence: "...using TypeStore named ISL type grammar: `type::`..." 😛

src/isl/isl_type.rs
```diff
@@ -122,32 +57,35 @@ impl NamedIslType {
...
-            None => {
-                return Err(invalid_schema_error_raw(
-                    "Named type must have a name field in its definition",
-                ))
-            } // If a named type doesn't have name field throw an error
+            None => None, // If there is no name field then it is an anonymous type
         };
 
+        if contains_annotations && type_name.is_none() {
+            // If a named type doesn't have name field throw an error
```
zslayton, line 78: A nit for some of these comments: Rust doesn't have a `throw` mechanism.
```suggestion
            // If a named type doesn't have name field return an error
```

src/isl/isl_type.rs
```diff
@@ -159,53 +97,17 @@ impl NamedIslType {
...
-                            + " does not exist",
-                    ))
-                }
-            };
+
+            // set the isl type name for any error that is thrown while parsing its constraints
+            let mut isl_type_name = format!("{:?}", ion_struct);
+            if type_name.is_some() {
+                isl_type_name = type_name.to_owned().unwrap();
+            }
```
zslayton, line 105: This `format!` seems expensive. Why do anonymous types need a name? That is, why bother building a new string representation we won't use?

If there's a good reason, consider deferring the `format!` until you're sure it's necessary:
```suggestion
            let isl_type_name = match type_name {
                Some(name) => name,
                None => format!("{:?}", ion_struct)
            };
```

src/isl/isl_type_reference.rs
```diff
@@ -13,17 +13,32 @@ use ion_rs::IonType;
 #[derive(Debug, Clone, PartialEq)]
 pub enum IslTypeRef {
     /// Represents a reference to one of ISL's built-in core types:
-    CoreIslType(IonType),
+    CoreIsl(IonType),
```
zslayton, line 16: I have a mild preference for naming this variant `Core`; the spec already refers to them as "[Core Types](https://amzn.github.io/ion-schema/docs/spec.html#core-types)".

src/isl/isl_type_reference.rs
```diff
@@ -13,17 +13,32 @@ use ion_rs::IonType;
...
     // TODO: add ImportType(Import) where ImportType could either point to a schema represented by an id with all the types or a single type from inside it
     /// represents an unnamed type definition reference
-    AnonymousType(AnonymousIslType),
+    Anonymous(IslTypeImpl),
 }
 
 // TODO: add a check for nullable type reference
 impl IslTypeRef {
+    /// Creates a [IslTypeRef::CoreIsl] using the [IonType] referenced inside it
+    pub fn core_isl(ion_type: IonType) -> IslTypeRef {
```
zslayton, line 28: Same thought here RE: `core`.

src/isl/mod.rs
```diff
@@ -1,36 +1,76 @@
+//! Provides a way to construct ISL types/constraints programmatically into an internal representation.
```
zslayton, line 1:
```suggestion
//! Provides a way to construct ISL types/constraints programmatically.
```

src/isl/mod.rs
```diff
@@ -67,43 +107,43 @@ mod isl_tests {
...
+        IslType::named("my_int".to_owned(), vec![IslConstraint::type_constraint(IslTypeRef::anonymous(vec![IslConstraint::type_constraint(IslTypeRef::core_isl(IonType::Integer))]))])
     ),
     case::type_constraint_with_nested_multiple_types(
         load_named_type(r#" // For a schema with nested multiple types
                 type:: { name: my_int, type: { type: int }, type: { type: my_int } }
             "#),
-        IslType::NamedIslType(NamedIslType::new("my_int".to_owned(), vec![IslConstraint::Type(IslTypeRef::AnonymousType(AnonymousIslType::new(vec![IslConstraint::Type(IslTypeRef::CoreIslType(IonType::Integer))]))), IslConstraint::Type(IslTypeRef::AnonymousType(AnonymousIslType::new(vec![IslConstraint::Type(IslTypeRef::NamedType("my_int".to_owned()))])))]))
+        IslType::named("my_int".to_owned(), vec![IslConstraint::type_constraint(IslTypeRef::anonymous(vec![IslConstraint::type_constraint(IslTypeRef::core_isl(IonType::Integer))])), IslConstraint::type_constraint(IslTypeRef::anonymous(vec![IslConstraint::Type(IslTypeRef::named("my_int".to_owned()))]))])
```
zslayton, line 140: Making these methods generic over `Into<String>` and something similar for a `Vec` would eliminate a lot of `.to_owned()`s and `vec!`s.

src/types.rs
```diff
@@ -221,75 +160,73 @@ mod type_definition_tests {
...
+        IslType::named("my_int".to_owned(), vec![IslConstraint::type_constraint(IslTypeRef::anonymous(vec![IslConstraint::type_constraint(IslTypeRef::core_isl(IonType::Integer))])), IslConstraint::type_constraint(IslTypeRef::anonymous(vec![IslConstraint::type_constraint(IslTypeRef::named("my_int".to_owned()))]))]),
+        TypeDefinition::named("my_int".to_owned(), vec![Constraint::type_constraint(1), Constraint::type_constraint(3)])
     ),
     case::all_of_constraint(
         /* For a schema with all_of type as below:
             { all_of: [{ type: int }] }
         */
-        IslType::AnonymousIslType(AnonymousIslType::new(vec![IslConstraint::AllOf(vec![IslTypeRef::AnonymousType(AnonymousIslType::new(vec![IslConstraint::Type(IslTypeRef::CoreIslType(IonType::Integer))]))])])),
-        TypeDefinition::AnonymousTypeDefinition(AnonymousTypeDefinition::new(vec![Constraint::AllOf(AllOfConstraint::new(vec![1]))]))
+        IslType::anonymous(vec![IslConstraint::all_of(vec![IslTypeRef::anonymous(vec![IslConstraint::type_constraint(IslTypeRef::core_isl(IonType::Integer))])])]),
```
zslayton, line 205: Similar thought here RE: eliminating `to_owned()`s and `vec!`s. We can do this later, though.

src/isl/mod.rs
```diff
@@ -1,36 +1,76 @@
+//! Provides a way to construct ISL types/constraints programmatically into an internal representation.
+//!
+//! This module consists of three submodules that implement public facing API for constructing ISL types/constraint:
+//!
+//! * `isl_type` module represents a public facing schema type [IslType] which converts given ion content in the schema file
```
zslayton, line 5: This explanation covers a lot, but it's currently being exposed as part of the library's documentation (via the `//!` doc comment format). The information it contains straddles the line between being relevant to end users and being relevant to `ion-schema-rust` developers.

As a `TODO`, we should do one of the following:
1. Make this a developer comment (`//` instead of `//!`)
OR
2. Write the explanation with end users in mind as. In particular:
* Remove mentions of what's "public facing"
* Remove mentions of internal representations
* Remove mentions of how type resolution works
OR
3. Split it into two comments, one for the user's benefit and one for the maintainers.

My preference is number 3.

src/isl/mod.rs
```diff
@@ -1,36 +1,76 @@
...
+//! and a resolved type TypeDefinition) using [TypeStore].
+//!
+//! * `isl_constraint` module represents a public facing API for schema constraints [IslConstraint]
+//! which converts given ion content in the schema file into an internal representation of ISL constraint(not-yet-resolved constraints).
+//! This stores [IslTypeRef]s for the given ISL constraint.
+//!
+//! * `isl_type_reference` module provides an internal representation of a schema type reference.
+//! The type reference grammar is defined in the [Ion Schema Spec]: https://amzn.github.io/ion-schema/docs/spec.html#grammar
+//!
+//! ## Example usage of `isl` module to create an [IslType]:
```
zslayton, line 17: Having this example in the module documentation is excellent! 👍

src/isl/mod.rs
```diff
@@ -1,36 +1,76 @@
...
+//!                             IslConstraint::type_constraint(
+//!                                 IslTypeRef::core_isl(IonType::Boolean)
+//!                             )
+//!                         ]
+//!                     )
+//!                 ]
+//!             )
+//!         ]
+//!     );
+//!     assert_eq!(isl_type.constraints().len(), 2);
```
zslayton, line 54: TODO: We might consider actually creating a schema system and validating something trivial here unless that's too verbose.

src/isl/isl_type_reference.rs
desaikd: Yes, I think that should be fine as far as it is similar to what the spec says.

src/isl/isl_constraint.rs
```diff
@@ -1,8 +1,70 @@
...
 /// Represents a public facing API for schema constraints [IslConstraint] which stores IslTypeRef
 #[derive(Debug, Clone, PartialEq)]
 pub enum IslConstraint {
     AllOf(Vec<IslTypeRef>),
     Type(IslTypeRef),
 }
+
+impl IslConstraint {
+    /// Creates a [IslConstraint::Type] using the [IslTypeRef] referenced inside it
+    pub fn type_constraint(isl_type: IslTypeRef) -> IslConstraint {
```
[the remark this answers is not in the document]
desaikd: We won't be able to use type here as `type` is reserved keyword for rust.

src/isl/mod.rs
desaikd: Maybe we can add an example for schema system within the schema system module and keep this as it is for it represents `isl` module example. Thoughts?

src/isl/isl_constraint.rs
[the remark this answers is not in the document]
zslayton: Ah, good point! Could you add a dev comment explaining why it doesn't follow the naming convention we use elsewhere?

src/isl/mod.rs
zslayton: My sense is that anyone reading the module documentation is hoping to figure out how to build a `Schema` from `Isl*` types. (Said another way: an end user would only build an `IslType` programmatically for one reason.)  I think this example should show building a `Schema` from an `IslType` (via the `SchemaSystem`) and the schema system module's example can show building a `Schema` from some an Ion/ISL document. The latter could also link to the former.

desaikd: Reference issue: https://github.com/amzn/ion-schema-rust/issues/33

popematt (approved):

src/isl/isl_type.rs
```diff
@@ -86,6 +86,12 @@ impl IslTypeImpl {
             ));
         }
 
+        // set the isl type name for any error that is thrown while parsing its constraints
```
zslayton, line 89: There are still a few references to "throwing" errors in the comments.

src/isl/mod.rs
```diff
@@ -107,43 +111,43 @@ mod isl_tests {
...
             "#),
-        IslType::named("my_int".to_owned(), vec![IslConstraint::type_constraint(IslTypeRef::core_isl(IonType::Integer))])
+        IslType::named("my_int", [IslConstraint::type_constraint(IslTypeRef::core(IonType::Integer))])
     ),
     case::type_constraint_with_self_reference_type(
         load_named_type(r#" // For a schema with self reference type
                 type:: { name: my_int, type: my_int }
             "#),
-        IslType::named("my_int".to_owned(), vec![IslConstraint::type_constraint(IslTypeRef::named("my_int".to_owned()))])
+        IslType::named("my_int", [IslConstraint::type_constraint(IslTypeRef::named("my_int"))])
```
zslayton, line 126: This is a really nice improvement, thanks for setting up the generics!

zslayton (approved):
This looks great, thanks for the extra polish!

PR #35, opened 2021-10-06
adds examples for loading a schema

desaikd:
*Description of changes:*
adds examples for loading a schema using schema system

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

### DIFF
--- a/examples/cli.yaml
+++ b/examples/cli.yaml
@@ -1,0 +1,20 @@
+name: ion-schema-rust
+author: Ion Team
+about: An implementation of Amazon Ion Schema in Rust
+args:
+  - authorities:
+      short: a
+      long: authorities
+      required: true
+      multiple: true
+      min_values: 1
+      value_name: AUTHORITIES
+      help: Set Authorities to be added to the schema system for loading a schema
+      takes_value: true
+  - input:
+        short: f
+        long: input
+        required: true
+        value_name: INPUT
+        help: Provide the Ion Schema file that needs to be loaded
+        takes_value: true
--- a/examples/load_schema.rs
+++ b/examples/load_schema.rs
@@ -1,0 +1,42 @@
+#[macro_use]
+extern crate clap;
+use clap::App;
+use ion_schema_rust::authority::{DocumentAuthority, FileSystemDocumentAuthority};
+use ion_schema_rust::result::IonSchemaResult;
+use ion_schema_rust::system::SchemaSystem;
+use std::path::Path;
+
+fn main() -> IonSchemaResult<()> {
+    // load the YAML file which has all the CLI commands information stored in it
+    let yaml = load_yaml!("cli.yaml");
+    let matches = App::from_yaml(yaml).get_matches();
+
+    // Extract the user provided authorities
+    let authorities: Vec<_> = matches.values_of("authorities").unwrap().collect();
+
+    // Extract schema file provided by user
+    let schema_id = matches.value_of("input").unwrap();
+
+    // Set up authorities vector
+    let mut document_authorities: Vec<Box<dyn DocumentAuthority>> = vec![];
+
+    for authority in authorities {
+        document_authorities.push(Box::new(FileSystemDocumentAuthority::new(Path::new(
+            authority,
+        ))))
+    }
+
+    // Create a new schema system from given document authorities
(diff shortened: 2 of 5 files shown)


examples/cli.yaml
```diff
@@ -0,0 +1,20 @@
...
+about: An implementation of Amazon Ion Schema in Rust
+args:
+  - authorities:
+      short: a
+      long: authorities
+      required: true
+      multiple: true
+      min_values: 1
+      value_name: AUTHORITIES
+      help: Set Authorities to be added to the schema system for loading a schema
```
popematt, line 12: question—how does the syntax for authorities work? Do you specify the type of authority and then any additional args? (Eg. file system authority, and then the base file path)

desaikd: For now there is just one authority implementation available that is the `FileSystemDocumentAuthority` hence I haven't provided additional option to specify the type of authority rather it just specifies base file path. I can change the `help` here to be more specific that this is just for file system authority.

zslayton: > I can change the help here to be more specific that this is just for file system authority.

👍

Since this is CLI help text, I'd suggest only describing the user-facing behavior and not the library's types.

```suggestion
      help: One or more directories that will be searched for the requested schema.
```

popematt: > For now there is just one authority implementation available that is the FileSystemDocumentAuthority

Yes... but the Kotlin implementation also has an authority that loads `ion-schema-schemas` from a JAR. I know that Rust doesn't have JARs, but if we're going to call this argument "Authority", then we should plan on there being more than one authority.

On the other hand, I think what Zack is proposing is a really good idea. If it just describes user-facing behavior, then we don't even need to call it an Authority anywhere.

examples/cli.yaml
```diff
@@ -0,0 +1,24 @@
...
+args:
+  - authorities:
+      short: b
+      long: base-file-path
+      required: true
+      multiple: true
+      min_values: 1
+      value_name: BASE-FILE-PATH
+      help: One or more directories that will be searched for the requested schema.
+      takes_value: true
```
zslayton, line 17: I suggest aligning the arg name (`authorities`) and the flag name (`base-file-path`). For a new user picking up this tool, I think an intuitive arg/flag name combination would be `directories`/`--directory`.

examples/cli.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - input:
+        short: s
+        long: schema
+        required: true
+        value_name: SCHEMA
```
zslayton, line 22: Similar thought here: let's name the `input` arg to `schema` so it aligns with the flag name.

examples/cli.yaml
```diff
@@ -0,0 +1,24 @@
...
+        help: Provide the Ion Schema file that needs to be loaded.
```
zslayton, line 23:
```suggestion
        help: The Ion Schema file to load.
```

src/system.rs
```diff
@@ -369,7 +369,7 @@ impl SchemaSystem {
     /// Requests each of the provided [Authority]s, in order, to resolve the requested schema id
     /// until one successfully resolves it.
     /// If an Authority throws an exception, resolution silently proceeds to the next Authority.
-    fn load_schema<A: AsRef<str>>(&mut self, id: A) -> IonSchemaResult<Rc<Schema>> {
+    pub fn load_schema<A: AsRef<str>>(&mut self, id: A) -> IonSchemaResult<Rc<Schema>> {
```
zslayton, line 372: In the future, we may want to hide the `Rc`-ness of this return value by making `Schema` private (`SchemaDefinition`?) and providing a public wrapper type (`Schema`?).

zslayton (approved):
Some naming suggestions below.

popematt (approved):

examples/load_schema.rs
```diff
@@ -0,0 +1,42 @@
+#[macro_use]
+extern crate clap;
+use clap::App;
+use ion_schema_rust::authority::{DocumentAuthority, FileSystemDocumentAuthority};
+use ion_schema_rust::result::IonSchemaResult;
+use ion_schema_rust::system::SchemaSystem;
```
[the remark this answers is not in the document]
desaikd: Reference issue: #37

Issue #34, opened 2021-09-30
Implement `IslImport` to represent an import in ISL


desaikd:
As per #32, there is an `isl` module created to represent ISL types/constraints. Adding an `IslImport` would let us have a representation of imports as defined in [Ion Schema Spec](https://amzn.github.io/ion-schema/docs/spec.html#imports)

Issue #22, opened 2021-09-01
`Schema#get_types()` should only return types that are defined by given `Schema`


desaikd:
As per https://github.com/amzn/ion-schema-rust/pull/21#discussion_r700519917,
> Since this is a method on Schema, this should only expose the types that are part of this schema.

`get_types()` should return only the types defined by given `Schema` rather than returning all imported types as well.
We will have to change `Schema` definition as below to store a `Vec<TypeId>` pointing to only the types that are defined within the schema and aren't imported from other schema.

```rust
struct Schema {
        id: String,
        type_store: Rc<TypeStore>,
        types: Vec<TypeId>, // only the types that are defined within this schema
}
```

PR #38, opened 2021-10-13
adds changes for implmentation of `IslImport`

desaikd:
*Issue #34, #22*

*Description of changes:*
This PR works on adding changes for `IslImport`

*Changes:*
- Implementation of IslImport to represent an import in ISL.
- Changes in `Resolver` `load_schema` and `schema_from_elements` method signature to not load imports to `type_store` when in the root of schema.
- New methods in type_store and pending_types to store imports separately.
- Changes in schema ` get_types` to rightly return only named types within that schema and not include the import types
- Changes in `SchemaTypeIterator` to return appropriate types within the schema and changes in `get_imports` to return import types.

*Tests:*
Ran all the unit tests as well as examples to verify if everything works fine.

*Questions:*
- Currently `get_types` returns named types + core types within the given schema. Should there be a `core_types` property in `TypeStore` which could store all the `core_types` differently. These can also be prefilled before we start loading a schema.
- `schema.rs` contains a method called `get_import` which gets imports for given schema id from within the schema.
- Does this mean we need an equivalent `Import` to `IslImport` which could be returned for `get_import` method.

### DIFF
--- a/src/authority.rs
+++ b/src/authority.rs
@@ -1,7 +1,8 @@
-use crate::result::IonSchemaResult;
+use crate::result::{unresolvable_schema_error_raw, IonSchemaResult};
 use ion_rs::result::IonError;
 use ion_rs::value::owned::OwnedElement;
 use ion_rs::value::reader::{element_reader, ElementReader};
+use std::collections::HashMap;
 use std::fmt::Debug;
 use std::fs;
 use std::path::{Path, PathBuf};
@@ -42,6 +43,35 @@ impl DocumentAuthority for FileSystemDocumentAuthority {
         // if absolute_path exists for the given id then load schema with file contents
         let ion_content = fs::read(absolute_path)?;
         let iterator = element_reader().iterate_over(&ion_content)?;
+        let schema_content = iterator.collect::<Result<Vec<OwnedElement>, IonError>>()?;
+        Ok(schema_content)
+    }
+}
+
+/// An [Authority] implementation that attempts to resolve schema ids to ion elements using the map.
+#[derive(Debug, Clone)]
+pub struct MapDocumentAuthority {
+    ion_content_by_id: HashMap<String, String>, // This map represents (id, ion content) which can used to resolve schema ids to Vec<OwnedElement>
+}
+
+impl MapDocumentAuthority {
+    pub fn new(ion_content_by_id: HashMap<String, String>) -> Self {
+        Self { ion_content_by_id }
+    }
+}
+
+impl DocumentAuthority for MapDocumentAuthority {
+    /// Returns a vector of [OwnedElement]s based on given schema id using ion_content_by_id map
+    fn elements(&self, id: &str) -> IonSchemaResult<Vec<OwnedElement>> {
+        // if ion content exists for the given id  in the map then return ion content as OwnedElements
+        let ion_content = self
+            .ion_content_by_id
+            .get(id)
+            .ok_or(unresolvable_schema_error_raw(format!(
+                "MapDocumentAuthority does not contain schema with id: {}",
+                id
+            )))?;
+        let iterator = element_reader().iterate_over(ion_content.as_bytes())?;
         let schema_content = iterator.collect::<Result<Vec<OwnedElement>, IonError>>()?;
         Ok(schema_content)
     }
--- a/src/authority.rs
+++ b/src/authority.rs
@@ -55,8 +55,13 @@ pub struct MapDocumentAuthority {
 }
 
 impl MapDocumentAuthority {
-    pub fn new(ion_content_by_id: HashMap<String, String>) -> Self {
-        Self { ion_content_by_id }
+    pub fn new<'a, I: IntoIterator<Item = (&'a str, &'a str)>>(ion_content_by_id: I) -> Self {
+        Self {
+            ion_content_by_id: ion_content_by_id
+                .into_iter()
+                .map(|(id, ion)| (id.to_owned(), ion.to_owned()))
+                .collect(),
+        }
     }
 }
 
--- a/src/isl/isl_import.rs
+++ b/src/isl/isl_import.rs
@@ -19,6 +19,7 @@ impl IslImport {
             IslImport::TypeAlias(isl_import_impl) => isl_import_impl.id(),
         }
     }
+
     /// Parse constraints inside an [OwnedElement] to an [IslImport]
     pub fn parse_from_ion_element(value: &OwnedElement) -> IonSchemaResult<IslImport> {
         let import = try_to!(value.as_struct());
@@ -39,6 +40,14 @@ impl IslImport {
         let alias = import
             .get("as")
             .and_then(|alias| alias.as_str().and_then(|a| Some(a.to_owned())));
+
+        if alias.is_none() {
+            return Ok(IslImport::Type(IslImportImpl::new(
+                id.to_owned(),
+                type_name.to_owned(),
+                None,
+            )));
+        }
 
         Ok(IslImport::TypeAlias(IslImportImpl::new(
             id.to_owned(),
@@ -69,5 +78,13 @@ impl IslImportImpl {
 
     fn id(&self) -> &String {
         &self.id
+    }
+
+    pub fn type_name(&self) -> &String {
+        &self.type_name
+    }
+
+    pub fn alias(&self) -> &Option<String> {
+        &self.alias
     }
 }
--- a/src/isl/isl_import.rs
+++ b/src/isl/isl_import.rs
@@ -1,0 +1,73 @@
+use crate::result::{invalid_schema_error, invalid_schema_error_raw, IonSchemaResult};
+use ion_rs::value::owned::OwnedElement;
+use ion_rs::value::{Element, Struct};
+
+/// Represents an import in an ISL schema.
+/// For more information: https://amzn.github.io/ion-schema/docs/spec.html#imports
+#[derive(Debug, Clone, PartialEq)]
+pub enum IslImport {
+    Schema(String),
+    Type(IslImportImpl),
+    TypeAlias(IslImportImpl),
+}
+
+impl IslImport {
+    pub fn id(&self) -> &String {
+        match self {
+            IslImport::Schema(id) => id,
+            IslImport::Type(isl_import_impl) => isl_import_impl.id(),
+            IslImport::TypeAlias(isl_import_impl) => isl_import_impl.id(),
+        }
+    }
+    /// Parse constraints inside an [OwnedElement] to an [IslImport]
+    pub fn parse_from_ion_element(value: &OwnedElement) -> IonSchemaResult<IslImport> {
+        let import = try_to!(value.as_struct());
+        let id = match import.get("id") {
+            Some(import_id) => try_to!(import_id.as_str()),
+            None => {
+                return Err(invalid_schema_error_raw(
+                    "import must have an id field in its definition",
+                ))
+            }
+        };
+
+        let type_name = match import.get("type") {
+            Some(type_name) => try_to!(type_name.as_str()),
+            None => return Ok(IslImport::Schema(id.to_owned())),
+        };
+
+        let alias = import
+            .get("as")
+            .and_then(|alias| alias.as_str().and_then(|a| Some(a.to_owned())));
+
+        Ok(IslImport::TypeAlias(IslImportImpl::new(
+            id.to_owned(),
+            type_name.to_owned(),
+            alias,
+        )))
+    }
+}
+
+/// Represents typed and type aliased [IslImport]s
+/// Typed import grammar: `{ id: <ID>, type: <TYPE_NAME> }`
+/// Type aliased import grammar: `{ id: <ID>, type: <TYPE_NAME>, as: <TYPE_ALIAS> }`
+#[derive(Debug, Clone, PartialEq)]
+pub struct IslImportImpl {
+    id: String,
+    type_name: String,
+    alias: Option<String>,
+}
+
+impl IslImportImpl {
+    pub fn new(id: String, type_name: String, alias: Option<String>) -> Self {
+        Self {
+            id,
+            type_name,
+            alias,
+        }
+    }
+
+    fn id(&self) -> &String {
+        &self.id
+    }
+}
--- a/src/isl/isl_import.rs
+++ b/src/isl/isl_import.rs
@@ -7,8 +7,8 @@ use ion_rs::value::{Element, Struct};
 #[derive(Debug, Clone, PartialEq)]
 pub enum IslImport {
     Schema(String),
-    Type(IslImportImpl),
-    TypeAlias(IslImportImpl),
+    Type(IslImportType),
+    TypeAlias(IslImportType),
 }
 
 impl IslImport {
@@ -21,7 +21,7 @@ impl IslImport {
     }
 
     /// Parse constraints inside an [OwnedElement] to an [IslImport]
-    pub fn parse_from_ion_element(value: &OwnedElement) -> IonSchemaResult<IslImport> {
+    pub fn from_ion_element(value: &OwnedElement) -> IonSchemaResult<IslImport> {
         let import = try_to!(value.as_struct());
         let id = match import.get("id") {
             Some(import_id) => try_to!(import_id.as_str()),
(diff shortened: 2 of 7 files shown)


src/isl/isl_import.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub fn id(&self) -> &String {
+        match self {
+            IslImport::Schema(id) => id,
+            IslImport::Type(isl_import_impl) => isl_import_impl.id(),
+            IslImport::TypeAlias(isl_import_impl) => isl_import_impl.id(),
+        }
+    }
+
+    /// Parse constraints inside an [OwnedElement] to an [IslImport]
+    pub fn parse_from_ion_element(value: &OwnedElement) -> IonSchemaResult<IslImport> {
```
zslayton, line 24:
```suggestion
    pub fn from_ion_element(value: &OwnedElement) -> IonSchemaResult<IslImport> {
```

src/schema.rs
```diff
@@ -121,19 +124,19 @@ mod schema_tests {
         load(r#" // For a schema with nested self reference type as below:
             type:: { name: my_int, type: { type: my_int } }
         "#).into_iter(),
-        2 // this includes my_int type and the anonymous type that uses my_int
+        1 // this includes my_int type and the anonymous type that uses my_int
```
zslayton, line 127: These numbers have been updated but the comments seem out of date.

Confirming my understanding: the anonymous type is _not_ returned, right? Because it isn't an "exported" type (i.e. a named type another schema could import)?

src/system.rs
```diff
@@ -43,22 +44,67 @@ impl PendingTypes {
 
     /// Adds all the types from Context into given TypeStore and clears Context types for loading next set of types
     /// this is used after a schema named type/root type is loaded entirely into Context
```
zslayton, line 46: This comment is also out of date, I believe.

src/isl/isl_import.rs
```diff
@@ -0,0 +1,90 @@
...
+            alias,
+        )))
+    }
+}
+
+/// Represents typed and type aliased [IslImport]s
+/// Typed import grammar: `{ id: <ID>, type: <TYPE_NAME> }`
+/// Type aliased import grammar: `{ id: <ID>, type: <TYPE_NAME>, as: <TYPE_ALIAS> }`
+#[derive(Debug, Clone, PartialEq)]
+pub struct IslImportImpl {
```
zslayton, line 64: In the spec, these two grammar elements are referred to as `IMPORT_TYPE` and `IMPORT_TYPE_ALIAS`. Could we name this `IslImportType`?

```suggestion
pub struct IslImportType {
```

src/schema.rs
```diff
@@ -65,25 +60,33 @@ impl Schema {
 
 /// Provides an Iterator which returns [Type]s inside a [Schema]
 pub struct SchemaTypeIterator {
-    types: Rc<TypeStore>,
+    type_store: Rc<TypeStore>,
     index: usize,
+    types: Vec<TypeId>,
```
zslayton, line 65: Allocating a `Vec` for each iterator strikes me as a bit expensive. Now that this iterator holds a reference to the type store, could we access the types via the `Rc<TypeStore>`?

src/system.rs
```diff
@@ -43,22 +44,67 @@ impl PendingTypes {
...
-    pub fn update_type_store(&mut self, type_store: &mut TypeStore) -> IonSchemaResult<()> {
+    pub fn update_type_store(
+        &mut self,
+        type_store: &mut TypeStore,
+        load_isl_import: Option<&IslImport>,
```
zslayton, line 50: Could you add a description for the parameters to the doc comment for this function?

src/system.rs
```diff
@@ -43,22 +44,67 @@ impl PendingTypes {
...
-                TypeDefinition::Named(named_type_def) => type_store.add_named_type(named_type_def),
-                TypeDefinition::Anonymous(anonymous_type_def) => {
-                    type_store.add_anonymous_type(anonymous_type_def)
-                }
-            };
+
+            if load_isl_import.is_some() {
+                match type_def.to_owned() {
+                    TypeDefinition::Named(named_type_def) => match load_isl_import.unwrap() {
+                        IslImport::Schema(_) => type_store.add_import_type(named_type_def),
```
zslayton, line 63: I'm not sure I follow the flow here; the `_` in `Schema(_)` should be an ID that leads us to add lots of types to the `type_store`. Instead, we're adding a single named `TypeDefinition` and ignoring the ID?

I think it might help to break this into two functions.

src/system.rs
```diff
@@ -164,21 +210,28 @@ pub type TypeId = usize;
 /// Defines a cache that can be used to store resolved [Type]s of a [Schema]
 #[derive(Debug, Clone)]
 pub struct TypeStore {
+    imports: HashMap<String, TypeId>, // stores all the imported types of a schema
```
zslayton, line 213: `import` is becoming rather overloaded; sometimes it's the grammatical element and sometimes it's a type that was imported. I think this is the latter; if that's the case, it might help to rename this `imported_types` (or `imported_type_ids_by_name` if you want to be verbose but consistent).

src/system.rs
```diff
@@ -164,21 +210,28 @@ pub type TypeId = usize;
...
     ids_by_name: HashMap<String, TypeId>,
```
zslayton, line 214: Now that this _doesn't_ include the imported types, I suggest adding a comment to explain what it does store.

src/system.rs
```diff
@@ -164,21 +210,28 @@ pub type TypeId = usize;
...
             types_by_id: Vec::new(),
         }
     }
 
-    /// Returns [Type]s stored in the [TypeStore] to be used by [SchemaTypeIterator]
-    pub fn get_types(&self) -> &[TypeDefinition] {
-        &self.types_by_id
+    /// Returns [TypeId]s stored in the [TypeStore] to be used by [SchemaTypeIterator]
+    pub fn get_types(&self) -> Vec<TypeId> {
+        self.ids_by_name.values().cloned().collect()
```
zslayton, line 229: Ahh, now I see why you added a `Vec` to the iterator. You could skip that by having this method return an iterator over `self.ids_by_name.values()` and storing that in the `SchemaTypeIterator`. (This would involve adding a lifetime, but I think the blast radius is small?)

src/system.rs
```diff
@@ -164,21 +210,28 @@ pub type TypeId = usize;
...
+    }
+
+    /// Returns import [TypeId]s stored in the [TypeStore] to be used by [SchemaTypeIterator]
+    pub fn get_imports(&self) -> Vec<TypeId> {
+        self.imports.values().cloned().collect()
```
zslayton, line 234: Same thought here RE: iterator/lifetime. If it's not too complicated, we should do it.

src/system.rs
```diff
@@ -214,6 +274,31 @@ impl TypeStore {
...
+        }
+        let type_id = self.types_by_id.len();
+        self.imports.insert(name.to_owned(), type_id);
+        self.types_by_id.push(TypeDefinition::Named(type_def));
+        type_id
+    }
+
+    /// Adds the [NamedTypeDefinition] and the associated alias as the [TypeAlias] import in [TypeStore]
+    ///  and returns the [TypeId] for it. If the name already exists in the [TypeStore] it returns the associated [TypeId]
+    pub fn add_alias_import_type(&mut self, alias: &str, type_def: TypeDefinitionImpl) -> TypeId {
```
zslayton, line 292: How about:
```suggestion
    pub fn add_aliased_import_type(&mut self, alias: &str, type_def: TypeDefinitionImpl) -> TypeId {
```
or, to be consistent with the grammar's element naming:
```suggestion
    pub fn add_import_type_alias(&mut self, alias: &str, type_def: TypeDefinitionImpl) -> TypeId {
```

src/schema.rs
desaikd: Yes. correct. I will modify the comments for this.

src/isl/isl_import.rs
```diff
@@ -0,0 +1,90 @@
...
+        let type_name = match import.get("type") {
+            Some(type_name) => try_to!(type_name.as_str()),
+            None => return Ok(IslImport::Schema(id.to_owned())),
+        };
+
+        let alias = import
+            .get("as")
+            .and_then(|alias| alias.as_str().and_then(|a| Some(a.to_owned())));
+
+        if alias.is_none() {
```
popematt, line 44: minor question—I'm curious; is there any particular reason you used `if` here rather than `match` like the last two conditions?

src/system.rs
```diff
@@ -214,6 +274,31 @@ impl TypeStore {
...
+        if let Some(exists) = self.imports.get(alias) {
+            return exists.to_owned();
+        }
+        let type_id = self.types_by_id.len();
+        self.imports.insert(alias.to_owned(), type_id);
+        self.types_by_id.push(TypeDefinition::Named(type_def));
+        type_id
+    }
```
popematt, line 300: I think you could probably collapse these two methods into one since they are nearly duplicates of each other.
(I think my Rust is right, but you should double check it before accepting this suggestion.)

```suggestion
    /// Adds the [NamedTypeDefinition] and the associated name as the imports of [TypeStore]
    ///  and returns the [TypeId] for it. If the name already exists in the [TypeStore] it returns the associated [TypeId]
    pub fn add_import_type(&mut self, alias: Option<&str>, type_def: TypeDefinitionImpl) -> TypeId {
        let name = alias.or_else( || type_def.name().as_ref().unwrap() );
        if let Some(exists) = self.imports.get(name) {
            return exists.to_owned();
        }
        let type_id = self.types_by_id.len();
        self.imports.insert(name.to_owned(), type_id);
        self.types_by_id.push(TypeDefinition::Named(type_def));
        type_id
    }
```

src/system.rs
```diff
@@ -329,6 +416,7 @@ impl Resolver {
         &mut self,
         id: A,
         type_store: &mut TypeStore,
+        load_isl_import: Option<&IslImport>, // if its the root of schema set it to None otherwise will be set to the IslImport to be loaded
```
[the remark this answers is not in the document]
desaikd: Sorry I should have been more clear on this. I meant:
```rust
// If we are loading the root schema then this will be set to `None` ( i.e. in the beginning when 
// this method is called from the load_schema method of schema_system it is set to `None`)
// Otherwise if we are loading an import of the schema then this will be set to `Some(isl_import)` 
// to be loaded (i.e. Inside schema_from_elements while loading imports this will be set to 
// `Some(isl_import)`)
```
Does this make sense? I can add more clarification in the comments.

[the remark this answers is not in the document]
popematt: Yes, that makes sense.

src/system.rs
popematt: +1 especially to `imported_type_ids_by_name`

src/system.rs
desaikd: Reference issue: https://github.com/amzn/ion-schema-rust/issues/40

src/schema.rs
```diff
@@ -35,6 +35,11 @@ impl Schema {
 
     /// Returns an iterator over the imports of this [Schema].
     fn imports(&self) -> SchemaTypeIterator {
+        todo!()
+    }
+
+    /// Returns an iterator over the imported types of this [Schema].
+    fn import_types(&self) -> SchemaTypeIterator {
```
zslayton, line 42: From the perspective of an end user (who won't care much about ISL's `IMPORT_TYPE` grammar element), I think `imported_types` would fit expectations better:
```suggestion
    fn imported_types(&self) -> SchemaTypeIterator {
```

src/authority.rs
```diff
@@ -46,3 +47,32 @@ impl DocumentAuthority for FileSystemDocumentAuthority {
         Ok(schema_content)
     }
 }
+
+/// An [Authority] implementation that attempts to resolve schema ids to ion elements using the map.
+#[derive(Debug, Clone)]
+pub struct MapDocumentAuthority {
+    ion_content_by_id: HashMap<String, String>, // This map represents (id, ion content) which can used to resolve schema ids to Vec<OwnedElement>
+}
```
zslayton, line 55: Nice! I like that storing this as a `HashMap<String, String>` allows us to write tests that fail because the Ion/ISL that was stored in the map was invalid. Good thinking!

src/schema.rs
```diff
@@ -45,6 +50,7 @@ impl Schema {
     }
 
     /// Returns an iterator over the types in this schema.
+    /// This includes the core types and named types defined within this schema.
```
zslayton, line 53: Were we going to eventually remove the `core` types from this as a `TODO`?

src/system.rs
```diff
@@ -42,12 +42,127 @@ impl PendingTypes {
...
+        import_type_name: &str,
+        type_store: &mut TypeStore,
+    ) -> Option<IonSchemaResult<TypeDefinitionImpl>> {
+        let type_def = self
+            .get_type_id_by_name(import_type_name, type_store)
+            .and_then(|id| self.types_by_id[id].to_owned());
+
+        match type_def {
+            None => None,
+            Some(type_def) => match type_def {
```
zslayton, line 128: This could be one more `and_then` if you like.

src/system.rs
```diff
@@ -42,12 +42,127 @@ impl PendingTypes {
...
+                TypeDefinition::Named(named_type_def) => Some(Ok(named_type_def)),
+                TypeDefinition::Anonymous(_) => Some(unresolvable_schema_error(format!(
+                    "Unable to load import type: {} for schema",
+                    import_type_name
+                ))),
```
zslayton, line 133: If I'm reading this right, this would be the situation in which the user requested a type by name and found an anonymous type in the type store? If that's accurate, I think this is use case for [`unreachable!`](https://doc.rust-lang.org/std/macro.unreachable.html#). We don't expect this to ever happen and there's nothing a user could do with the error if we handed it to them.

src/system.rs
```diff
@@ -42,12 +42,127 @@ impl PendingTypes {
...
+    fn update_type_store_with_all_types(&self, type_store: &mut TypeStore) -> IonSchemaResult<()> {
+        for optional_type in &self.types_by_id {
+            // return an error if any of the type in types_by_id vector is None/Unresolved
+            let type_def = optional_type
+                .to_owned()
+                .ok_or(unresolvable_schema_error_raw(
+                    "Unable to load schema due to unresolvable type",
+                ))?;
+
+            match type_def.to_owned() {
```
zslayton: `to_owned()` is being called twice here; I think we can eliminate one of them.

src/system.rs
```diff
@@ -41,27 +42,160 @@ impl PendingTypes {
...
         }
-        self.types_by_id.clear();
-        self.ids_by_name.clear();
+        Ok(())
+    }
+
+    // helper method to update type store with all the types from this PendingTypes
+    // import_type_name: this argument represents whether the import type is SchemaImport or a TypeImport (includes both TypeImport and TypeAliasImport)
+    //                   None - represents its a schema import which imports all types into imported_type_ids_by_name section of type_store
+    //                   Some(_) - represents a type import which import all types into types_by_id of type_store except specified import type as it will be already loaded by parent method that uses this helper method
```
zslayton, line 161: I'm a bit surprised that `cargo fmt` allowed a line this wide. 🤷

src/system.rs
```diff
@@ -41,27 +42,160 @@ impl PendingTypes {
         }
     }
 
-    /// Adds all the types from Context into given TypeStore and clears Context types for loading next set of types
-    /// this is used after a schema named type/root type is loaded entirely into Context
-    pub fn update_type_store(&mut self, type_store: &mut TypeStore) -> IonSchemaResult<()> {
+    /// Adds all the types from Context into given [TypeStore] including adding all the imported types into imports of [TypeStore].
```
zslayton, line 45: Looks like there are still some dangling references to `Context`.

src/system.rs
```diff
@@ -447,4 +635,171 @@ mod schema_system_tests {
...
+                      type: string,
+                    }
+                    
+                    schema_footer::{
+                    }
+                "#,
+            ),
+        ]
+        .iter()
+        .map(|(id, ion)| (id.to_owned().to_owned(), ion.to_owned().to_owned()))
```
zslayton, line 690: Why the double `to_owned().to_owned()`s?

src/system.rs
```diff
@@ -447,4 +635,171 @@ mod schema_system_tests {
...
+        .collect();
+        let mut schema_system =
+            SchemaSystem::new(vec![Box::new(MapDocumentAuthority::new(map_authority))]);
```
zslayton, line 693: Food for thought: maybe `MapDocumentAuthority`'s constructor should take an `IntoIterator<Item=(&str, &str)>`? Users wouldn't have to call `map`/`collect`, and the implementation detail of which kind of map it's using would stay hidden.

src/system.rs
```diff
@@ -745,10 +741,7 @@ mod schema_system_tests {
                     }
                 "#,
             ),
-        ]
-        .iter()
-        .map(|(id, ion)| (id.to_owned().to_owned(), ion.to_owned().to_owned()))
-        .collect();
```
zslayton: Nice, this really streamlined things.

Issue #10, opened 2021-08-03
Implement all remaining constraints according to Ion Schema specification


desaikd:
Currently there is an implementation for `AllOf` and `Type` Constraint as below:
```rust
/// Implements an `all_of` constraint of Ion Schema
/// [all_of]: https://amzn.github.io/ion-schema/docs/spec.html#all_of
#[derive(Debug, Clone)]
pub struct AllOfConstraint {
    type_references: Vec<TypeRef>,
}

impl AllOfConstraint {
    pub fn new(types: Vec<TypeRef>) -> Self {
        ...
    }

   // other methods to be implemneted for this constraints
}

impl ConstraintValidator for AllOfConstraint {
    fn validate(&self, value: OwnedElement, issues: &mut Violations) {
       ...
    }
}

/// Tries to create an [AllOf] constraint from the given OwnedElement
impl TryFrom<&OwnedElement> for AllOfConstraint {
    type Error = IonSchemaError;

    fn try_from(ion: &OwnedElement) -> Result<Self, Self::Error> {
        ...
    }
}
```

Implement all other constraint using a `struct` and `TryFrom<&OwnedElement>` similar to `AllOf`.

Constraints implementation checklist:

- General Constraints 
  - [x] annotations
  - [x] type
  - [x] valid_values
- Blob/Clob Constraints 
  - [x] byte_length
- String/Symbol Constraints 
  - [x] codepoint_length
  - [x] regex
- Decimal Constraints 
  - [x] precision
  - [x] scale
- Timestamp Constraints 
  - [x] timestamp_offset
  - [x] timestamp_precision
- Container Constraints 
  - [x] container_length
  - [x] content
  - [x] element
- List/S-expression/Document Constraints
  - [x] contains
  - [x] ordered_elements
- Struct Constraints 
  - [x] fields
- Logic Constraints
  - [x] all_of
  - [x] any_of
  - [x] one_of
  - [x] not

desaikd:
Support for  "valid_values" and "regex" are added with `v0.4.0` release.

PR #45, opened 2021-11-12
adds changes for implementation of ISL ranges

desaikd:
*Issue #10*

*Description of changes:*
This PR works on creating ISL ranges.

*Changes:*
* adds changes for ISL `Range`
  *  `RangeBoundaryValue`: represents a range boundary value (i.e. min, max or a value in terms of `RangeBoundaryValueType`)
  * `RangeBoundaryValueType`: represents a range boundary value which can be a `Decimal`/`Integer`/...
  * `RangBoundaryType`: defines exclusivity of the range boundary
* adds a bugfix for unordered types comparison (a new `PartialEq` implementation for `IslTypeImpl`)
* adds unit tests related to ranges

*Test:*
Adds unit tests for ISL `Range`s

### DIFF
--- a/src/isl/isl_type.rs
+++ b/src/isl/isl_type.rs
@@ -124,13 +124,14 @@ impl IslTypeImpl {
 // Related issue: https://github.com/amzn/ion-rust/issues/200
 impl PartialEq for IslTypeImpl {
     fn eq(&self, other: &Self) -> bool {
-        let matching = self.constraints.iter().all(|constraint| {
-            other
-                .constraints
-                .iter()
-                .find(|other_constraint| &constraint == other_constraint)
-                .is_some()
-        });
-        matching && self.constraints.len() == other.constraints.len() && self.name == other.name
+        self.constraints.len() == other.constraints.len()
+            && self.name == other.name
+            && self.constraints.iter().all(|constraint| {
+                other
+                    .constraints
+                    .iter()
+                    .find(|other_constraint| &constraint == other_constraint)
+                    .is_some()
+            })
     }
 }
--- a/src/isl/isl_type.rs
+++ b/src/isl/isl_type.rs
@@ -34,7 +34,7 @@ impl IslType {
 /// Represents both named and anonymous [IslType]s and can be converted to a solid [TypeDefinition] using TypeStore
 /// Named ISL type grammar: `type:: { name: <NAME>, <CONSTRAINT>...}`
 /// Anonymous ISL type grammar: `{ <CONSTRAINT>... }`
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone)]
 pub struct IslTypeImpl {
     name: Option<String>,
     constraints: Vec<IslConstraint>,
@@ -117,5 +117,20 @@ impl IslTypeImpl {
             constraints.push(constraint);
         }
         Ok(IslTypeImpl::new(type_name, constraints))
+    }
+}
+
+// OwnedStruct doesn't preserve field order hence the PartialEq won't work properly for unordered constraints
+// Related issue: https://github.com/amzn/ion-rust/issues/200
+impl PartialEq for IslTypeImpl {
+    fn eq(&self, other: &Self) -> bool {
+        let matching = self.constraints.iter().all(|constraint| {
+            other
+                .constraints
+                .iter()
+                .find(|other_constraint| &constraint == other_constraint)
+                .is_some()
+        });
+        matching && self.constraints.len() == other.constraints.len() && self.name == other.name
     }
 }
--- a/src/isl/mod.rs
+++ b/src/isl/mod.rs
@@ -132,6 +132,7 @@ mod isl_tests {
     use ion_rs::types::timestamp::Timestamp;
     use ion_rs::value::reader::element_reader;
     use ion_rs::value::reader::ElementReader;
+    use ion_rs::value::AnyInt;
     use ion_rs::IonType;
     use rstest::*;
 
@@ -248,9 +249,9 @@ mod isl_tests {
                     range::[min, 5]
                 "#
             ),
-            Range::int_range(
+            Range::range(
                 RangeBoundaryValue::Min,
-                RangeBoundaryValue::int_value(5, RangeBoundaryType::Inclusive)
+                RangeBoundaryValue::int_value(AnyInt::I64(5), RangeBoundaryType::Inclusive)
             )
         ),
         case::range_with_float(
@@ -259,7 +260,7 @@ mod isl_tests {
                     range::[2e1, 5e1]
                 "#
             ),
-            Range::float_range(
+            Range::range(
                 RangeBoundaryValue::float_value(2e1, RangeBoundaryType::Inclusive),
                 RangeBoundaryValue::float_value(5e1, RangeBoundaryType::Inclusive)
             )
@@ -270,7 +271,7 @@ mod isl_tests {
                     range::[20.4, 50.5]
                 "#
             ),
-            Range::decimal_range(
+            Range::range(
                 RangeBoundaryValue::decimal_value(Decimal::new(204, -1), RangeBoundaryType::Inclusive),
                 RangeBoundaryValue::decimal_value(Decimal::new(505, -1), RangeBoundaryType::Inclusive)
             )
@@ -281,7 +282,7 @@ mod isl_tests {
                     range::[2020-01-01T, 2021-01-01T]
                 "#
             ),
-            Range::timestamp_range(
+            Range::range(
                 RangeBoundaryValue::timestamp_value(Timestamp::with_year(2020).with_month(1).with_day(1).build().unwrap(), RangeBoundaryType::Inclusive),
                 RangeBoundaryValue::timestamp_value(Timestamp::with_year(2021).with_month(1).with_day(1).build().unwrap(), RangeBoundaryType::Inclusive)
             )
@@ -311,6 +312,11 @@ mod isl_tests {
         case::range_with_min_upper_bound(load_range(
             r#"
                 range::[5, min]
+            "#
+        )),
+        case::range_with_mismatched_bounds(load_range(
+            r#"
+                range::[5, 7.834]
             "#
         )),
         // TODO: uncomment below test case once we have a comparator for timestamp in ion-rust 
--- a/src/isl/mod.rs
+++ b/src/isl/mod.rs
@@ -80,6 +80,7 @@ pub mod isl_constraint;
 pub mod isl_import;
 pub mod isl_type;
 pub mod isl_type_reference;
+mod util;
 
 /// Provides an internal representation of an schema file
 #[derive(Debug, Clone)]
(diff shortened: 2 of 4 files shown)


src/isl/isl_type.rs
```diff
@@ -119,3 +119,18 @@ impl IslTypeImpl {
         Ok(IslTypeImpl::new(type_name, constraints))
     }
 }
+
+// OwnedStruct doesn't preserve field order hence the PartialEq won't work properly for unordered constraints
+// Related issue: https://github.com/amzn/ion-rust/issues/200
+impl PartialEq for IslTypeImpl {
```
popematt, line 125: non-blocking question—can you explain why we're using `PartialEq` instead of `Eq` for all of the types in this library? As far as I can tell, the only difference is reflexivity, and I can't think of any case for ISL where `a != a`. (I know that `nan` exists and `nan != nan`, but in the context of ISL, something like `valid_values: [ X ]` has reflexive equality even if `X` is `nan`.)

src/isl/util.rs
```diff
@@ -0,0 +1,286 @@
...
+///                       | range::[ <EXCLUSIVITY><RANGE_TYPE>, max ]
+/// Grammar: <RANGE_TYPE> ::= <DECIMAL>
+///                | <FLOAT>
+///                | <INT>
+///                | <NUMBER>
+///                | <TIMESTAMP>
+///                | <TIMESTAMP_PRECISION_VALUE>
+/// For more information on [Range]: <https://amzn.github.io/ion-schema/docs/spec.html#constraints>
+#[derive(Debug, Clone, PartialEq)]
+pub enum Range {
```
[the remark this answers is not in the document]
desaikd: Yes, I will create an issue for it.

src/isl/mod.rs
```diff
@@ -223,4 +228,100 @@ mod isl_tests {
...
+    fn owned_struct_to_range(range1: IonSchemaResult<Range>, range2: IonSchemaResult<Range>) {
+        // determine that both the ranges are created with no errors
+        assert_eq!(range1.is_ok(), true);
+        assert_eq!(range2.is_ok(), true);
+
+        // assert if both the ranges are same
+        assert_eq!(range1.unwrap(), range2.unwrap());
+    }
+
+    #[rstest(
```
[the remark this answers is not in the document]
desaikd: I can add the tests for mismatch types but for `range::` annotation I feel it should be added wherever range is actually used (e.g. `byte_length: range::[2,5]`).

src/isl/util.rs
```diff
@@ -0,0 +1,286 @@
...
+                value.as_timestamp().unwrap().to_owned(),
+                range_boundary_type,
+            )),
+            _ => return invalid_schema_error("Unsupported range type specified"),
+        }
+    }
+}
+
+/// Represents the range boundary types in terms of exclusivity (i.e. inclusive or exclusive)
+#[derive(Debug, Clone, PartialEq, PartialOrd)]
```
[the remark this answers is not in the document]
desaikd: I had added this because of the comparison to `start > end` validation but as it depends on timestamp ordering which is not yet implemented, I can remove it for now.

src/isl/mod.rs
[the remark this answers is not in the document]
popematt: Okay

src/isl/util.rs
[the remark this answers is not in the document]
popematt: If it's something that we for sure need later, don't remove it now just because I don't understand it.

The `start > end` validation relies on comparing `RangeBoundaryValue`s or possibly `RangeBoundaryValueType`s. I understand why the `RangeBoundaryValueType` should have `PartialOrd`, but I don't understand why `RangeBoundaryType` has any ordering at all. Is `Inclusive` greater than `Exclusive` or is it the other way around?

src/isl/util.rs
```diff
@@ -0,0 +1,286 @@
...
+        let range_type = Range::validate_range_boundaries(&start, &end)?;
+        match range_type {
+            RangeBoundaryValueType::Decimal(_) => Ok(Range::Decimal(start, end)),
+            RangeBoundaryValueType::Float(_) => Ok(Range::Float(start, end)),
+            RangeBoundaryValueType::Integer(_) => Ok(Range::Integer(start, end)),
+            RangeBoundaryValueType::IntegerNonNegative(_) => {
+                Ok(Range::IntegerNonNegative(start, end))
+            }
+            RangeBoundaryValueType::Timestamp(_) => Ok(Range::Timestamp(start, end)),
+        }
```
zslayton, line 92: If you make a consolidated `range` function, this would just be `range(start, end)`.

src/isl/util.rs
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    // helper method to which validates a non negative integer range boundary value
+    fn validate_non_negative_integer_range_boundary_value(
+        value: &AnyInt,
+    ) -> IonSchemaResult<usize> {
+        match value.as_i64() {
+            Some(v) => {
+                if v > 0 {
```
zslayton, line 101: I think this should be `>=`?

src/isl/util.rs
```diff
@@ -0,0 +1,286 @@
...
+                Ok(Range::IntegerNonNegative(start, end))
+            }
+            RangeBoundaryValueType::Timestamp(_) => Ok(Range::Timestamp(start, end)),
+        }
+    }
+
+    // helper method to which validates a non negative integer range boundary value
+    fn validate_non_negative_integer_range_boundary_value(
+        value: &AnyInt,
+    ) -> IonSchemaResult<usize> {
```
zslayton, line 98: The size of `usize` varies by platform and is usually used for addresses in memory or offsets. On some platforms, this would end up being a 4-byte integer.

Should we just return the same `AnyInt` so we can continue to support values of arbitrary size?

src/isl/util.rs
```diff
@@ -0,0 +1,286 @@
...
+    }
+    pub fn int_non_negative_value(value: usize, range_boundary_type: RangeBoundaryType) -> Self {
+        RangeBoundaryValue::Value(
+            RangeBoundaryValueType::IntegerNonNegative(value),
+            range_boundary_type,
+        )
+    }
+    pub fn float_value(value: f64, range_boundary_type: RangeBoundaryType) -> Self {
+        RangeBoundaryValue::Value(RangeBoundaryValueType::Float(value), range_boundary_type)
+    }
```
zslayton, line 214: Similar TODO thought for these: I think we could have a consolidated `value` method?

src/isl/util.rs
```diff
@@ -0,0 +1,286 @@
...
+
+    pub fn timestamp_range(
+        start: RangeBoundaryValue,
+        end: RangeBoundaryValue,
+    ) -> IonSchemaResult<Range> {
+        Range::validate_range_boundaries(&start, &end)?;
+        Ok(Range::Timestamp(start, end))
+    }
+
+    pub fn from_ion_element(value: &OwnedElement, is_non_negative: bool) -> IonSchemaResult<Range> {
```
zslayton, line 68: The `is_non_negative` flag feels a bit unfortunate to me. It's only relevant in the integer case to distinguish between `Integer` and `IntegerNonNegative` but has to be specified no matter which type is expected. It also has to be passed through to each of the nested calls to other `from_ion_element` implementations.

Looking at [the grammar](https://amzn.github.io/ion-schema/docs/spec.html#grammar), it seems like `RANGE_TYPE` doesn't include a `NonNegativeInteger` variant: 
```
<RANGE_TYPE> ::= <DECIMAL>
               | <FLOAT>
               | <INT>
               | <NUMBER>
               | <TIMESTAMP>
               | <TIMESTAMP_PRECISION_VALUE>
```

Instead, it [creates a type](https://amzn.github.io/ion-schema/docs/spec.html#byte_length-constraint) to model this constraint:

```
type::{
  name: range_int_non_negative,
  type: list,
  annotations: required::[range],
  ordered_elements: [
    { one_of: [ int_non_negative, { valid_values: [min] } ] },
    { one_of: [ int_non_negative, { valid_values: [max] } ] },
  ],
  container_length: 2,
}
```

I'm inclined to say we should follow suit here. We would remove the `NonNegativeInteger` variant from the `RangeBoundaryValueType` and model this as a type instead. `from_ion_element` would then be able to infer everything it needs from the Ion data itself.

src/isl/util.rs
desaikd: If I understand this correctly, you are suggesting we should represent both `integer` and `non_negative_integers` as `AnyInt`. I think that should be okay as we already annotate range boundaries with it's type.

src/isl/util.rs
```diff
@@ -0,0 +1,286 @@
...
+    Timestamp(RangeBoundaryValue, RangeBoundaryValue),
+}
+
+impl Range {
+    pub fn decimal_range(
+        start: RangeBoundaryValue,
+        end: RangeBoundaryValue,
+    ) -> IonSchemaResult<Range> {
+        Range::validate_range_boundaries(&start, &end)?;
+        Ok(Range::Decimal(start, end))
```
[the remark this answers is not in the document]
desaikd: This is verified inside `validate_range_boundaries` when both the boundaries are `Value`s.

src/isl/util.rs
desaikd: I have created an issue for this to change it after I have ordered_elements implemented as a constraint. (Reference issue: https://github.com/amzn/ion-schema-rust/issues/46)

src/isl/util.rs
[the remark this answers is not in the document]
desaikd: Reference issue: https://github.com/amzn/ion-schema-rust/issues/47

src/isl/util.rs
```diff
@@ -0,0 +1,286 @@
...
+        Ok(Range::IntegerNonNegative(start, end))
+    }
+
+    pub fn timestamp_range(
+        start: RangeBoundaryValue,
+        end: RangeBoundaryValue,
+    ) -> IonSchemaResult<Range> {
+        Range::validate_range_boundaries(&start, &end)?;
+        Ok(Range::Timestamp(start, end))
+    }
```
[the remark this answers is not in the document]
desaikd: Reference issue: https://github.com/amzn/ion-schema-rust/issues/48

src/isl/util.rs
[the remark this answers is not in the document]
desaikd: I have also added unit test for this as popematt@ suggested in his comments.

src/isl/util.rs
```diff
@@ -79,33 +91,18 @@ impl Range {
...
-        }
+        // validate both range boundary values and returns created `Range`
+        Range::range(start, end)
     }
 
     // helper method to which validates a non negative integer range boundary value
     fn validate_non_negative_integer_range_boundary_value(
         value: &AnyInt,
-    ) -> IonSchemaResult<usize> {
+    ) -> IonSchemaResult<&AnyInt> {
```
zslayton, line 101: Does this still need to return a value?

zslayton (approved):

src/isl/util.rs
desaikd: No. I can remove it and just return `IonSchemaResult<()>` instead.

Issue #11, opened 2021-08-03
Implement `Violation` for constraints


desaikd:
Implement violation which contains all the errors/violations found during the validation of a `Constraint`.
``` rust
#[derive(Debug, Clone)]
pub struct Violations {
    violations: Vec<Violation>,
}

impl Violations {
    fn add_violation(violation: Violation) {
        todo!()
    }
}

#[derive(Debug, Clone)]
pub struct Violation {}
```

desaikd:
This has been implemented with #49

Issue #9, opened 2021-08-03
Implement `validate()` for all constraints


desaikd:
Currently there is a `ConstraintValidator` trait that has validate method for validating a value according to the constraint's validation logic. Implement `validate()` for all constraints which then can be used to validate a type.

```rust
/// Provides validation for schema Constraint
pub trait ConstraintValidator {
    /// Checks this constraint against the provided value,
    /// adding [Violation]s and/or [ViolationChild]ren to issues
    /// if the constraint is violated.
    fn validate(&self, value: OwnedElement, issues: &mut Violations);
}
```

Validation implementation checklist:

- General Constraints 
  - [x] annotations
  - [x] type
  - [x] valid_values
- Blob/Clob Constraints 
  - [x] byte_length
- String/Symbol Constraints 
  - [x] codepoint_length
  - [x] regex
- Decimal Constraints 
  - [x] precision
  - [x] scale
- Timestamp Constraints 
  - [x] timestamp_offset
  - [x] timestamp_precision
- Container Constraints 
  - [x] container_length
  - [x] content
  - [x] element
- List/S-expression/Document Constraints
  - [x] contains
  - [x] ordered_elements
- Struct Constraints 
  - [x] fields
- Logic Constraints
  - [x] all_of
  - [x] any_of
  - [x] one_of
  - [x] not

PR #49, opened 2021-11-24
adds changes for validation of logic constraints

desaikd:
*Issues: #11, #9*

*Description of changes:*
This PR works on changes for validation of logic constraints

*Changes:*
* adds Violations implementation to be used for specifying errors while
validation
* adds changes for storing core_types separately into type_store(this
simplifies the validation for core types)
* filled validate method for all logic constraints and `TypeDefinition`

*Test:*
adds unit tests based on validation of logic constraints

### DIFF
--- a/src/constraint.rs
+++ b/src/constraint.rs
@@ -11,7 +11,7 @@ pub trait ConstraintValidator {
     /// Checks this constraint against the provided value,
     /// adding [Violation]s and/or [ViolationChild]ren to issues
     /// if the constraint is violated.
-    fn validate(&self, value: OwnedElement, issues: &mut impl Violations, type_store: &TypeStore);
+    fn validate(&self, value: &OwnedElement, issues: &mut impl Violations, type_store: &TypeStore);
 }
 
 /// Defines schema Constraints
@@ -104,7 +104,7 @@ impl Constraint {
 
     pub fn validate(
         &self,
-        value: OwnedElement,
+        value: &OwnedElement,
         issues: &mut impl Violations,
         type_store: &TypeStore,
     ) {
@@ -147,13 +147,13 @@ impl AllOfConstraint {
 }
 
 impl ConstraintValidator for AllOfConstraint {
-    fn validate(&self, value: OwnedElement, issues: &mut impl Violations, type_store: &TypeStore) {
+    fn validate(&self, value: &OwnedElement, issues: &mut impl Violations, type_store: &TypeStore) {
         let mut valid_types = vec![];
         let mut all_of_violation = ViolationImpl::new("all_of", "all_types_not_matched", "");
         for type_id in &self.type_ids {
             let type_def = type_store.get_type_by_id(*type_id).unwrap();
             let checkpoint = all_of_violation.checkpoint();
-            type_def.validate(&value, &mut all_of_violation, type_store);
+            type_def.validate(value, &mut all_of_violation, type_store);
             if checkpoint.is_valid(&all_of_violation) {
                 valid_types.push(type_id);
             }
@@ -196,7 +196,7 @@ impl AnyOfConstraint {
 }
 
 impl ConstraintValidator for AnyOfConstraint {
-    fn validate(&self, value: OwnedElement, issues: &mut impl Violations, type_store: &TypeStore) {
+    fn validate(&self, value: &OwnedElement, issues: &mut impl Violations, type_store: &TypeStore) {
         let mut valid_types = vec![];
         let mut any_of_violation = ViolationImpl::new(
             "any_of",
@@ -206,7 +206,7 @@ impl ConstraintValidator for AnyOfConstraint {
         for type_id in &self.type_ids {
             let type_def = type_store.get_type_by_id(*type_id).unwrap();
             let checkpoint = any_of_violation.checkpoint();
-            type_def.validate(&value, &mut any_of_violation, type_store);
+            type_def.validate(value, &mut any_of_violation, type_store);
             if checkpoint.is_valid(&any_of_violation) {
                 valid_types.push(type_id);
             }
@@ -244,13 +244,13 @@ impl OneOfConstraint {
 }
 
 impl ConstraintValidator for OneOfConstraint {
-    fn validate(&self, value: OwnedElement, issues: &mut impl Violations, type_store: &TypeStore) {
+    fn validate(&self, value: &OwnedElement, issues: &mut impl Violations, type_store: &TypeStore) {
         let mut valid_types = vec![];
         let mut one_of_violation = ViolationImpl::new("one_of", "", "");
         for type_id in &self.type_ids {
             let type_def = type_store.get_type_by_id(*type_id).unwrap();
             let checkpoint = one_of_violation.checkpoint();
-            type_def.validate(&value, &mut one_of_violation, type_store);
+            type_def.validate(value, &mut one_of_violation, type_store);
             if checkpoint.is_valid(&one_of_violation) {
                 valid_types.push(type_id);
             }
@@ -294,11 +294,11 @@ impl NotConstraint {
 }
 
 impl ConstraintValidator for NotConstraint {
-    fn validate(&self, value: OwnedElement, issues: &mut impl Violations, type_store: &TypeStore) {
+    fn validate(&self, value: &OwnedElement, issues: &mut impl Violations, type_store: &TypeStore) {
         let type_def = type_store.get_type_by_id(self.type_id).unwrap();
         let mut not_violation =
             ViolationImpl::new("not", "type_matched", "value unexpectedly matches type");
-        type_def.validate(&value, &mut not_violation, type_store);
+        type_def.validate(value, &mut not_violation, type_store);
         if not_violation.is_valid() {
             issues.add_violation(not_violation);
         }
@@ -330,8 +330,8 @@ impl TypeConstraint {
 }
 
 impl ConstraintValidator for TypeConstraint {
-    fn validate(&self, value: OwnedElement, issues: &mut impl Violations, type_store: &TypeStore) {
+    fn validate(&self, value: &OwnedElement, issues: &mut impl Violations, type_store: &TypeStore) {
         let type_def = type_store.get_type_by_id(self.type_id).unwrap();
-        type_def.validate(&value, issues, type_store);
+        type_def.validate(value, issues, type_store);
     }
 }
--- a/src/constraint.rs
+++ b/src/constraint.rs
@@ -2,7 +2,8 @@ use crate::isl::isl_constraint::IslConstraint;
 use crate::isl::isl_type_reference::IslTypeRef;
 use crate::result::IonSchemaResult;
 use crate::system::{PendingTypes, TypeId, TypeStore};
-use crate::violation::Violations;
+use crate::types::TypeValidator;
+use crate::violation::{ViolationImpl, Violations};
 use ion_rs::value::owned::OwnedElement;
 
 /// Provides validation for schema Constraint
@@ -10,7 +11,7 @@ pub trait ConstraintValidator {
     /// Checks this constraint against the provided value,
     /// adding [Violation]s and/or [ViolationChild]ren to issues
     /// if the constraint is violated.
-    fn validate(&self, value: OwnedElement, issues: &mut Violations);
+    fn validate(&self, value: OwnedElement, issues: &mut impl Violations, type_store: &TypeStore);
 }
 
 /// Defines schema Constraints
@@ -100,6 +101,23 @@ impl Constraint {
             }
         }
     }
+
+    pub fn validate(
+        &self,
+        value: OwnedElement,
+        issues: &mut impl Violations,
+        type_store: &TypeStore,
+    ) {
+        match self {
+            Constraint::AllOf(all_of) => all_of.validate(value, issues, type_store),
+            Constraint::AnyOf(any_of) => any_of.validate(value, issues, type_store),
+            Constraint::Not(not) => not.validate(value, issues, type_store),
+            Constraint::OneOf(one_of) => one_of.validate(value, issues, type_store),
+            Constraint::Type(type_constraint) => {
+                type_constraint.validate(value, issues, type_store)
+            }
+        }
+    }
 }
 
 /// Implements an `all_of` constraint of Ion Schema
@@ -129,8 +147,25 @@ impl AllOfConstraint {
 }
 
 impl ConstraintValidator for AllOfConstraint {
-    fn validate(&self, value: OwnedElement, issues: &mut Violations) {
-        todo!()
+    fn validate(&self, value: OwnedElement, issues: &mut impl Violations, type_store: &TypeStore) {
+        let mut valid_types = vec![];
+        let mut all_of_violation = ViolationImpl::new("all_of", "all_types_not_matched", "");
+        for type_id in &self.type_ids {
+            let type_def = type_store.get_type_by_id(*type_id).unwrap();
+            let checkpoint = all_of_violation.checkpoint();
+            type_def.validate(&value, &mut all_of_violation, type_store);
+            if checkpoint.is_valid(&all_of_violation) {
+                valid_types.push(type_id);
+            }
+        }
+        if valid_types.len() != self.type_ids.len() {
+            all_of_violation.message = format!(
+                "value matches {} types, expected {}",
+                valid_types.len(),
+                self.type_ids.len()
+            );
+            issues.add_violation(all_of_violation)
+        }
     }
 }
 
@@ -161,8 +196,24 @@ impl AnyOfConstraint {
 }
 
 impl ConstraintValidator for AnyOfConstraint {
-    fn validate(&self, value: OwnedElement, issues: &mut Violations) {
-        todo!()
+    fn validate(&self, value: OwnedElement, issues: &mut impl Violations, type_store: &TypeStore) {
+        let mut valid_types = vec![];
+        let mut any_of_violation = ViolationImpl::new(
+            "any_of",
+            "no_types_matched",
+            "value matches none of the types",
+        );
+        for type_id in &self.type_ids {
+            let type_def = type_store.get_type_by_id(*type_id).unwrap();
+            let checkpoint = any_of_violation.checkpoint();
+            type_def.validate(&value, &mut any_of_violation, type_store);
+            if checkpoint.is_valid(&any_of_violation) {
+                valid_types.push(type_id);
+            }
+        }
+        if valid_types.len() == 0 {
+            issues.add_violation(any_of_violation)
+        }
     }
 }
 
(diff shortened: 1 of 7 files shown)


src/violation.rs
```diff
@@ -1,15 +1,98 @@
...
+    fn add_violation(&mut self, violation: ViolationImpl);
+
+    /// Returns `true` if no violations were found, otherwise `false`
+    fn is_valid(&self) -> bool;
+
+    /// Returns the [Violation]s that were found during validation
+    fn violations(&self) -> &[ViolationImpl];
+
+    /// Creates a checkpoint to track the number of violations from this point onwards
+    fn checkpoint(&self) -> Checkpoint;
```
popematt, line 13: question—If the `validate` functions returned a `Option<Violation>` rather than using side-effects, we wouldn't need the `checkpoint` function, and it _seems_ like the net result would be a simpler implementation. Do you know if it's possible to do this, and if it would actually result in a simpler or more complicated implementation?

src/violation.rs
```diff
@@ -1,15 +1,98 @@
-// TODO: Implement Violations for constraints
-#[derive(Debug, Clone)]
-pub struct Violations {
-    violations: Vec<Violation>,
+/// Represents Violations found during validation of values based on schema types
+pub trait Violations {
```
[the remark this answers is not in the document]
desaikd: Sure. I can rename these.

src/violation.rs
desaikd: I feel it might complicate this if I use an `Option<Violation>` here because at the end we need an accumulated result of all `Violation`s and if every time a new `Violation` is created then it needs to be added to the `Violations` object after its returned. Also, for container type validations the children violations needs to be added as well.

src/schema.rs
```diff
@@ -197,4 +213,92 @@ mod schema_tests {
...
+            load(r#"
+                false
+                "hello"
+                hey
+            "#),
+            load_schema_from_text(r#" // For a schema with one_of constraint as below: 
+                type:: { name: one_of_type, one_of: [int, decimal] }
+            "#),
+        ),
+        // TODO: add a test case for all_od constraint
```
zslayton, line 266:
```suggestion
        // TODO: add a test case for all_of constraint
```

src/schema.rs
```diff
@@ -197,4 +213,92 @@ mod schema_tests {
...
+                type:: { name: any_of_type, any_of: [int, decimal, bool] }
+            "#),
+        ),
+    )]
+    fn type_validation(
+        valid_values: Vec<OwnedElement>,
+        invalid_values: Vec<OwnedElement>,
+        schema: Rc<Schema>,
+    ) {
+        let types: Vec<TypeRef> = schema.get_types().collect();
```
zslayton, line 288: Making a `Vec` and then referring to `types[0]` is a little less clear to me than:
```rust
// There is only a single type in each schema defined above.
let type_ref: TypeRef = schema.get_types().first().expect("Loaded schema was empty.");
```

src/system.rs
```diff
@@ -135,6 +138,12 @@ impl PendingTypes {
                             import_type_name
                         )
                     }
+                    TypeDefinition::Core(_) => {
+                        unreachable!(
+                            "The TypeDefinition for the imported type '{}' was core type.",
```
zslayton, line 143:
```suggestion
                            "The TypeDefinition for the imported type '{}' was a core type.",
```

src/system.rs
```diff
@@ -145,6 +154,12 @@ impl PendingTypes {
...
```
zslayton, line 159:
```suggestion
                            "The TypeDefinition for the imported type '{}' was a core type.",
```

src/system.rs
```diff
@@ -253,6 +272,23 @@ impl PendingTypes {
         type_id + type_store.types_by_id.len()
     }
 
+    /// Adds the [CoreTypeDefinition] and the associated ion_type in the [PendingTypes] and returns the [TypeId] for it
+    /// If the given name already exists in the [TypeStore] or [PendingTypes] it returns the associated [TypeId]
+    pub fn add_core_type(&mut self, ion_type: IonType, type_store: &mut TypeStore) -> TypeId {
+        let core_type = &format!("{:?}", ion_type);
```
zslayton, line 278: We probably shouldn't take a direct dependency on whatever format `ion-rust` uses for `IonType`'s `Debug` implementation. While it's not likely to change, it reserves the right to.

Why not make the map a `BTreeMap<IonType, TypeId>`? (We could change it to a `HashMap` after `IonType` is tweaked to derive `Hash`.)

src/system.rs
```diff
@@ -360,6 +398,13 @@ impl TypeStore {
             .or_else(|| self.imported_type_ids_by_name.get(name))
     }
 
+    /// Provides the [TypeId] associated with given [IonType] if it exists in the [TypeStore]  
+    /// Otherwise returns None
+    pub fn get_core_type_id_by_ion_type(&self, ion_type: &IonType) -> Option<&TypeId> {
+        self.core_type_ids_by_ion_type
+            .get(&format!("{:?}", ion_type))
```
zslayton, line 405: It would also be nice to avoid allocating/formatting the text each time we want to look up a core type; they're likely to be commonly referenced. Storing the `IonType` instead of a `String` will address this.

src/types.rs
```diff
@@ -30,13 +32,21 @@ impl TypeRef {
     pub fn new(id: TypeId, type_store: Rc<TypeStore>) -> Self {
         Self { id, type_store }
     }
+
+    pub fn validate(&self, value: OwnedElement, issues: &mut impl Violations) {
+        let type_def = self.type_store.get_type_by_id(self.id).unwrap();
+        for constraint in type_def.constraints() {
+            constraint.validate(value.to_owned(), issues, &self.type_store);
```
zslayton, line 39: Why does `validate` consume `value`? (Said another way: why can't `value` be a `&OwnedElement` instead of an `OwnedElement`?) We shouldn't need to clone the element (with `to_owned()`) each time we recurse.

src/violation.rs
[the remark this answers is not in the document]
zslayton: > question—Is it possible to make these into immutable, plain-old data objects?

I second this question. Making this a trait suggests that we intend to have several implementations of the `Violations` collections, but I'm not sure I understand why we'd do that. I expected `Violations` to be a simple wrapper around `Vec<Violation>`. Having this be a trait requires the `validate` method to be generic by accepting `&mut impl Violations` as a parameter.

src/violation.rs
```diff
@@ -1,15 +1,98 @@
...
+    /// adds a violation into the collection of [Violation]s
+    fn add_violation(&mut self, violation: ViolationImpl);
+
+    /// Returns `true` if no violations were found, otherwise `false`
+    fn is_valid(&self) -> bool;
+
+    /// Returns the [Violation]s that were found during validation
+    fn violations(&self) -> &[ViolationImpl];
+
+    /// Creates a checkpoint to track the number of violations from this point onwards
```
zslayton, line 12: This could use a bit more explanation. I'm not sure what a checkpoint is in this context.

src/violation.rs
zslayton: @popematt Do we know why `ion-schema-kotlin` tries to report all possible violations rather than just the first one?

src/violation.rs
```diff
@@ -1,15 +1,98 @@
...
+
+    fn is_valid(&self) -> bool {
+        self.violations.is_empty()
+    }
+
+    fn violations(&self) -> &[ViolationImpl] {
+        &self.violations
+    }
+
+    fn checkpoint(&self) -> Checkpoint {
```
zslayton, line 40: Why not making this `fn len(&self) -> usize` for simplicity?

EDIT: I see below that this is a placeholder for more complicated logic.

src/violation.rs
```diff
@@ -1,15 +1,98 @@
...
+        Checkpoint::new(self.violations.len())
+    }
+}
+
+// TODO: this can be extended to store children violation count as well for container constraints
+// Represents a checkpoint to be used while validation
+// This determines if children or nested types/constraints have the same number of violations since the checkpoint was created
+pub struct Checkpoint {
```
zslayton, line 48: Is this how `ion-schema-kotlin` does this? I'm wondering what the complexity will buy us.

src/violation.rs
```diff
@@ -1,15 +1,98 @@
...
+
+    pub fn is_valid(&self, violation: &ViolationImpl) -> bool {
+        self.violations_count == violation.violations.len()
+    }
+}
+
+// Represents a single Violation with detailed error message, error code and the constraint for which the validation failed
 #[derive(Debug, Clone)]
-pub struct Violation {}
+pub struct ViolationImpl {
```
zslayton, line 64: To me, this name suggests that there is a trait called `Violation`. I'd stick with `pub struct Violation`.

src/system.rs
desaikd: I had mainly used the string format because we don't have an implementation of `Hash` for `IonType` to be used with `HashMap` but I can try to use `BTreeMap` and see if it works here.

src/types.rs
desaikd: Yes I will change this to be `&OwnedElement` instead.

src/violation.rs
popematt: Basically, we don't think there's ever a use case for knowing only one of the violations. Uses cases fall into two categories:
1. You only want to find out whether the data is valid or not. If this is your use case, then you want _none_ of the violations—just a boolean answer. In `ion-schema-kotlin` you can call `isValid()` for this.
2. You want to explain why the data is invalid. If this is your use case, then you should want _all_ of the violations at once. Firstly, because it's really annoying to work with an API that will only expose 1 error at a time as you're trying to fix things. Secondly, because sometimes you need more than one (child) violation to explain what is wrong. (Eg. the `one_of` constraint when it matches two types.)

That being said, there is really only ever one top-level violation ("value does not match type ___"), and the individual reasons why it does not match are children of that top level violation. However, `ion-schema-kotlin` models violations in (what I believe to be) a confusing way. 
* The object returned by `validate()` is a `Violations`.
* `Violations` has no message of its own and is an aggregation of `Violation`. 
* `Violation` extends `Violations`, but has a message, and aggregates `Violation` and `ViolationChild`. It represents a specific reason why the data is invalid.
* `ViolationChild` extends `Violations`, but has a message and optionally a field name or list index, and aggregates `Violation` and `ViolationChild`. It represents a specific location in the data that is invalid.

So, `ion-schema-kotlin` attempts to model violations as a forest when it should probably be a tree.

src/violation.rs
popematt: Yes, this is what `ion-schema-kotlin` does, but it's only needed because `ion-schema-kotlin` relies on side-effects to produce the complete tree of violations. The logical constraints `one_of`, `any_of`, and `all_of` use the check point to see if there are any violations for each type that they test. 

Simplified version of what `validate` does in the logical constraints:
```
val validTypes = mutableListOf<Type>()
types.forEach { t ->
    val checkpoint = violation.checkpoint()
    t.validate(value, violation)
    if (checkpoint.isValid()) {
        validTypes.add(it())
    }
}
// And then more logic to handle checking the number of matching types in `validTypes`
```
The `Checkpoint` type is a tuple of the sizes of `violations` and `children` of the current `Violation`. If there was only a single list of child objects (instead of `violations` and `children`) we could replace `Checkpoint` with an `Int`.

If we could do could make `validate` a pure function, it would be even better:
```
// Assume we have a kotlin sealed class like Rust Option type.
// typeResults is inferred to be Map<Type, Option<Violation>>

val typeResults = types.associateWith { t -> t.validate(value) }
val validTypes = typeResults.filterValues { it is Option.None }.keys
```

IMO, the complexity of `Checkpoint` is not worth it.

src/violation.rs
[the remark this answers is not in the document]
desaikd: Because the types can have constraints which can have type references, removing `trait` `Violations` would mean, I will not be able to pass `ViolationsImpl`(`ViolationTree`: which has no message) and `ViolationImpl`(`ViolationLeaf`: which has more details on where exactly the issue was found through message) interchangeably. The calls to validate method of constraints/types should expect the same struct/trait hence I have Violations as a trait. Is there any other way that you suggest to use Violations here? @zslayton @popematt.

src/system.rs
desaikd: I have tried using `BTreeMap` although it does require `Ord` implmentation for `IonType` which isn't there currently. Hence, I have still kept everything as it is using `String` instead of `IonType` for the `HashMap`. 
Reference issue: #50

src/system.rs
desaikd: Reference issue: #50

src/violation.rs
desaikd: Renamed to `ViolationTree` and `VioaltionLeaf`.

src/violation.rs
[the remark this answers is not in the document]
popematt: > Having this be a trait requires the validate method to be generic by accepting &mut impl Violations as a parameter.

If `validate` is a pure function (as I suggest elsewhere), then it shouldn't have to accept `Violations` as a parameter. Types and Constraints should be able to have a signature something like this:

```rust
fn validate(&self, value: &OwnedElement) -> Option<Violation>
```

> I will not be able to pass ViolationsImpl(ViolationTree: which has no message) and ViolationImpl(ViolationLeaf: which has more details on where exactly the issue was found through message)

I don't think it's worth keeping around any Violations implementation that has no message. We can simplify violations by having only one type of `Violation` that can have 0..many children and can be used for the root, interior, and leaf nodes of the tree.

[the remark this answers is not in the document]
desaikd: I understand it now. Yes. In that case I will remove `trait Violations` and just keep single `struct Violation` which can have a message as well as `Vec<Violation>` (describing children violations)

src/violation.rs
desaikd: Removed `Checkpoint` struct using the pure validate function which return `Option<Violation>`.

src/types.rs
```diff
@@ -55,16 +65,42 @@ impl TypeDefinition {
...
             TypeDefinition::Named(named_type) => named_type.constraints(),
             TypeDefinition::Anonymous(anonymous_type) => anonymous_type.constraints(),
+            _ => &[],
         }
     }
 }
 
+impl TypeValidator for TypeDefinition {
+    fn is_valid(&self, value: &OwnedElement) -> bool {
+        todo!()
```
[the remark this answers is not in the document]
desaikd: added implementation of `is_valid` based on `validate` function

popematt (commented):
I was talking to someone else about Rust who reminded me of the difference between `Result` and `Option`. I know I suggested `Option<Violation>`, but my main point at the time was that we should use pure functions and immutable data types.

Do you think it's worth considering whether we should actually be using `Result<(), Violation>` instead of `Option<Violation>` given that invalid data might be considered an error?

src/constraint.rs
```diff
@@ -161,8 +188,22 @@ impl AnyOfConstraint {
...
+            "any_of",
+            "no_types_matched",
+            "value matches none of the types",
+        );
+        let mut type_results = vec![];
+        for type_id in &self.type_ids {
+            let type_def = type_store.get_type_by_id(*type_id).unwrap();
+            type_results.push(type_def.validate(value, type_store));
+        }
+        let total_valid_types = type_results.iter().filter(|t| t.is_none()).count();
```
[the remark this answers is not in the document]
desaikd: This code style can be used for a single violation object not for multiple violations or `Vec<Violation>`. Although with the new change for adding violations due to storing violations for types that didn't match, it uses above mentioned `Some(violation) ...` format .

src/types.rs
```diff
@@ -147,12 +189,22 @@ impl PartialEq for TypeDefinitionImpl {
...
-        todo!()
+    fn is_valid(&self, value: &OwnedElement, type_store: &TypeStore) -> bool {
+        let violation = self.validate(value, type_store);
+        violation.is_none()
     }
 
-    fn validate(&self, value: &OwnedElement, issues: &mut Violations) {
-        todo!()
+    fn validate(&self, value: &OwnedElement, type_store: &TypeStore) -> Option<Violation> {
+        let mut issues = Violation::new("", "", "");
```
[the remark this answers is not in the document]
desaikd: Added messages for these `Violation`.

src/violation.rs
```diff
@@ -1,15 +1,31 @@
...
+        }
     }
-}
 
-// TODO: Fill the struct
-#[derive(Debug, Clone)]
-pub struct Violation {}
+    pub fn add_violation(&mut self, violation: Violation) {
+        self.violations.push(violation);
+    }
```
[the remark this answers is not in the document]
desaikd: Added `with_violations` method in `Violation` implementation which resolves his issue.

src/types.rs
```diff
@@ -55,12 +72,37 @@ impl TypeDefinition {
...
+impl TypeValidator for TypeDefinition {
+    fn is_valid(&self, value: &OwnedElement, type_store: &TypeStore) -> bool {
+        todo!()
+    }
+
+    fn validate(&self, value: &OwnedElement, type_store: &TypeStore) -> Option<Violation> {
+        match self {
+            TypeDefinition::Named(named_type) => named_type.validate(value, type_store),
+            TypeDefinition::Anonymous(anonymous_type) => anonymous_type.validate(value, type_store),
+            TypeDefinition::Core(ion_type) => {
```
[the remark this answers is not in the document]
desaikd: Core type doesn't have a struct of its own as it only contains an `IonType` within it and so doesn't have a separate `validate` method for it.

src/types.rs
```diff
@@ -55,12 +72,37 @@ impl TypeDefinition {
...
+                let mut issues = Violation::new("", "", "");
+                if value.ion_type() != *ion_type {
+                    issues.add_violation(Violation::new(
+                        "type_constraint",
+                        "type_mismatched",
+                        &format!("expected type {:?}, found {:?}", ion_type, value.ion_type()),
+                    ));
+                    return Some(issues);
+                }
+                return None;
```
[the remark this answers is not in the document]
desaikd: Added messages for all `TypeDefinition` and `TypeRef` violations.

src/constraint.rs
```diff
@@ -161,8 +188,22 @@ impl AnyOfConstraint {
...
+        if total_valid_types == 0 {
+            return Some(any_of_violation);
```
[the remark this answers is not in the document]
desaikd: Violations for all the types that did not match will be added in the `violations` field of `any_of_violation`. 
Also, changed all the `validate` methods to get rid of eager allocation of `Violation`.

desaikd:
> I was talking to someone else about Rust who reminded me of the difference between `Result` and `Option`. I know I suggested `Option<Violation>`, but my main point at the time was that we should use pure functions and immutable data types.
> 
> Do you think it's worth considering whether we should actually be using `Result<(), Violation>` instead of `Option<Violation>` given that invalid data might be considered an error?

Added `ValidationResult` in `result` module which basically is  a `Result<(), Violation>` where `Violation` implements `Error`.

src/violation.rs
```diff
@@ -1,74 +1,17 @@
...
-    }
-
-    fn checkpoint(&self) -> Checkpoint {
-        Checkpoint::new(self.violations.len())
-    }
+/// Represents [Violation] found during validation with detailed error message, error code and the constraint for which the validation failed
+#[derive(Debug, Clone, Error)]
+pub struct Violation {
+    constraint: String, // represents the constraint that created this violation
+    code: String,       // represents an error code that indicates the type of the violation
```
zslayton, line 9: Are we copying these error code definitions from `ion-schema-kotlin`? I can see why they're useful, but I wonder if they should be formally defined instead of just a convention.

src/violation.rs
```diff
@@ -1,74 +1,17 @@
...
+    message: String,    // represents the detailed error message for this violation
+    violations: Vec<Violation>,
```
zslayton, line 11: This tree node structure looks much better. 👍

src/violation.rs
```diff
@@ -77,22 +20,28 @@ impl ViolationImpl {
...
+    pub fn violations(&self) -> &[Violation] {
         &self.violations
     }
+}
 
-    fn checkpoint(&self) -> Checkpoint {
-        Checkpoint::new(self.violations.len())
+impl fmt::Display for Violation {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "A validation error occurred: {}", self.message)
```
zslayton, line 45: `fmt::Display` is analogous to Java's `toString`; I think it should produce a minimal text representation of the `Violation` rather than a sentence. Its caller would provide the sentence.

What's our use case for this impl at the moment? Could we get away with just deriving `Debug` and using `{:?}` in our format strings?

desaikd: Added `Display` because it was was required  to be implemented by `Error` bound.

src/violation.rs
desaikd: Yes, I have used `ion-schema-kotlin` for reference to these error codes. But I will create an issue to make this an enum which defines all the error codes to be used by a `Violation`.

src/violation.rs
zslayton: Ah, that makes sense. Thanks for explaining. We can punt on changing the format.

zslayton (approved):

PR #55, opened 2022-01-31
Implements `contains()` for `Range`s

desaikd:
*Description of changes:*
This PR works on changes for implementation of `contains()` in `Range`.

*Changes:*
* adds `contains()` for Range struct
* Note: TODO - implement contains for Timestamp ranges.

*Tests:*
* adds unit tests for `contains()`

### DIFF
--- a/src/isl/mod.rs
+++ b/src/isl/mod.rs
@@ -248,6 +248,11 @@ mod isl_tests {
         )
     }
 
+    // helper function to return OwnedElements for range `contains` tests
+    fn elements<T: Into<OwnedElement> + std::clone::Clone>(values: &[T]) -> Vec<OwnedElement> {
+        values.iter().cloned().map(|v| v.into()).collect()
+    }
+
     #[rstest(
         range1,
         range2,
@@ -349,8 +354,8 @@ mod isl_tests {
                 range::[0, 10]
             "#
             ),
-            vec![5.into(), 0.into(), 10.into()],
-            vec![(-5).into(), 11.into()]
+            elements(&[5, 0, 10]),
+            elements(&[-5, 11])
         ),
         case::int_range_with_min(
             load_range(
@@ -358,8 +363,8 @@ mod isl_tests {
                 range::[min, 10]
             "#
             ),
-            vec![5.into(), (-5).into(), 0.into()],
-            vec![11.into()]
+            elements(&[5, -5, 0]),
+            elements(&[11])
         ),
         case::int_range_with_max(
             load_range(
@@ -367,8 +372,8 @@ mod isl_tests {
                 range::[0, max]
             "#
             ),
-            vec![5.into(), 11.into()],
-            vec![(-5).into()]
+            elements(&[5, 0, 11]),
+            elements(&[-5])
         ),
         case::int_range_with_exclusive(
             load_range(
@@ -376,8 +381,8 @@ mod isl_tests {
                 range::[exclusive::0, exclusive::10]
             "#
             ),
-            vec![5.into(), 9.into()],
-            vec![(-5).into(), 0.into(), 10.into()]
+            elements(&[5, 9]),
+            elements(&[-5, 0, 10])
         ),
         case::decimal_range(
             load_range(
@@ -385,8 +390,8 @@ mod isl_tests {
                 range::[0.0, 10.0]
             "#
             ),
-            vec![Decimal::new(55,-1).into(), Decimal::new(0, 0).into(), Decimal::new(100, -1).into()],
-            vec![Decimal::new(-55, -1).into(), Decimal::new(115, -1).into()]
+            elements(&[Decimal::new(55,-1), Decimal::new(0, 0), Decimal::new(100, -1)]),
+            elements(&[Decimal::new(-55, -1), Decimal::new(115, -1)])
         ),
         case::decimal_range_with_min(
             load_range(
@@ -394,8 +399,8 @@ mod isl_tests {
                 range::[min, 10.0]
             "#
             ),
-            vec![Decimal::new(50, -1).into(), Decimal::new(-55, -1).into(), Decimal::new(0, 0).into()],
-            vec![Decimal::new(115, -1).into()]
+            elements(&[Decimal::new(50, -1), Decimal::new(-55, -1), Decimal::new(0, 0)]),
+            elements(&[Decimal::new(115, -1)])
         ),
         case::decimal_range_with_max(
             load_range(
@@ -403,8 +408,8 @@ mod isl_tests {
                 range::[0.0, max]
             "#
             ),
-            vec![Decimal::new(55, -1).into(), Decimal::new(115, -1).into()],
-            vec![Decimal::new(-55, -1).into()]
+            elements(&[Decimal::new(55, -1), Decimal::new(115, -1)]),
+            elements(&[Decimal::new(-55, -1)])
         ),
         case::decimal_range_with_exclusive(
             load_range(
@@ -412,8 +417,8 @@ mod isl_tests {
                 range::[exclusive::1.0, exclusive::10.0]
             "#
             ),
-            vec![Decimal::new(50, -1).into(), Decimal::new(95, -1).into()],
-            vec![Decimal::new(-55, -1).into(), Decimal::new(10, -1).into(), Decimal::new(100, -1).into()]
+            elements(&[Decimal::new(50, -1), Decimal::new(95, -1)]),
+            elements(&[Decimal::new(-55, -1), Decimal::new(10, -1), Decimal::new(100, -1)])
         ),
         case::float_range(
             load_range(
@@ -421,8 +426,8 @@ mod isl_tests {
                 range::[1e2, 5e2]
             "#
             ),
-            vec![2e2.into(), 1e2.into(), 5e2.into()],
-            vec![(-1e2).into() ,1e1.into(), 6e2.into()]
+            elements(&[2e2, 1e2, 5e2]),
+            elements(&[-1e2,1e1, 6e2, f64::NAN, 0e0, -0e0])
         ),
         case::float_range_with_min(
             load_range(
@@ -430,8 +435,8 @@ mod isl_tests {
                 range::[min, 2e5]
             "#
             ),
-            vec![2e5.into(), (-2e5).into()],
-            vec![3e5.into()]
+            elements(&[f64::NEG_INFINITY, 2.2250738585072014e-308, 2e5, -2e5, 0e0, -0e0]),
+            elements(&[3e5, f64::NAN])
         ),
         case::float_range_with_max(
             load_range(
@@ -439,8 +444,8 @@ mod isl_tests {
                 range::[1e5, max]
             "#
             ),
-            vec![1e5.into(), 5e5.into(), 1e6.into()],
-            vec![(-5e5).into(), 1e2.into()]
+            elements(&[1e5, 5e5, 1e6, 1.7976931348623157e308, f64::INFINITY]),
+            elements(&[-5e5, 1e2, f64::NAN])
         ),
         case::float_range_with_exclusive(
         load_range(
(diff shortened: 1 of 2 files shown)


src/isl/util.rs
```diff
@@ -27,6 +29,224 @@ pub enum Range {
...
+                        }
+                    }
+                    (Max, _) => {
+                        unreachable!("Lower range boundary value must not be max")
+                    }
+                    (_, Min) => {
+                        unreachable!("Upper range boundary value must not be min")
+                    }
+                }
+            }
```
popematt, line 242: suggestion—The comparisons for the `(Min, Value)`, `(Value, Value)`, and `(Value, Max)` cases have repeated code. It would eliminate some of the repeated code if you refactored this so that the value we're testing is unwrapped once at the beginning of the block and so that upper and lower range bounds are considered separately (like below).
```rust
Range::Decimal(start, end) => {
    let unwrapped_value = &value.as_decimal().unwrap();
    let is_in_lower_bound = match start {
        Min => true,
        Value(start_value, boundary_type) => match start_value {
            Decimal(min_value) => match boundary_type {
                RangeBoundaryType::Inclusive => min_value <= unwrapped_value,
                RangeBoundaryType::Exclusive => min_value < unwrapped_value,
            },
            _ => unreachable!("Decimal range can only have decimals as lower and upper range boundary value"),
        },
        Max => unreachable!("Cannot have 'Max' as the lower range boundary")
    };
    // let is_in_upper_bound = match end {
    // ...
    is_in_upper_bound && is_in_lower_bound
}
```

src/isl/util.rs
```diff
@@ -27,6 +29,224 @@ pub enum Range {
 }
 
 impl Range {
+    /// Provides a boolean value to specify whether the given value is within the range or not
+    pub fn contains(&self, value: &OwnedElement) -> IonSchemaResult<bool> {
```
popematt, line 33: issue—from the spec (emphasis added):
> A number range includes all values of any numeric type (i.e. `float`, `int`, and `decimal`) that fall within that range mathematically. When testing a value for inclusion, the range bounds and the value that is being tested _will all be converted to `decimal` values for the sake of comparison_. `number ranges` do not include values of any other type.

...but this implementation does not convert any of the other number types to a decimal.

I think this is also a good opportunity to separate out the normalization to decimal from the actual comparison, which will help eliminate a lot of very similar code that repeats the comparison logic for each type.

----

question—given that numbers are treated identically when it comes to comparison, do you think it would be worth refactoring `Range` into `NumberRange` and `TimestampRange`?

src/isl/mod.rs
```diff
@@ -337,4 +338,135 @@ mod isl_tests {
...
+            vec![5.into(), 0.into(), 10.into()],
+            vec![(-5).into(), 11.into()]
+        ),
+        case::int_range_with_min(
+            load_range(
+            r#"
+                range::[min, 10]
+            "#
+            ),
+            vec![5.into(), (-5).into(), 0.into()],
```
popematt, line 361: suggestion—it would be good to also add valid test cases for `0e0`, `0d0`, `-0e0`, `-0d0`, `5d0`, `5e0`, `10e0`, and `10d0`.

src/isl/mod.rs
```diff
@@ -337,4 +338,135 @@ mod isl_tests {
...
+        case::float_range(
+            load_range(
+            r#"
+                range::[1e2, 5e2]
+            "#
+            ),
+            vec![2e2.into(), 1e2.into(), 5e2.into()],
+            vec![(-1e2).into() ,1e1.into(), 6e2.into()]
+        ),
+        case::float_range_with_min(
```
popematt, line 427: suggestion—add test case for lowest possible float value

src/isl/mod.rs
```diff
@@ -337,4 +338,135 @@ mod isl_tests {
...
+            vec![2e5.into(), (-2e5).into()],
+            vec![3e5.into()]
+        ),
+        case::float_range_with_max(
+            load_range(
+            r#"
+                range::[1e5, max]
+            "#
+            ),
+            vec![1e5.into(), 5e5.into(), 1e6.into()],
```
popematt, line 442: suggestion—add valid test case for max possible float value

src/isl/mod.rs
```diff
@@ -337,4 +338,135 @@ mod isl_tests {
...
+            vec![(-5e5).into(), 1e2.into()]
```
popematt, line 443: suggestion—add invalid test case for `nan`, `+inf`, and `-inf`

popematt (changes requested):
As I was reviewing this and suggesting more test cases, it occurred to me that these test cases should be covered by the `ion-schema-tests`. Is it possible to integrate with `ion-schema-tests` even if we can't run all of them right now? (Maybe the test runner can `unwind` any panics that are caused by a `todo!` and mark the test as "skipped" somehow.)

There might be a lot of effort saved if you could develop directly against `ion-schema-tests` instead of having to replicate _every_ possible edge case with unit tests.

src/isl/util.rs
```diff
@@ -27,6 +29,224 @@ pub enum Range {
...
+        use RangeBoundaryValue::*;
+        use RangeBoundaryValueType::*;
+        match self {
+            Range::Integer(start, end) => {
+                match (start, end) {
+                    (Min, Max) => {
+                        unreachable!("Range boundaries can not be min and max together (i.e. range::[min, max] is not allowed)")
+                    },
```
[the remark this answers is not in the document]
desaikd: Yes. Many of these `unreachable!` statements are because they have been already checked while constructing the `Range`.

src/isl/util.rs
desaikd: I can certainly normalize ranges to decimal for comparison. 

From my understanding, you are suggesting to refactor `Range` into `NumberRange` and `TimestampRange` as all number `Range`s will be converted to `Decimal` under the hood. But from the reference of `ion-schema-kotlin` [implementation](https://github.com/amzn/ion-schema-kotlin/blob/master/src/com/amazon/ionschema/internal/util/RangeIonTimestamp.kt#L53) and from above statement it does seem like all types of ranges are being converted to `Decimal`, so maybe a single method implementation could be used for all of them. Unless, there was a specific reason you suggested this?

desaikd:
> As I was reviewing this and suggesting more test cases, it occurred to me that these test cases should be covered by the `ion-schema-tests`. Is it possible to integrate with `ion-schema-tests` even if we can't run all of them right now? (Maybe the test runner can `unwind` any panics that are caused by a `todo!` and mark the test as "skipped" somehow.)
> 
> There might be a lot of effort saved if you could develop directly against `ion-schema-tests` instead of having to replicate _every_ possible edge case with unit tests.

Yes, I can integrate with `ion-schema-tests` in a separate PR. As for `Range#contains` we would need separate unit test cases from `ion-schema-tests` as it doesn't include `Range#contains` specific tests.

popematt:
> > As I was reviewing this and suggesting more test cases, it occurred to me that these test cases should be covered by the `ion-schema-tests`. Is it possible to integrate with `ion-schema-tests` even if we can't run all of them right now? (Maybe the test runner can `unwind` any panics that are caused by a `todo!` and mark the test as "skipped" somehow.)
> > There might be a lot of effort saved if you could develop directly against `ion-schema-tests` instead of having to replicate _every_ possible edge case with unit tests.
> 
> Yes, I can integrate with `ion-schema-tests` in a separate PR for all ISL constraints. As for `Range#contains` we would need separate unit test cases from `ion-schema-tests` as it doesn't include `Range#contains` specific tests.

You're right that it wouldn't contain tests specifically for `Range#contains`, but it does contain test cases that could have caught the some things here (via the `valid_values` tests). I'm certainly not suggesting that we do away with unit testing. Anyway, this is not something that needs to be resolved for this PR—it was just an observation that I had. Sorry to distract from the main point of the PR.

src/isl/util.rs
popematt: > From my understanding, you are suggesting to refactor Range into NumberRange and TimestampRange as all number Ranges will be converted to Decimal under the hood. But from the reference of ion-schema-kotlin implementation and from above statement it does seem like all types of ranges are being converted to Decimal, so maybe a single method implementation could be used for all of them. Unless, there was a specific reason you suggested this?

I don't have a clearly formed idea about this—I was just throwing an idea out there to see if you had an opinion on it. Don't worry about it unless it seems to you like it would make the implementation simpler.

desaikd: Sure. Thanks. I will try to restructure this for using `Decimal` comparison and see if we can change something here.

src/isl/util.rs
```diff
@@ -27,6 +29,224 @@ pub enum Range {
...
 impl Range {
+    /// Provides a boolean value to specify whether the given value is within the range or not
+    pub fn contains(&self, value: &OwnedElement) -> IonSchemaResult<bool> {
+        use RangeBoundaryValue::*;
+        use RangeBoundaryValueType::*;
+        match self {
+            Range::Integer(start, end) => {
+                match (start, end) {
+                    (Min, Max) => {
+                        unreachable!("Range boundaries can not be min and max together (i.e. range::[min, max] is not allowed)")
```
zslayton, line 40: Are all of these `unreachable!` states being checked when the `Range` is being constructed? For example, is this `unreachable!` because we proved that the range wasn't `range::[min, max]` when we parsed it?

I'm confirming that this is a code path we genuinely don't expect to encounter. If this is the first time we're checking for this case, we should return an error instead.

desaikd: Yes, its been checked while the construction of the range.

src/isl/util.rs
```diff
@@ -35,210 +35,147 @@ impl Range {
...
-                                        Ok(int_start_value < &value.as_any_int().unwrap().as_i64().unwrap() || (*boundary_type == RangeBoundaryType::Inclusive && int_start_value == &value.as_any_int().unwrap().as_i64().unwrap()))
-                                    }
-                                    AnyInt::BigInt(big_int_start_value) => {
-                                        Ok(big_int_start_value < &value.as_any_int().unwrap().as_big_int().unwrap() || (*boundary_type == RangeBoundaryType::Inclusive && big_int_start_value == value.as_any_int().unwrap().as_big_int().unwrap()))
-                                    }
-                                }
-                            },
-                            _ => unreachable!("Integer range can only have integers as lower and upper range boundary value")
-                        }
+                let unwrapped_value = value.as_any_int().unwrap();
```
zslayton, line 38: A few thoughts here:

1. If this is a `Range<integer>` and I ask whether a `float` is in the range, I think this should return an `Err` instead of panicking.
2. I think eventually we'll need an enum variant for `Range<number>` [as described here](https://amzn.github.io/ion-schema/docs/spec.html#rangenumber). This is where we'd convert any/all types to `decimal` during comparison. We can do this in a follow-on PR.
3. (Just an observation, no change needed.) We don't have any use for `value` after it gets unwrapped. This seems like a situation where variable shadowing would work nicely.

```rust
let value = value.as_any_int().unwrap();
```

This creates a new variable with the same name (`value`) but a different type (`AnyInt` instead of `Option<AnyInt>`).

desaikd: For 2, I meant to create an issue to track it. Here's the reference issue: https://github.com/amzn/ion-schema-rust/issues/58

src/isl/util.rs
```diff
@@ -35,12 +35,16 @@ impl Range {
         use RangeBoundaryValueType::*;
         match self {
             Range::Integer(start, end) => {
-                let unwrapped_value = value.as_any_int().unwrap();
+                let value = value.as_any_int().ok_or_else(|| {
+                    return invalid_schema_error_raw(
+                        "Integer ranges can only have integer value for validation",
+                    );
+                })?;
```
zslayton, line 42: Here and elsewhere in the PR, you can simplify this to:
```rust
let value = value.as_any_int().ok_or_else(|| invalid_schema_error_raw("..."))?;
```
(That is: you can remove the explicit `return` and braces.)

zslayton (approved):

popematt (approved):

Issue #13, opened 2021-08-03
Defining ISL core types which are not part of `IonType`s


desaikd:
ISL core types include all the ion types as well some more types as defined in [Ion Schema spec](https://amzn.github.io/ion-schema/docs/spec.html#core-types):

> 
    scalars: blob, bool, clob, decimal, float, int, string, symbol, timestamp
    containers: list, sexp, struct
    others:
        lob: represents a blob or clob
        number: represents a decimal, float, or int
        text: represents a string or symbol
        document: represents a series of top-level values
        nothing: has no valid value (useful for reserving a field name)
        any: represents any of the types listed above as well as types derived from those types

define all other types which are not part of Ion types like: `lob`, `number`, `text` etc using schema itself as shown below:
```
 any: type::{
        one_of:[
          blob,
          bool,
          clob,
          decimal,
          document,
          float,
          int,
          string,
          symbol,
          timestamp,
          list,
          sexp,
          struct
        ]
      }
```

PR #59, opened 2022-02-15
Adds changes for core isl types

desaikd:
*Issue #13*

*Description of changes:*
This PR works on implementing [ISL core types](https://amzn.github.io/ion-schema/docs/spec.html#core-types)

*Changes:*
* adds `CoreIslType` and `CoreISLTypeDefinition` for lob, number, text, any
core types implementation
* renames `TypeDefintionImpl` method
`parse_from_isl_type_and_update_pending_types`
* adds changes to `TypeStore`, `PedningTypes` to store core types
* adds unit tests for this change
* `core_types.rs` stores `Hashmap` of ISL core types and their definition

*Test:*
adds tests for ISL core types

### DIFF
--- a/src/isl/core_types.rs
+++ b/src/isl/core_types.rs
@@ -1,0 +1,8 @@
+HashMap::from([
+    ("any", "type::{ name: any, one_of: [ blob, bool, clob, decimal,
+                                    float, int, string, symbol, timestamp,
+                                    list, sexp, struct ] }"),
+    ("lob" , "type::{ name: lob, one_of: [ blob, clob ] }"),
+    ("number", "type::{ name: number, one_of: [ decimal, float, int ] }"),
+    ("text", "type::{ name: text, one_of: [ string, symbol ] }")
+])
--- a/src/isl/isl_type_reference.rs
+++ b/src/isl/isl_type_reference.rs
@@ -89,12 +89,7 @@ impl IslTypeRef {
         type_store: &mut TypeStore,
         pending_types: &mut PendingTypes,
     ) -> IonSchemaResult<TypeId> {
-        let invalid_type_reference_error = unresolvable_schema_error(format!(
-            "Could not resolve type reference: {:?} does not exist",
-            alias
-        ));
-
-        return if let Some(type_id) = type_store.get_builtin_type_id(alias) {
+        if let Some(type_id) = type_store.get_builtin_type_id(alias) {
             // verify if the given alias is a Built-in type and if it is then return the type id from type_store
             // All Built-in types are preloaded into the type_store
             // Built-in types includes all ion types and derived types like any, lob, text, number, $int, $float, ...
@@ -107,11 +102,17 @@ impl IslTypeRef {
             if parent.0.eq(alias) {
                 Ok(parent.1)
             } else {
-                invalid_type_reference_error
+                unresolvable_schema_error(format!(
+                    "Could not resolve type reference: {:?} does not exist",
+                    alias
+                ))
             }
         } else {
-            invalid_type_reference_error
-        };
+            unresolvable_schema_error(format!(
+                "Could not resolve type reference: {:?} does not exist",
+                alias
+            ))
+        }
     }
 
     // TODO: break match arms into helper methods as we add more constraints
--- a/src/isl/isl_type_reference.rs
+++ b/src/isl/isl_type_reference.rs
@@ -2,43 +2,21 @@ use crate::isl::isl_constraint::IslConstraint;
 use crate::isl::isl_import::{IslImport, IslImportType};
 use crate::isl::isl_type::IslTypeImpl;
 use crate::result::{
-    invalid_schema_error, invalid_schema_error_raw, unresolvable_schema_error, IonSchemaResult,
+    invalid_schema_error, invalid_schema_error_raw, unresolvable_schema_error,
+    unresolvable_schema_error_raw, IonSchemaResult,
 };
 use crate::system::{PendingTypes, TypeId, TypeStore};
-use crate::types::{CoreISLTypeDefinition, TypeDefinitionImpl};
+use crate::types::TypeDefinitionImpl;
 use ion_rs::value::owned::OwnedElement;
-use ion_rs::value::reader::{element_reader, ElementReader};
 use ion_rs::value::{Element, Struct, SymbolToken};
 use ion_rs::IonType;
-use std::collections::HashMap;
-
-/// Represents all [Core ISL types]
-/// This also includes: lob, number, text, any
-/// [Core ISL types]: https://amzn.github.io/ion-schema/docs/spec.html#core-types
-#[derive(Debug, Clone, PartialEq)]
-pub enum CoreIslType {
-    IonType(IonType),
-    Other(IslTypeImpl),
-}
-
-impl CoreIslType {
-    pub fn core_ion_type(ion_type: IonType) -> Self {
-        Self::IonType(ion_type)
-    }
-
-    pub fn other_isl_type(isl_type: IslTypeImpl) -> Self {
-        Self::Other(isl_type)
-    }
-}
 
 /// Provides an internal representation of a schema type reference.
 /// The type reference grammar is defined in the [Ion Schema Spec]
 /// [Ion Schema spec]: https://amzn.github.io/ion-schema/docs/spec.html#grammar
 #[derive(Debug, Clone, PartialEq)]
 pub enum IslTypeRef {
-    /// Represents a reference to one of ISL's built-in core types:
-    CoreIsl(CoreIslType),
-    /// Represents a reference to a named type (including aliases)
+    /// Represents a reference to a named type (including aliases and built-in types)
     Named(String),
     /// Represents a type reference defined as an inlined import of a type from another schema
     TypeImport(IslImportType),
@@ -48,34 +26,14 @@ pub enum IslTypeRef {
 
 // TODO: add a check for nullable type reference
 impl IslTypeRef {
-    /// Creates a [IslTypeRef::CoreIsl] using the [IonType] referenced inside it
-    pub fn core(ion_type: IonType) -> IslTypeRef {
-        IslTypeRef::CoreIsl(CoreIslType::IonType(ion_type))
-    }
-
-    /// Creates a [IslTypeRef::CoreIsl] using the [IonType] referenced inside it
+    /// Creates a [IslTypeRef::Named] using the [IonType] referenced inside it
     pub fn named<A: Into<String>>(name: A) -> IslTypeRef {
         IslTypeRef::Named(name.into())
     }
 
-    /// Creates a [IslTypeRef::CoreIsl] using the [IonType] referenced inside it
+    /// Creates a [IslTypeRef::Anonymous] using the [IonType] referenced inside it
     pub fn anonymous<A: Into<Vec<IslConstraint>>>(constraints: A) -> IslTypeRef {
         IslTypeRef::Anonymous(IslTypeImpl::new(None, constraints.into()))
-    }
-
-    pub fn core_isl_type_from_type_name(type_name: &str) -> IslTypeRef {
-        // get the core types map and related text value for given type_name
-        let core_types_map: HashMap<&str, &str> = include!("core_types.rs");
-        let text = core_types_map.get(type_name).unwrap().to_owned();
-        let pending_types = PendingTypes::new();
-        let isl_type = IslTypeImpl::from_owned_element(
-            &element_reader()
-                .read_one(text.as_bytes())
-                .expect("parsing failed unexpectedly"),
-            &mut vec![],
-        )
-        .unwrap();
-        IslTypeRef::CoreIsl(CoreIslType::Other(isl_type))
     }
 
     /// Tries to create an [IslTypeRef] from the given OwnedElement
@@ -94,23 +52,6 @@ impl IslTypeRef {
                     })
                     .and_then(|type_name| {
                         let ion_type = match type_name {
-                            "int" => IslTypeRef::core(IonType::Integer),
-                            "float" => IslTypeRef::core(IonType::Float),
-                            "decimal" => IslTypeRef::core(IonType::Decimal),
-                            "timestamp" => IslTypeRef::core(IonType::Timestamp),
-                            "string" => IslTypeRef::core(IonType::String),
-                            "symbol" => IslTypeRef::core(IonType::Symbol),
-                            "bool" => IslTypeRef::core(IonType::Boolean),
-                            "blob" => IslTypeRef::core(IonType::Blob),
-                            "clob" => IslTypeRef::core(IonType::Clob),
-                            "sexp" => IslTypeRef::core(IonType::SExpression),
-                            "list" => IslTypeRef::core(IonType::List),
-                            "struct" => IslTypeRef::core(IonType::Struct),
-                            "any" => IslTypeRef::core_isl_type_from_type_name("any"),
-                            "text" => IslTypeRef::core_isl_type_from_type_name("text"),
-                            "lob" => IslTypeRef::core_isl_type_from_type_name("lob"),
-                            "number" => IslTypeRef::core_isl_type_from_type_name("number"),
-                            // TODO: add a match for document core type
                             _ => IslTypeRef::Named(type_name.to_owned()),
                         };
                         Ok(ion_type)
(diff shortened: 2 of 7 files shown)


src/isl/isl_type_reference.rs
```diff
@@ -5,18 +5,39 @@ use crate::result::{
...
+use std::collections::HashMap;
+
+/// Represents all [Core ISL types]
+/// This also includes: lob, number, text, any
+/// [Core ISL types]: https://amzn.github.io/ion-schema/docs/spec.html#core-types
+#[derive(Debug, Clone, PartialEq)]
+pub enum CoreIslType {
+    IonType(IonType),
+    Other(IslTypeImpl),
+}
```
popematt, line 22: question—Is `IonType` for types like `$string` and `$blob` and `Other` is for `string`, `blob`, `any`, `$any`, etc?

src/isl/isl_type_reference.rs
```diff
@@ -42,6 +63,21 @@ impl IslTypeRef {
...
+        // get the core types map and related text value for given type_name
+        let core_types_map: HashMap<&str, &str> = include!("core_types.rs");
+        let text = core_types_map.get(type_name).unwrap().to_owned();
+        let pending_types = PendingTypes::new();
+        let isl_type = IslTypeImpl::from_owned_element(
+            &element_reader()
+                .read_one(text.as_bytes())
+                .expect("parsing failed unexpectedly"),
+            &mut vec![],
+        )
```
zslayton, line 76: This seems like it could be precomputed and stored in the map.

src/isl/isl_type_reference.rs
```diff
@@ -42,6 +63,21 @@ impl IslTypeRef {
         IslTypeRef::Anonymous(IslTypeImpl::new(None, constraints.into()))
     }
 
+    pub fn core_isl_type_from_type_name(type_name: &str) -> IslTypeRef {
+        // get the core types map and related text value for given type_name
+        let core_types_map: HashMap<&str, &str> = include!("core_types.rs");
+        let text = core_types_map.get(type_name).unwrap().to_owned();
```
zslayton, line 69: This will panic if the `type_name` is unexpected. If that's the intent, let's document it and use `expect` to provide context if it ever happens.

src/isl/isl_type_reference.rs
```diff
@@ -29,7 +50,7 @@ pub enum IslTypeRef {
 impl IslTypeRef {
     /// Creates a [IslTypeRef::CoreIsl] using the [IonType] referenced inside it
     pub fn core(ion_type: IonType) -> IslTypeRef {
```
zslayton, line 52: We should reevaluate these method names now that there are two kinds of `core` type.

src/system.rs
```diff
@@ -272,20 +271,29 @@ impl PendingTypes {
         type_id + type_store.types_by_id.len()
     }
 
-    /// Adds the [CoreTypeDefinition] and the associated ion_type in the [PendingTypes] and returns the [TypeId] for it
+    /// Adds the [CoreISLTypeDefinition] in the [PendingTypes] and returns the [TypeId] for it
     /// If the given name already exists in the [TypeStore] or [PendingTypes] it returns the associated [TypeId]
-    pub fn add_core_type(&mut self, ion_type: IonType, type_store: &mut TypeStore) -> TypeId {
-        let core_type = &format!("{:?}", ion_type);
-        if let Some(exists) = self.core_type_ids_by_ion_type.get(core_type) {
+    pub fn add_core_type(
```
popematt, line 276: If we're going to load these types lazily, why not put them in the same hash map as the imported types? They are basically an import (since they come from outside of the schema) even if the import is implicit.

Also, since the core types are always available to all schemas, would it make sense to eagerly load them instead?

src/isl/core_types.rs
```diff
@@ -0,0 +1,8 @@
+HashMap::from([
+    ("any", "type::{ name: any, one_of: [ blob, bool, clob, decimal,
+                                    float, int, string, symbol, timestamp,
+                                    list, sexp, struct ] }"),
+    ("lob" , "type::{ name: lob, one_of: [ blob, clob ] }"),
+    ("number", "type::{ name: number, one_of: [ decimal, float, int ] }"),
+    ("text", "type::{ name: text, one_of: [ string, symbol ] }")
+])
```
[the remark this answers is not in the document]
desaikd: I can probably use `phf` to make this a static constant.

src/types.rs
```diff
@@ -101,16 +138,23 @@ impl TypeValidator for TypeDefinition {
...
+                    }
+                    return Ok(());
                 }
-                return Ok(());
-            }
+                CoreISLTypeDefinition::Other(other_type) => {
+                    let type_id = type_store.get_core_type_id(core_type).unwrap();
+                    let type_def = type_store.get_type_by_id(*type_id).unwrap();
+                    type_def.validate(value, type_store)
+                }
```
popematt, line 156: This is basically the same logic as what we have for `TypeDefinition::Named::validate(value, typestore)`. I think we're going to run into maintenance problems if we have two separate code paths for the `Other` and `Named` types. Can we (internally) represent types like `any` and `lob` as `Named` types instead of having `Other`?

src/isl/isl_type_reference.rs
desaikd: As per this implementation:
```
IonType - string, blob, clob, int, decimal, float (no '$' prefixed ion types)
Other - any, lob, number, text (no '$' prefixed ion types)
```

src/isl/core_types.rs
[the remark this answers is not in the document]
popematt: Rather than add a new dependency, why not create a regular constant using `const`? Is there something about Rust that prevents us from creating a constant like this example (pseudocode):

```
const core_isl_types = listOf(
  new IslType(
    name = "lob",
    constraints = listOf(
      new TypeRef("blob"),
      new TypeRef("clob")
    )
  ),
  new IslType(
    name = "text",
    constraints = listOf(
      new TypeRef("string"),
      new TypeRef("symbol")
    )
  ),
  // etc.
)
```

src/types.rs
desaikd: Only reason to separate out `Others` from `Named` types is to have `core_types` stored in a separate `hashmap` inside `TypeStore`/ `PendingTypes`. This will eventually be helpful for getting types defined within a `schema` to not include the `core_types`.

src/isl/isl_type_reference.rs
```diff
@@ -5,18 +5,39 @@ use crate::result::{
...
 use ion_rs::IonType;
+use std::collections::HashMap;
+
+/// Represents all [Core ISL types]
+/// This also includes: lob, number, text, any
+/// [Core ISL types]: https://amzn.github.io/ion-schema/docs/spec.html#core-types
+#[derive(Debug, Clone, PartialEq)]
+pub enum CoreIslType {
+    IonType(IonType),
+    Other(IslTypeImpl),
```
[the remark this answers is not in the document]
desaikd: This PR doesn't include the '$' prefixed ion types but as per our offline discussion I intend to add them along with all other `Core` types.
I am going with the suggested naming as per our offline discussion,
```
- CoreIslType - BuiltInType
- IonType -  AtomicType
- Other (lob, number, text, any, and '$' prefixed types `type:: { name: '$string', type: nullable:: string }` ) - DerivedType
```
Does this make sense? Any suggestions/changes for this @zslayton @popematt?

src/types.rs
popematt: If we're storing them in separate places, we could still use the same `TypeDefinition` implementation, right? We'll still have separate code for retrieving the types, but we don't need to have duplicate code for evaluating them.

src/isl/core_types.rs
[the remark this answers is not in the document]
popematt: Actually, I think the use of `phf` and programmatically constructing the types are probably orthogonal to each other. Can we programmatically build the types _and_ use `phf`?

src/types.rs
desaikd: Yes. `TypeDefinition` is still the same but this is an enum variant of `TypeDefinition`. 
```
pub enum TypeDefinition {
    Named(TypeDefinitionImpl),
    Anonymous(TypeDefinitionImpl),
    Core(CoreISLTypeDefinition),
}

```
But I do understand this should not be exposed to public APIs and hence I can change the methods of `TypeStore` to use `TypeDefinition` for adding and getting types instead of `CoreISLTypeDefinition`. Does this sound about right?

src/isl/core_types.rs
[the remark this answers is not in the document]
desaikd: Yes, I can use `phf` to construct static constant hashmap of  `(String, IslType)`.

src/isl/isl_type_reference.rs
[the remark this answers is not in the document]
popematt: These names are very good. 

I still think that we could probably eliminate `DerivedType` in favor of reusing the regular named type definition, but it's not a hill I would die on.

src/isl/core_types.rs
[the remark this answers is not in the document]
zslayton: > Is there something about Rust that prevents us from creating a constant [...]

You can make a `const` array, but you can't create a `const HashMap` in Rust because heap allocation is not (yet) something that's available in a `const` context. `phf` sidesteps this by generating code to construct the immutable map at compile time.

src/isl/isl_type_reference.rs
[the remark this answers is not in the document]
zslayton: I think you can get away with two variants of `BuiltInType`: `Atomic` and `Derived`.

> I still think that we could probably eliminate DerivedType in favor of reusing the regular named type definition, but it's not a hill I would die on.

I'm in favor of doing this at the `IslTypeReference` level.

desaikd:
Commit changes for https://github.com/amzn/ion-schema-rust/pull/59/commits/099f7070bf66a9db2be12e0e12a1d57b03af7b67:
* Removed CoreISLType enum from the ISL internal representation layer
* Removed enum variant CoreIsl from IslTypeRef, 'Named(String)' enum
varint will now represent both aliases and builtin types
* Renamed CoreIslTpeDeifnition to be BuitlInTypeDefinition
* Atomic(IonType, bool) enum variant represents all ion types and all
'$' prefixed ion types where bool represents whether they allow null or
not
* Derived(TypeDefinitionImpl) enum variant represents all derived ISL
types like any, number, text, lob
* All the BuiltInTypeDefinition(Atomic + Derived) will be preloaded to
type_store (type_ids: 0-31)
* Resolved bug: Changed PendingTypes#get_total_types() to include
type_store length for correct type_id calculation
*  TypeStore#new() now returns an IonSchemaResult<Self> due to preload()
* Adds more validation and load_schema tests for buitlin types
* Includes 'InvalidNull' enum variant for Violations

src/types.rs
desaikd: Preloading all the builtin types to `TypeStore`.

src/system.rs
desaikd: Preloading all the builtin types to `TypeStore`.

src/isl/isl_type_reference.rs
desaikd: Renamed all builtin types:
   - CoreIslTypeDeifnition --> BuitlInTypeDefinition
   - Atomic(IonType, bool) --> all ion types and all '$' prefixed ion types where bool represents whether they allow null or not
   **note: This can be changed to just be `Atomic(IonType)`. Once `nullable` is implemented '$' prefixed ion types can be added to `Derived` with `type:: { name: '$string', type: nullable:: string }`**
   - Derived(TypeDefinitionImpl) --> all derived ISL types like any, number, text, lob, $any, $number, $text, $lob

src/isl/isl_type_reference.rs
desaikd: This issues is resolved as preloading the `type_store` with all builtin types which doesn't use any `HashMap`

src/isl/isl_type_reference.rs
```diff
@@ -42,6 +63,21 @@ impl IslTypeRef {
         IslTypeRef::Anonymous(IslTypeImpl::new(None, constraints.into()))
     }
 
+    pub fn core_isl_type_from_type_name(type_name: &str) -> IslTypeRef {
+        // get the core types map and related text value for given type_name
+        let core_types_map: HashMap<&str, &str> = include!("core_types.rs");
```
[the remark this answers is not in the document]
desaikd: Removed usage of `Hashmap` in favor of preloading all the built-in types which uses an array. Hence not including dependency to `phf`.

src/isl/isl_type_reference.rs
popematt: I think this approach is fine, except for one thing (which is non-blocking for this PR).

> Once nullable is implemented '$' prefixed ion types can be added to Derived with `type:: { name: '$string', type: nullable:: string }`

Actually, `nullable::string` is not equivalent to `$string`—`nullable::string` would allow `null.null` and `null.string`, whereas `$string` would allow only `null.string`. Also, [`nullable` is a bit messy](https://github.com/amzn/ion-schema/issues/38), and I don't think it's wise to use it as part of defining some of the types that are part of the Ion Schema language.

On further thought, you also cannot define `$string` as `{ one_of: [ string, { valid_values: [null.string] }] }` because of the implicit `type: any`.

That leaves us with 3 options.
1.  `$string` is the Atomic type, and then we derive `string` as `{ type: $string, not: { valid_values: [null.string] } }`
2. Both `$string` and `string` are Atomic types
3. We create a new Atomic type `any_null` (or something like that), so that we can get around the implicit `type: any` problem I mentioned above.

But again, this is something that does not need to be decided right now.

src/isl/isl_type_reference.rs
```diff
@@ -103,45 +84,98 @@ impl IslTypeRef {
...
     ) -> IonSchemaResult<TypeId> {
-        match type_reference {
-            IslTypeRef::CoreIsl(ion_type) => {
-                // TODO: create CoreType struct for storing ISLCoreType type definition instead of Type
-                // inserts ISLCoreType as a Type into type_store
-                Ok(pending_types.add_core_type(ion_type.to_owned(), type_store))
-            }
-            IslTypeRef::Named(alias) => {
+        use IonType::*;
+        let invalid_type_reference_error = unresolvable_schema_error_raw(format!(
```
popematt, line 94: question—How do you decide when to use `unresolvable_schema_error_raw` vs `unresolvable_schema_error`?

src/isl/isl_type_reference.rs
```diff
@@ -103,45 +84,98 @@ impl IslTypeRef {
...
+                            return unresolvable_schema_error(format!(
+                                "Could not resolve type reference: {:?} does not exist",
+                                alias
+                            ))
+                        }
                     },
                 }
             }
+        }
+        .ok_or(invalid_type_reference_error)
```
popematt, line 162: The match condition for all of the built-in types is a huge wall of essentially repeated code, and it means that we have more than one place where we need to make sure that all of the built-in types are enumerated correctly. Since the built-in types are preloaded, we know that they will already be in the `type_store`, so rather than matching on type names, would something like this (rust-ish) code sample work?

```rust
return if let Some(type_id) = type_store.get_built_in_type(alias) {
    type_id
} else if let Some(type_id) = pending_types.get_type_id_by_name(alias, type_store) {
    type_id
} else if let Some(parent) = pending_types.get_parent() && parent.0.eq(alias) {
    parent.1
} else {
    invalid_type_reference_error = unresolvable_schema_error_raw(format!(
        "Could not resolve type reference: {:?} does not exist",
        alias
    ))
}
```

If we absolutely must match the name before we try to get it from the `type_store`, then we should so something like this instead of having a match condition for each possible type name:
```rust
const BUILT_IN_TYPE_NAMES: &[&str] = &["$string", "string" /* etc. */];

// ...

return if BUILT_IN_TYPE_NAMES.into_iter().contains(alias) {
    type_store.get_built_in_type(alias)
} //...
```

src/violation.rs
```diff
@@ -50,6 +50,7 @@ impl fmt::Display for Violation {
 #[derive(Debug, Clone)]
 pub enum ViolationCode {
     NoTypesMatched,
+    InvalidNull,
```
popematt, line 53: minor suggestion—it would be good to add a comment indicating how/when this is to be used. Some of the other `ViolationCode`s are pretty self-explanatory, but I think this one is not.

src/isl/isl_type_reference.rs
desaikd: Basically `unresolvable_schema_error_raw` returns `IonSchemaError` whereas `unresolvable_schema_error` returns `IonSchemaResult` which is `IonSchemaError` wrapped inside a `Result`.

src/isl/isl_type_reference.rs
desaikd: I think for now we have option#2 implemented, which seems fine. Only if we have a strong preference for `Atomic` types to be atomic meaning only have `string` and not `$string`, we can go for option#1 once we have `valid_values`.

popematt (approved):

src/isl/isl_type_reference.rs
```diff
@@ -143,6 +83,37 @@ impl IslTypeRef {
...
+            Ok(type_id)
+        } else if let Some(parent) = pending_types.get_parent() {
+            // if it is a self referencing type resolve it using parent information from type_store
+            if parent.0.eq(alias) {
+                Ok(parent.1)
+            } else {
+                invalid_type_reference_error
+            }
+        } else {
+            invalid_type_reference_error
```
zslayton, line 113: Eagerly defining `invalid_type_reference_error` means that we'll be allocating/formatting an error string on every call even if nothing went wrong. I suggest defining these inline with messages that help distinguish between the two outcomes.

```suggestion
              unresolvable_schema_error(format!(
                "Could not resolve type reference: expected parent type to be {:?}",
                alias
              ));
            }
        } else {
            unresolvable_schema_error(format!(
              "Could not resolve type reference: {:?} does not exist",
              alias
            ));
```

I'm not sure I understand the first ("parent") error case, so provide a more meaningful error than my example :)

src/system.rs
```diff
@@ -354,25 +360,95 @@ impl PendingTypes {
     }
 }
 
+/// Represents a hashmap of BuiltIn derived ISL types
```
zslayton, line 363: This is no longer a `HashMap`.

src/system.rs
```diff
@@ -354,25 +360,95 @@ impl PendingTypes {
...
-        Self {
-            core_type_ids_by_name: HashMap::new(),
+    pub fn new() -> IonSchemaResult<Self> {
+        let mut type_store = Self {
+            builtin_type_ids_by_name: HashMap::new(),
             imported_type_ids_by_name: HashMap::new(),
             ids_by_name: HashMap::new(),
             types_by_id: Vec::new(),
+        };
+        type_store.preload()?;
```
zslayton, line 399: If `preload()` fails, it means there's a bug in the library. The user can't do anything to recover in that case. I think it would be appropriate to panic with a descriptive message instead of returning an `IonSchemaResult`.

src/system.rs
```diff
@@ -406,14 +482,19 @@ impl TypeStore {
...
-    pub fn get_core_type_id(&self, core_isl_type: &CoreISLTypeDefinition) -> Option<&TypeId> {
-        let core_type_name = match core_isl_type {
-            CoreISLTypeDefinition::IonType(ion_type) => format!("{}", ion_type),
-            CoreISLTypeDefinition::Other(other_type) => other_type.name().to_owned().unwrap(),
+    pub fn get_builtin_type_id(&self, type_name: &str) -> Option<TypeId> {
+        let type_name = match type_name {
+            "int" => "integer",
+            "bool" => "boolean",
+            "$int" => "$integer",
+            "$bool" => "$boolean",
```
zslayton, line 492: This looks like it might be a band-aid. Why does the type store use names that aren't the ones from the Ion spec?

src/types.rs
```diff
@@ -60,15 +63,15 @@ impl TypeRef {
...
 }
 
-/// Represents a [CoreISLTypeDefinition] which stoes a resolved core ISl type using [TypeStore]
+/// Represents a [BuiltInTypeDefinition] which stores a resolved builtin ISl type using [TypeStore]
 #[derive(Debug, Clone, PartialEq)]
-pub enum CoreISLTypeDefinition {
-    IonType(IonType),
-    Other(TypeDefinitionImpl),
+pub enum BuiltInTypeDefinition {
+    Atomic(IonType, bool), // bool represents if the atomic ion type is nullable or not
```
zslayton, line 69: Since the `bool` often requires an explanatory comment, I wonder if it's worth instead creating an enum:

```rust
enum Nullability {
  Nullable,
  NotNullable
}
```

to make it a bit more obvious? (The `Reader` does this sort of thing in `ion-rust` and I keep meaning to find another representation for it 😞)

src/system.rs
desaikd: Because I am using `IonType` which has different naming than [Ion Schema Spec](https://amzn.github.io/ion-schema/docs/spec.html#type-system). I can make the type_store use names according to Ion Schema spec but then when validating I would have to convert all those strings to `IonType` and then validate. So either way will have to use a band-aid.

src/isl/isl_type_reference.rs
desaikd: parent is used for checking if the `type_reference` is a self reference. So if the current reference/alias is not same as parent reference/alias then the current type reference does not exist.
Added this change.

zslayton (approved):
🚀

src/system.rs
zslayton: As a TODO: let's make a dedicated function to get the Ion Schema name from a given `IonType` rather than doing mappings on the spot. Then we can document that function so readers understand why the names are different.

PR #60, opened 2022-02-24
Adds changes for integrating ion-schema-tests

desaikd:
*Description of changes:*
This PR adds changes for integrating [ion-schema-tests](https://github.com/amzn/ion-schema-tests)

*Changes:*
* Bug fix: Found null instead of lists while converting
`Vec<OwnedElement>` to `Vec<IslTypeRef>`
* Removed an invalid check to require all 'type::' annotated types to be
named types
* Bug Fix: Found `null.*` as alias for type reference
* Made types module public to be used for tests
* Added `$null` in derived built-in types
* Added `schema_type_from_element()`
* Adds `type: any` as default type constraint if no type constraint
available

*Tests:*
* Adds validation tests from [ion-schema-tests](https://github.com/amzn/ion-schema-tests)

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -1,10 +1,11 @@
 use crate::constraint::Constraint;
+use crate::isl::isl_constraint::IslConstraint;
 use crate::isl::isl_type::IslTypeImpl;
 use crate::result::{IonSchemaResult, ValidationResult};
 use crate::system::{PendingTypes, TypeId, TypeStore};
 use crate::violation::{Violation, ViolationCode};
-use ion_rs::value::owned::OwnedElement;
-use ion_rs::value::Element;
+use ion_rs::value::owned::{text_token, OwnedElement};
+use ion_rs::value::{Builder, Element};
 use ion_rs::IonType;
 use std::rc::Rc;
 
@@ -220,9 +221,38 @@ impl TypeDefinitionImpl {
         let type_id = pending_types.add_type(type_store);
 
         // convert IslConstraint to Constraint
+        let mut found_type_constraint = false;
         for isl_constraint in isl_type.constraints() {
+            match isl_constraint {
+                IslConstraint::Type(_) => {
+                    found_type_constraint = true;
+                }
+                _ => {}
+            }
             let constraint =
                 Constraint::resolve_from_isl_constraint(isl_constraint, type_store, pending_types)?;
+            constraints.push(constraint);
+        }
+
+        // add `type: any` as a default type constraint if there is no type constraint found
+        if !found_type_constraint {
+            // set the isl type name for any error that is returned while parsing its constraints
+            let isl_type_name = match type_name.to_owned() {
+                Some(name) => name,
+                None => "".to_owned(),
+            };
+
+            let isl_constraint = IslConstraint::from_ion_element(
+                "type",
+                &OwnedElement::new_symbol(text_token("any")),
+                &isl_type_name,
+                &mut vec![],
+            )?;
+            let constraint = Constraint::resolve_from_isl_constraint(
+                &isl_constraint,
+                type_store,
+                pending_types,
+            )?;
             constraints.push(constraint);
         }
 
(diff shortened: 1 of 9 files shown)


tests/ion_schema_tests_runner.rs
```diff
@@ -0,0 +1,167 @@
...
+                for valid_value in valid_values {
+                    match schema_type.validate(&valid_value, type_store) {
+                        Ok(_) => {
+                            continue;
+                        }
+                        Err(error) => {
+                            println!("ERROR: {} for {:?}", error, valid_value);
+                            test_failed = true;
+                        }
+                    };
```
popematt, line 136: I think this can be simplified.

```suggestion
                    if let Error(error) = schema_type.validate(&valid_value, type_store) {
                        println!("ERROR: {} for {:?}", error, valid_value);
                        test_failed = true;
                    };
```

tests/ion_schema_tests_runner.rs
```diff
@@ -0,0 +1,167 @@
...
+                }
+
+                if test_failed {
+                    // move to next test file if any test failed
+                    continue;
+                }
+
```
popematt, line 143: Why the `break`? Ideally, we should report all of the test failures.
```suggestion
```

tests/ion_schema_tests_runner.rs
```diff
@@ -0,0 +1,167 @@
...
+                                "ERROR: expected invalid value to return error for {:?}",
+                                &invalid_value
+                            );
+                            test_failed = true;
+                            break;
+                        }
+                        Err(_) => {
+                            continue;
+                        }
+                    };
```
popematt, line 157: Ideally, all test failures should be reported, not just the first one encountered, so I don't think we should have the `break`. Also, I _think_ this can be simplified.
```suggestion
                    if let Ok(_) = schema_type.validate(&invalid_value, type_store) {
                        println!(
                            "ERROR: expected invalid value to return error for {:?}",
                            &invalid_value
                        );
                        test_failed = true;
                    };
```

tests/ion_schema_tests_runner.rs
```diff
@@ -0,0 +1,167 @@
...
+                        .unwrap(),
+                );
+            } else {
+                continue;
+            }
+        }
+
+        match type_def {
+            None => {}
+            Some(schema_type) => {
```
popematt, line 126: The `None` case is unnecessary. You can replace the `match` with an `if let`.
```rust
if let Some(schema_type) = type_def {
  // ...
}
```

tests/ion_schema_tests_runner.rs
```diff
@@ -0,0 +1,167 @@
...
+            } else if annotations.contains(&&text_token("type")) {
+                // get type definition for type validation
+                type_def = Some(
+                    schema_system
+                        .schema_type_from_element(&element, type_store)
+                        .unwrap(),
+                );
+            } else {
+                continue;
+            }
```
popematt, line 121: `else` block is redundant.
```suggestion
            }
```

src/types.rs
```diff
@@ -322,56 +352,56 @@ mod type_definition_tests {
             { name: my_int, type: { type: my_int } }
          */
         IslType::named("my_int", [IslConstraint::type_constraint(IslTypeRef::anonymous([IslConstraint::type_constraint(IslTypeRef::named("my_int"))]))]),
-        TypeDefinition::named("my_int", [Constraint::type_constraint(33)]) // 0-31 are built-in types which are preloaded to the type_store
+        TypeDefinition::named("my_int", [Constraint::type_constraint(34)]) // 0-32 are built-in types which are preloaded to the type_store
```
popematt, line 355: Using numbers to refer to specific types is hard to read. Is there any way we can look these type references up by name in the tests here?

popematt (commented):
Is the `ion_schema_tests_runner` included in `cargo test --workspace` or does it require a different command?

Is it possible to have separately reported test cases (like `#[case(some_file.isl)]`) for each `ion-schema-tests` file instead of one giant test case for everything? (Long term maybe we can switch to [something like this](https://blog.cyplo.net/posts/2018/12/generate-rust-tests-from-data/) so that we don't need to hand-maintain a list of test cases.)

tests/ion_schema_tests_runner.rs
desaikd: With the new change now all errors found during the test are reported.

tests/ion_schema_tests_runner.rs
```diff
@@ -5,163 +5,152 @@ use ion_rs::value::{Element, Sequence};
...
+#[test_resources("ion-schema-tests/core_types/*.isl")]
+#[test_resources("ion-schema-tests/constraints/all_of/*.isl")]
+#[test_resources("ion-schema-tests/constraints/any_of/*.isl")]
+#[test_resources("ion-schema-tests/constraints/not/invalid.isl")]
+#[test_resources("ion-schema-tests/constraints/not/nested.isl")]
+#[test_resources("ion-schema-tests/constraints/not/string.isl")]
+#[test_resources("ion-schema-tests/constraints/one_of/*.isl")]
+#[test_resources("ion-schema-tests/constraints/type/*.isl")]
+// `test_resources` breaks for test-case names containing `$` and it doesn't allow
+// to rename test-case names hence using `rstest` for `$*.isl` test files
```
zslayton, line 44: Do we have an issue to track this? It would be good to add a link here.

zslayton (approved):
I agree with @popematt that we need to be able to create type constraints by name instead of by ID. Please open an issue to track this.

popematt (approved):

src/types.rs
desaikd: Reference issue: https://github.com/amzn/ion-schema-rust/issues/61

PR #64, opened 2022-03-31
adds changes for fields validation

desaikd:
*Issue #9*

*Description of changes:*
This PR works on adding validation logic for fields constraint.

*Changes:*
* fills `validate()` for `FieldsConstraint`
* fix bug for invalid empty fields constraint
* Moves `occurs_validation()` to `TypeDefinition` implementation as its used by both `ordered_elements` and `fields` constraint
* adds unit test for `fields` validation

*Test:*
adds unit tests for `fields` validation

### DIFF
--- a/src/constraint.rs
+++ b/src/constraint.rs
@@ -1,15 +1,12 @@
 use crate::isl::isl_constraint::{IslConstraint, IslOccurs};
 use crate::isl::isl_type_reference::IslTypeRef;
-use crate::isl::util::Range;
 use crate::result::{IonSchemaResult, ValidationResult};
 use crate::system::{PendingTypes, TypeId, TypeStore};
 use crate::types::{TypeDefinition, TypeValidator};
 use crate::violation::{Violation, ViolationCode};
 use ion_rs::value::owned::OwnedElement;
-use ion_rs::value::{Element, Sequence};
+use ion_rs::value::{Element, Sequence, Struct};
 use std::collections::HashMap;
-use std::convert::TryInto;
-use std::iter::Peekable;
 
 /// Provides validation for schema Constraint
 pub trait ConstraintValidator {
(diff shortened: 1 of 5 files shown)


src/constraint.rs
```diff
@@ -609,6 +527,53 @@ impl FieldsConstraint {
...
+                    &format!("expected struct ion found {}", value.ion_type()),
+                    violations,
+                ));
+            }
+            Some(ion_struct) => ion_struct,
+        };
+
+        // get the values corresponding to the field_name and perform occurs_validation based on the type_def
+        for (field_name, element) in &self.fields {
+            let type_id = &self.fields.get(field_name).unwrap();
```
zslayton, line 557:
```rust
//                          v--- Isn't `element`...
       for (field_name, element) in &self.fields {
            let type_id = &self.fields.get(field_name).unwrap();
//                 ^---- ...the same thing as `type_id`?
```

zslayton (approved):

src/constraint.rs
```diff
@@ -609,6 +527,52 @@ impl FieldsConstraint {
...
+                    violations,
+                ));
+            }
+            Some(ion_struct) => ion_struct,
+        };
+
+        // get the values corresponding to the field_name and perform occurs_validation based on the type_def
+        for (field_name, type_id) in &self.fields {
+            let type_def = type_store.get_type_by_id(*type_id).unwrap();
+            let mut values = ion_struct.get_all(field_name).peekable();
```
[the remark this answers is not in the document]
desaikd: It was because `Iterator#next()` method required a  mutable reference.

src/constraint.rs
```diff
@@ -608,7 +648,97 @@ impl FieldsConstraint {
...
+
+        // Verify if open content exists in the struct fields
+        if !open_open_content {
+            for (field_name, value) in ion_struct.iter() {
+                if !self.fields.contains_key(field_name.text().unwrap()) {
+                    violations.push(Violation::new(
+                        "fields",
+                        ViolationCode::InvalidOpenContent,
+                        &format!(
+                            "Found open open_content in the struct: {:?}: {:?}",
```
popematt, line 690: Typo?
```suggestion
                            "Found open content in the struct: {:?}: {:?}",
```

src/constraint.rs
```diff
@@ -608,7 +648,97 @@ impl FieldsConstraint {
...
+                .unwrap()
+            {
+                violations.push(Violation::new(
+                    "fields",
+                    ViolationCode::TypeMismatched,
+                    &format!(
+                        "Expected {:?} of type {:?}: found {}",
+                        occurs_range,
+                        type_def,
+                        values.len()
```
popematt, line 720: minor issue—technically, we're checking the number of fields with a given name, not the number of elements with a type.
```suggestion
                        "Expected {:?} of field {:?}: found {}",
                        occurs_range,
                        field_name,
                        values.len()
```

src/constraint.rs
```diff
@@ -17,7 +16,12 @@ pub trait ConstraintValidator {
     /// adding [Violation]s and/or [ViolationChild]ren to `Err(violation)`
     /// if the constraint is violated.
     /// Otherwise, if the value passes the validation against the constraint then returns `Ok(())`.
-    fn validate(&self, value: &OwnedElement, type_store: &TypeStore) -> ValidationResult;
+    fn validate(
+        &self,
+        value: &OwnedElement,
+        type_store: &TypeStore,
+        open_content: bool,
```
popematt, line 23: issue—the `content: closed` pseudo-constraint is only ever useful with `fields` constraint, and so adding it here is polluting the API. Some others ways to solve this might be
* Pass the whole type definition to the `validate` function. This is marginally better because it feels less specific, but I think that it's really still just something that the `fields` constraint would use. Constraints shouldn't have to know about other constraints when they are testing a value.
* Add an `open_content: bool` field to `Constraint::Fields` and try to push to problem out to the translation between the `IslConstraint` and the `Constraint`. We might still need to pass around the type definition, but I think it makes more sense here because this is the most logical place (IMO) to resolve other discrepancies (such as this issue or syntactical sugar) between the language and the internal model.

src/constraint.rs
```diff
@@ -608,7 +648,97 @@ impl FieldsConstraint {
...
+                    ViolationCode::TypeMismatched,
+                    &format!("expected struct ion found {}", value.ion_type()),
+                    violations,
+                ));
+            }
+            Some(ion_struct) => ion_struct,
+        };
+
+        // Verify if open content exists in the struct fields
+        if !open_open_content {
```
popematt, line 683: Typo?
```suggestion
        if !open_content {
```

src/violation.rs
```diff
@@ -55,6 +55,8 @@ pub enum ViolationCode {
     TypeMatched,
     AllTypesNotMatched,
     TypeMismatched,
+    FieldsNotMatched,
```
popematt, line 58: nitpick/non-blocking—I notices that we have some `ViolationCode`s that use "NotMatched" and others use "Mismatched". Doesn't need to be fixed now, but I think it would be a good idea to create an issue to go and review all of the public enum names and the violation messages for stylistic consistency before we release v1.0.0.

src/constraint.rs
desaikd: Added `open_content` property to `FieldsConstraint` struct. `open_content` boolean value is passed to `FieldsConstraint` while converting from `IslConstraint` to `Constraint`.

src/violation.rs
desaikd: I will create an issue for this clean up.

PR #66, opened 2022-04-06
adds changes for `contains` constraint implementation

desaikd:
*Issue #9 #10*

*Description of changes:*
This PR works on adding implementation of contains constraint.

*Grammar:*
```ANTLR
<CONTAINS> ::= contains: [ <VALUE>... ]
```

*Ion Schema specification:*
https://amzn.github.io/ion-schema/docs/spec.html#contains

*List of changes:*
* adds implementation of `ContainsConstraint`
* adds `enum` variant `IslConstraint::Contains` and `Constraint::Contains`
* `contains()` method to create `IslConstraint::Contains` and
`Constraint::Contains` programmatically
* adds unit tests for contains constraint implementation (for
programmatically creating contains, loading schema using contains and
validation of contains)
* adds `ViolationCode` `MissingValue` for a missing value found during
contains validation

*Tests:*
added unit tests for `contains` implementation.
- adds tests for programmatically creating `contains` constraint
- adds tests for loading a schema using `contains` constraint 
- adds validation tests for `contains` constraint

### DIFF
--- a/src/constraint.rs
+++ b/src/constraint.rs
@@ -716,7 +716,7 @@ impl ConstraintValidator for FieldsConstraint {
     }
 }
 
-/// Implements an `contains` constraint of Ion Schema
+/// Implements Ion Schema's `contains` constraint
 /// [contains]: https://amzn.github.io/ion-schema/docs/spec.html#contains
 #[derive(Debug, Clone, PartialEq)]
 pub struct ContainsConstraint {
@@ -734,7 +734,7 @@ impl ContainsConstraint {
 impl ConstraintValidator for ContainsConstraint {
     fn validate(&self, value: &OwnedElement, type_store: &TypeStore) -> ValidationResult {
         // get the expected values from the contains constraint
-        let mut expected_values = self.values.clone();
+        let mut expected_values = self.values.iter();
 
         // Check for null sequence
         if value.is_null() {
@@ -745,31 +745,26 @@ impl ConstraintValidator for ContainsConstraint {
             ));
         }
 
-        let values: Vec<&OwnedElement> = match value.as_sequence() {
+        match value.as_sequence() {
             None => {
+                // return Violation if value is not an Ion sequence
                 return Err(Violation::new(
                     "contains",
                     ViolationCode::TypeMismatched,
                     &format!("expected list found {}", value.ion_type()),
                 ));
             }
-            Some(ion_sequence) => ion_sequence.iter().collect(),
-        };
-
-        for value in values {
-            if let Some(pos) = expected_values.iter().position(|v| v == value) {
-                expected_values.remove(pos);
+            Some(ion_sequence) => {
+                // if any of our expected values are not in the Ion sequence then return Violation
+                if !expected_values.all(|ev| ion_sequence.iter().find(|v| v == &ev).is_some()) {
+                    return Err(Violation::new(
+                        "contains",
+                        ViolationCode::MissingValue,
+                        &format!("{:?} has missing value(s): {:?}", value, expected_values),
+                    ));
+                }
             }
-        }
-
-        // return error if there were any values left in the expected values vector
-        if !expected_values.is_empty() {
-            return Err(Violation::new(
-                "contains",
-                ViolationCode::MissingValue,
-                &format!("{:?} has missing value(s): {:?}", value, expected_values),
-            ));
-        }
+        };
 
         Ok(())
     }
--- a/src/constraint.rs
+++ b/src/constraint.rs
@@ -733,9 +733,6 @@ impl ContainsConstraint {
 
 impl ConstraintValidator for ContainsConstraint {
     fn validate(&self, value: &OwnedElement, type_store: &TypeStore) -> ValidationResult {
-        // get the expected values from the contains constraint
-        let mut expected_values = self.values.to_owned();
-
         // Check for null sequence
         if value.is_null() {
             return Err(Violation::new(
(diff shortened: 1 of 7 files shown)


src/constraint.rs
```diff
@@ -703,3 +715,62 @@ impl ConstraintValidator for FieldsConstraint {
         Ok(())
     }
 }
+
+/// Implements an `contains` constraint of Ion Schema
```
zslayton, line 719:
```suggestion
/// Implements Ion Schema's `contains` constraint
```

src/constraint.rs
```diff
@@ -703,3 +715,62 @@ impl ConstraintValidator for FieldsConstraint {
...
+impl ContainsConstraint {
+    pub fn new(values: Vec<OwnedElement>) -> Self {
+        Self { values }
+    }
+}
+
+impl ConstraintValidator for ContainsConstraint {
+    fn validate(&self, value: &OwnedElement, type_store: &TypeStore) -> ValidationResult {
+        // get the expected values from the contains constraint
+        let mut expected_values = self.values.clone();
```
zslayton, line 737: Does this need to be `clone()`d?

src/constraint.rs
```diff
@@ -703,3 +715,62 @@ impl ConstraintValidator for FieldsConstraint {
...
+                ));
+            }
+            Some(ion_sequence) => ion_sequence.iter().collect(),
+        };
+
+        for value in values {
+            if let Some(pos) = expected_values.iter().position(|v| v == value) {
+                expected_values.remove(pos);
+            }
+        }
```
zslayton, line 763: Rather than cloning and collecting all of the collections, I think we can just do this:

```
// If any of our expected values are not in the Ion sequence, this is false
expected_values
    .iter()
    .all(|ev| ion_sequence.iter().contains(ev))
```

It's `O(n^2)`, but you've already got a TODO to address that.

https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.all
https://doc.rust-lang.org/std/vec/struct.Vec.html#method.contains

src/constraint.rs
desaikd: I am removing entries from the vector to verify if all entries were covered that's why using a second vector that is cloned.

src/constraint.rs
desaikd: Using `find` instead of `contains` to be able to use an iterator and to collect values to a vector.

https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find

src/constraint.rs
```diff
@@ -703,3 +715,57 @@ impl ConstraintValidator for FieldsConstraint {
...
+            ));
+        }
+
+        match value.as_sequence() {
+            None => {
+                // return Violation if value is not an Ion sequence
+                return Err(Violation::new(
+                    "contains",
+                    ViolationCode::TypeMismatched,
+                    &format!("expected list found {}", value.ion_type()),
```
popematt, line 754: nitpick—not necessarily a `list` could also be a `sexp` or `document`.

src/constraint.rs
```diff
@@ -703,3 +715,57 @@ impl ConstraintValidator for FieldsConstraint {
...
+    fn validate(&self, value: &OwnedElement, type_store: &TypeStore) -> ValidationResult {
+        // get the expected values from the contains constraint
+        let mut expected_values = self.values.iter();
+
+        // Check for null sequence
+        if value.is_null() {
+            return Err(Violation::new(
+                "contains",
+                ViolationCode::TypeMismatched,
+                &format!("Null sequence not allowed for contains constraint"),
```
popematt, line 744: minor suggestion—The wording here initially had me confused whether this was validating the value or validating that the schema was correct. I think it might be a little more clear to make the phrasing here be similar to the phrasing of the type mismatch on Line 754.
```suggestion
                &format!("expected a sequence found {}", value),
```

src/constraint.rs
```diff
@@ -703,3 +715,57 @@ impl ConstraintValidator for FieldsConstraint {
...
+            }
+            Some(ion_sequence) => {
+                // if any of our expected values are not in the Ion sequence then return Violation
+                if !expected_values.all(|ev| ion_sequence.iter().find(|v| v == &ev).is_some()) {
+                    return Err(Violation::new(
+                        "contains",
+                        ViolationCode::MissingValue,
+                        &format!("{:?} has missing value(s): {:?}", value, expected_values),
+                    ));
+                }
```
popematt, line 765: Please correct me if I'm wrong, but I think the error message might be misleading. For example, given this type and value:
```
type::{ contains: [a, b, c] }
```
```
[ a, d, c ]
```
The validation error message would be:
```
[a, d, c] has missing value(s): [a, b, c] 
```
But it would be more accurate to say:
```
[a, d, c] has missing value(s): [b] 
```

desaikd: Yes, this is correct. I initially had  `expected_values` keep track of this but when I changed to use `.all()` method this was removed. Sorry for the confusion but I am changing back to a for loop over given values as if stick with `.all()` I would not be able to manipulate `expected_values` to keep track of missing values.

src/constraint.rs
```diff
@@ -734,14 +734,14 @@ impl ContainsConstraint {
 impl ConstraintValidator for ContainsConstraint {
     fn validate(&self, value: &OwnedElement, type_store: &TypeStore) -> ValidationResult {
         // get the expected values from the contains constraint
-        let mut expected_values = self.values.iter();
+        let mut expected_values = self.values.to_owned();
```
popematt, line 737: Does this give you a copy of the data, or ownership of the original? If the original data gets mutated, it's going to lead to weird bugs.

src/constraint.rs
```diff
@@ -703,3 +715,65 @@ impl ConstraintValidator for FieldsConstraint {
...
+                }
+
+                // return Violation if there were any values left in the expected values vector
+                if !expected_values.is_empty() {
+                    return Err(Violation::new(
+                        "contains",
+                        ViolationCode::MissingValue,
+                        &format!("{:?} has missing value(s): {:?}", value, expected_values),
+                    ));
+                }
```
popematt, line 773: You can eliminate the need to own `expected_values` if you use a new local vec for storing missing values (rather than removing expected values that are not found).
```suggestion
                // for each value in ion sequence if it does not exist in the expected values
                // then add it to missing_values to keep track of missing values
                let mut missing_values = Vec::new();
                for expected_value in expected_values.iter() {
                    if None == ion_sequence.iter().position(|v| v == expected_value) {
                        missing_values.push(expected_value);
                    }
                }

                // return Violation if there were any values added to the missing values vector
                if !missing_values.is_empty() {
                    return Err(Violation::new(
                        "contains",
                        ViolationCode::MissingValue,
                        &format!("{:?} has missing value(s): {:?}", value, missing_values),
                    ));
                }
```

popematt (approved):

Issue #69, opened 2022-04-18
Fix Clippy reported problems and add it to CI build


desaikd:
[Clippy](https://github.com/rust-lang/rust-clippy) (Rust linter) reports a variety of style errors and suggested simplifications. We should fix them and make cargo clippy a GitHub Action so it's checked for each PR.

Reference issue from `ion-rust`: https://github.com/amzn/ion-rust/issues/348

desaikd:
This has been fixed with #70 and #71

PR #70, opened 2022-04-18
adds changes for clippy fixes and suggestions

desaikd:
*Issue: #69*

*Description of changes:*
 This PR works on adding all the changes suggested by Clippy (Rust linter).

*Changes:*
The changes are incorporated in 3 separate commits:
- Phase 1: clippy automated fixes for ion-schema crate
- Phase 2: clippy automated changes for tests folder
- Phase 3: manually added clippy suggestions
  * uses `ok_or_else` instead of `ok_or`
  * removes unnecessary `collect()` usage
  * remove `fn main()` from doctests
  * removes unnecessary lifetime usage
  * uses `assert!` instead of `assert_eq!` for `bool` comparison
  * uses `if let` instead of `unwrap`
  * `default()` implementations for TypeStore and PendingTypes
  * (Overridden clippy suggestion) uses self named constructors for `Range`
- Bug Fix: while loading a schema tested as per a doc test in `isl` module


zslayton (approved):
Excellent! 🚀

PR #74, opened 2022-04-21
adds changes for `element` constraint implementation

desaikd:
*Issue #9 #10*

*Description of changes:*
This PR works on adding implementation of `element` constraint.

*Grammar:*
```ANTLR
<ELEMENT> ::= element: <TYPE_REFERENCE>
```

*Ion Schema specification:*
https://amzn.github.io/ion-schema/docs/spec.html#element

*List of changes:*
* adds `Elemnt` enum variants for `IslConstraint` and `Constraint`
* adds `ElementConstraint` implementations
* adds validation logic for element
* adds `ViolationCode::ElementMismatched` enum variant
* adds unit tests for element

*Tests:*
added unit tests for `element` implementation.
- adds tests for programmatically creating `element` constraint
- adds tests for loading a schema using `element` constraint 
- adds validation test for `element` constraint

### DIFF
--- a/src/constraint.rs
+++ b/src/constraint.rs
@@ -994,7 +994,7 @@ impl ConstraintValidator for CodepointLengthConstraint {
     }
 }
 
-/// Implements a `element` constraint
+/// Implements the `element` constraint
 /// [element]: https://amzn.github.io/ion-schema/docs/spec.html#element
 #[derive(Debug, Clone, PartialEq)]
 pub struct ElementConstraint {
@@ -1027,10 +1027,11 @@ impl ConstraintValidator for ElementConstraint {
             return Err(Violation::new(
                 "element",
                 ViolationCode::TypeMismatched,
-                &format!("expected a container found {:?}", value),
+                &format!("expected a container but found {:?}", value),
             ));
         }
 
+        // this type_id was validated while creating `ElementConstraint` hence the unwrap here is safe
         let type_def = type_store.get_type_by_id(self.type_id).unwrap();
 
         // validate each element of the given value container
--- a/src/constraint.rs
+++ b/src/constraint.rs
@@ -31,6 +31,7 @@ pub enum Constraint {
     Contains(ContainsConstraint),
     ContentClosed,
     ContainerLength(ContainerLengthConstraint),
+    Element(ElementConstraint),
     Fields(FieldsConstraint),
     Not(NotConstraint),
     OneOf(OneOfConstraint),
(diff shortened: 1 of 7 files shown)


src/constraint.rs
```diff
@@ -977,3 +993,83 @@ impl ConstraintValidator for CodepointLengthConstraint {
...
+                    }
+                }
+            }
+            IonType::Struct => {
+                for (field_name, val) in value.as_struct().unwrap().iter() {
+                    if let Err(violation) = type_def.validate(val, type_store) {
+                        violations.push(violation);
+                    }
+                }
+            }
```
popematt, line 1051: It seems like there's a lot of places where you have to have separate branches for `struct` and `sequence`. Should we open an issue in `ion-rust` to add a `container` concept so that we could simplify this to `match value.as_container()`?

src/isl/isl_constraint.rs
```diff
@@ -171,6 +177,17 @@ impl IslConstraint {
                 value,
                 RangeType::NonNegativeInteger,
             )?)),
+            "element" => {
+                if value.ion_type() != IonType::Symbol && value.ion_type() != IonType::Struct {
+                    return Err(invalid_schema_error_raw(format!(
+                        "element constraint was a {:?} instead of a symbol/struct",
+                        value.ion_type()
+                    )));
+                }
```
popematt, line 186: question—This seems like it should be handled elsewhere. Doesn't the `from_ion_element` function validate its own input?

desaikd: Yes, This is already checked within `from_ion_element` and can be removed from here.

popematt (approved):

src/constraint.rs
```diff
@@ -977,3 +993,83 @@ impl ConstraintValidator for CodepointLengthConstraint {
         Ok(())
     }
 }
+
+/// Implements a `element` constraint
```
zslayton, line 997:
```suggestion
/// Implements the `element` constraint
```

src/constraint.rs
```diff
@@ -977,3 +993,83 @@ impl ConstraintValidator for CodepointLengthConstraint {
...
+impl ConstraintValidator for ElementConstraint {
+    fn validate(&self, value: &OwnedElement, type_store: &TypeStore) -> ValidationResult {
+        let mut violations: Vec<Violation> = vec![];
+
+        // Check for null container
+        if value.is_null() {
+            return Err(Violation::new(
+                "element",
+                ViolationCode::TypeMismatched,
+                &format!("expected a container found {:?}", value),
```
zslayton, line 1030:
```suggestion
                &format!("expected a container but found {:?}", value),
```

src/constraint.rs
```diff
@@ -977,3 +993,83 @@ impl ConstraintValidator for CodepointLengthConstraint {
...
+            ));
+        }
+
+        let type_def = type_store.get_type_by_id(self.type_id).unwrap();
```
zslayton, line 1034: I think this `unwrap` is safe because we've already validated the type ID. If that's right, could you add a comment noting that?

zslayton (approved):

PR #83, opened 2022-05-23
adds implementation of `precision` constraint

desaikd:
*Issue #9 #10*

*Description of changes:*
This PR works on adding implementation of `precision` constraint.

*Grammar:*
```ANTLR
<PRECISION> ::= precision: <INT>
              | precision: <RANGE<INT>>
```

*Ion Schema specification:*
https://amzn.github.io/ion-schema/docs/spec.html#precision

*List of changes:*
* adds changes to upgrade to v0.10.0 for `ion-rs`.
* adds `Precision` enum variants for `IslConstraint` and `Constraint`
* adds `PrecisionConstraint` implementations
* adds changes for validating precision range (_Note: precision range must be equal to or greater than 1_)
* adds validation logic for precision

*Tests:*
added unit tests for `precision` implementation.
- adds tests for programmatically creating `precision` constraint
- adds tests for loading a schema using `precision` constraint 
- adds validation tests for `precision` constraint

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-ion-rs = "0.7.0"
+ion-rs = "0.10.0"
 thiserror = "1.0"
 num-bigint = "0.3"
 num-traits = "0.2"
--- a/examples/schema.rs
+++ b/examples/schema.rs
@@ -1,11 +1,11 @@
 #[macro_use]
 extern crate clap;
 use clap::{App, ArgMatches};
-use ion_rs::text::writer::TextWriter;
 use ion_rs::value::reader::element_reader;
 use ion_rs::value::reader::ElementReader;
 use ion_rs::value::writer::{ElementWriter, Format, TextKind};
 use ion_rs::IonType;
+use ion_rs::{RawTextWriter, Writer};
 use ion_schema::authority::{DocumentAuthority, FileSystemDocumentAuthority};
 use ion_schema::result::IonSchemaResult;
 use ion_schema::system::SchemaSystem;
@@ -91,7 +91,7 @@ fn validate(command_args: &ArgMatches) -> IonSchemaResult<()> {
 
     // create a text writer to make the output
     let mut output = vec![];
-    let mut writer = TextWriter::new(&mut output);
+    let mut writer = RawTextWriter::new(&mut output);
 
     // validate owned_elements according to type_ref
     for owned_element in owned_elements {
--- a/src/constraint.rs
+++ b/src/constraint.rs
@@ -38,6 +38,7 @@ pub enum Constraint {
     OneOf(OneOfConstraint),
     OrderedElements(OrderedElementsConstraint),
     Occurs(OccursConstraint),
+    Precision(PrecisionConstraint),
     Type(TypeConstraint),
 }
 
@@ -124,6 +125,11 @@ impl Constraint {
             annotations_modifiers.contains(&"ordered"),
             annotations,
         ))
+    }
+
+    /// Creates a [Constraint::Precision] from a [Range] specifying a precision range.
+    pub fn precision(precision: Range) -> Constraint {
+        Constraint::Precision(PrecisionConstraint::new(precision))
     }
 
     /// Creates a [Constraint::Fields] referring to the fields represented by the provided field name and [TypeId]s.
@@ -237,6 +243,9 @@ impl Constraint {
                     )?;
                 Ok(Constraint::OrderedElements(ordered_elements))
             }
+            IslConstraint::Precision(precision_range) => Ok(Constraint::Precision(
+                PrecisionConstraint::new(precision_range.to_owned()),
+            )),
         }
     }
 
@@ -269,6 +278,7 @@ impl Constraint {
             Constraint::OrderedElements(ordered_elements) => {
                 ordered_elements.validate(value, type_store)
             }
+            Constraint::Precision(precision) => precision.validate(value, type_store),
         }
     }
 }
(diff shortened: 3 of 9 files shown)


Cargo.toml
```diff
@@ -22,7 +22,7 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-ion-rs = "0.7.0"
+ion-rs = "0.10.0"
```
zslayton, line 25: 👌

examples/schema.rs
```diff
@@ -91,7 +91,7 @@ fn validate(command_args: &ArgMatches) -> IonSchemaResult<()> {
 
     // create a text writer to make the output
     let mut output = vec![];
-    let mut writer = TextWriter::new(&mut output);
+    let mut writer = RawTextWriter::new(&mut output);
```
zslayton, line 94: Thinking out loud: there's not much difference for text, but `ion-rs` should encourage users to reach for the application-level `TextWriter` instead of the `RawTextWriter`. (Nothing to do here.)

src/constraint.rs
```diff
@@ -1274,3 +1284,66 @@ impl ConstraintValidator for AnnotationsConstraint {
         self.validate_unordered_annotations(value, type_store, violations)
     }
 }
+
+/// Implements an `precision` constraint of Ion Schema
```
zslayton, line 1288:
```suggestion
/// Implements Ion Schema's `precision` constraint
```

src/constraint.rs
```diff
@@ -1274,3 +1284,66 @@ impl ConstraintValidator for AnnotationsConstraint {
...
+                    "precision",
+                    ViolationCode::TypeMismatched,
+                    &format!(
+                        "expected a decimal but found {}",
+                        if value.is_null() {
+                            format!("{:?}", value)
+                        } else {
+                            format!("{}", value.ion_type())
+                        }
+                    ),
```
zslayton, line 1324: This works, but it requires two String allocations instead of one. I suggest:

```rust
if value.is_null() {
   format!("expected a decimal but found {:?}", value);
} else {
   format!("expected a decimal but found {}", value.ion_type());
}
```

Even better: `OwnedElement` should implement `Display` so we could just print the value regardless of whether it's null.

src/isl/util.rs
```diff
@@ -240,13 +242,44 @@ impl Range {
         })
     }
 
+    pub fn validate_precision_range(precision_range: &Range) -> IonSchemaResult<Range> {
+        // minimum precision must be greater than or equal to 0
```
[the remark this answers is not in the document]
desaikd: Yes, It's a typo in comment although the code does look for minimum precision as `1`.

src/isl/util.rs
```diff
@@ -240,13 +242,44 @@ impl Range {
...
     }
 
+    pub fn validate_precision_range(precision_range: &Range) -> IonSchemaResult<Range> {
+        // minimum precision must be greater than or equal to 1
+        // for more information: https://amzn.github.io/ion-schema/docs/spec.html#precision
+        match precision_range {
+            Range::IntegerNonNegative(min_value, max_value) => match min_value {
+                RangeBoundaryValue::Max => Ok(precision_range.to_owned()),
+                RangeBoundaryValue::Min => {
+                    return Range::range(
```
popematt, line 252: To fix [clippy's warning](https://github.com/amzn/ion-schema-rust/pull/83/files#annotation_3574931928):
```suggestion
                    Range::range(
```

src/isl/util.rs
```diff
@@ -240,13 +242,44 @@ impl Range {
         })
     }
 
+    pub fn validate_precision_range(precision_range: &Range) -> IonSchemaResult<Range> {
```
popematt, line 245: question—will this function catch the following edge cases (that are all valid for a `NonNegativeIntRange`)? Do we care?

1. `range::[min, 0]` – might be missed depending on whether the `min_value < max_value` check occurs before or after the `min` is substituted with `1`.
2. `range::[min, exclusive::1]` – `exclusive::1` is equivalent to `0` when used as the `max_value` of an integer range
3. `range::[exclusive::0, 10] – has a `0` as the lower bound, but is equivalent to `range::[1, 10]`, which is legal
4. `range::[1, max]` – equivalent to `range::[min, max]`, which is not allowed

I'm inclined to say that we need to make sure that (1) is covered. I'm not sure about (2). I think (3) is already illegal, and should remain that way, and that (4) is probably fine to allow.

src/schema.rs
```diff
@@ -616,6 +616,27 @@ mod schema_tests {
                         "#),
                 "annotations_type"
         ),
+        case::precision_constraint(
+            load(r#"
+                          42.
+                          42d0
+                          42d-0
```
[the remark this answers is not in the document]
popematt: > I'm not sure why precision's minimum is 1 instead of 0.

As soon as you have a digit, you have a precision of 1—even `0` has a precision of 1.

src/isl/util.rs
desaikd: Yes, according to these concerns I feel like having a separate function to create precision ranges in `utils` module would make sense.

desaikd: With new changes https://github.com/amzn/ion-schema-rust/pull/83/commits/e2d3372c24973979609350b320a605c6fac01b88, now substituting min with 1 while parsing (previously it was doing this after range was parsed), which will cover all the above concerns. Right now for checking if `start < end` for any range is commented because no timestamp comparator for ion-rust is implemented. I will get latest version of ion-rust once its released to allow this check.

src/isl/util.rs
```diff
@@ -270,11 +276,16 @@ impl Range {
...
+        value: &IntegerValue,
+        range_type: &RangeType,
     ) -> IonSchemaResult<usize> {
         match value.as_i64() {
             Some(v) => {
-                if v >= 0 {
+                // minimum precision must be greater than or equal to 1
+                // for more information: https://amzn.github.io/ion-schema/docs/spec.html#precision
+                if (range_type == &RangeType::PrecisionRange && v >= 1)
+                    || (range_type != &RangeType::PrecisionRange && v >= 0)
```
popematt, line 287: Cod re-use is good, but do the `Err` messages on lines 291, 298, 313, and 320 also need to be updated?

popematt (approved):
Aside from the comment I left about the error messages, it LGTM.

zslayton (approved):

PR #87, opened 2022-06-01
adds implementation of `timestamp_precision` constraint

desaikd:
*Issue #9 #10*

*Description of changes:*
This PR works on adding implementation of `timestamp_precision` constraint.

*Grammar:*
```ANTLR
<TIMESTAMP_PRECISION_VALUE> ::= year
                              | month
                              | day
                              | minute
                              | second
                              | millisecond
                              | microsecond
                              | nanosecond

<TIMESTAMP_PRECISION> ::= timestamp_precision: <TIMESTAMP_PRECISION_VALUE>
                        | timestamp_precision: <RANGE<TIMESTAMP_PRECISION_VALUE>>
```

*Ion Schema specification:*
https://amzn.github.io/ion-schema/docs/spec.html#timestamp_precision

*List of changes:*
* adds `TimestampPrecison` range implementation
* adds `TimestampPrecison` enum variants for `IslConstraint` and `Constraint`
* adds `TimestampPrecisonConstraint` implementations
* adds changes for validating timestamp precision range (_Note: timestamp precision range must be equal to or greater than 1_)
* adds validation logic for timestamp precision

*Tests:*
added unit tests for `timestamp_precision` implementation.
- adds tests for programmatically creating `timestamp_precision` constraint
- adds tests for loading a schema using `timestamp_precision` constraint 
- adds validation tests for `timestamp_precision` constraint

### DIFF
--- a/src/constraint.rs
+++ b/src/constraint.rs
@@ -83,7 +83,7 @@ impl Constraint {
     /// Creates a [Constraint::ContainerLength] from a [Range] specifying a length range.
     pub fn container_length(length: Range) -> Constraint {
         if !matches!(length, Range::IntegerNonNegative(_, _)) {
-            panic!("container_length constraint  must have a range of type IntegerNonNegative")
+            panic!("container_length constraint must have a range of type IntegerNonNegative")
         }
         Constraint::ContainerLength(ContainerLengthConstraint::new(length))
     }
@@ -91,7 +91,7 @@ impl Constraint {
     /// Creates a [Constraint::ByteLength] from a [Range] specifying a length range.
     pub fn byte_length(length: Range) -> Constraint {
         if !matches!(length, Range::IntegerNonNegative(_, _)) {
-            panic!("byte_length constraint  must have a range of type IntegerNonNegative")
+            panic!("byte_length constraint must have a range of type IntegerNonNegative")
         }
         Constraint::ByteLength(ByteLengthConstraint::new(length))
     }
@@ -99,7 +99,7 @@ impl Constraint {
     /// Creates a [Constraint::CodePointLength] from a [Range] specifying a length range.
     pub fn codepoint_length(length: Range) -> Constraint {
         if !matches!(length, Range::IntegerNonNegative(_, _)) {
-            panic!("codepoint_length constraint  must have a range of type IntegerNonNegative")
+            panic!("codepoint_length constraint must have a range of type IntegerNonNegative")
         }
         Constraint::CodepointLength(CodepointLengthConstraint::new(length))
     }
@@ -141,7 +141,7 @@ impl Constraint {
     /// Creates a [Constraint::Precision] from a [Range] specifying a precision range.
     pub fn precision(precision: Range) -> Constraint {
         if !matches!(precision, Range::IntegerNonNegative(_, _)) {
-            panic!("precision constraint  must have a range of type IntegerNonNegative")
+            panic!("precision constraint must have a range of type IntegerNonNegative")
         }
         Constraint::Precision(PrecisionConstraint::new(precision))
     }
@@ -149,7 +149,7 @@ impl Constraint {
     /// Creates a [Constraint::Scale] from a [Range] specifying a precision range.
     pub fn scale(scale: Range) -> Constraint {
         if !matches!(scale, Range::Integer(_, _)) {
-            panic!("scale constraint  must have a range of type Integer")
+            panic!("scale constraint must have a range of type Integer")
         }
         Constraint::Scale(ScaleConstraint::new(scale))
     }
@@ -157,7 +157,7 @@ impl Constraint {
     /// Creates a [Constraint::TimestampPrecision] from a [Range] specifying a precision range.
     pub fn timestamp_precision(precision: Range) -> Constraint {
         if !matches!(precision, Range::TimestampPrecision(_, _)) {
-            panic!("timestamp_precision constraint  must have a range of type TimestampPrecision")
+            panic!("timestamp_precision constraint must have a range of type TimestampPrecision")
         }
         Constraint::TimestampPrecision(TimestampPrecisionConstraint::new(precision))
     }
--- a/src/constraint.rs
+++ b/src/constraint.rs
@@ -1,6 +1,6 @@
 use crate::isl::isl_constraint::IslConstraint;
 use crate::isl::isl_type_reference::IslTypeRef;
-use crate::isl::util::{Annotation, Range, TimestampPrecisionValue};
+use crate::isl::util::{Annotation, Range, TimestampPrecision};
 use crate::result::{IonSchemaResult, ValidationResult};
 use crate::system::{PendingTypes, TypeId, TypeStore};
 use crate::types::{TypeDefinition, TypeValidator};
@@ -82,16 +82,25 @@ impl Constraint {
 
     /// Creates a [Constraint::ContainerLength] from a [Range] specifying a length range.
     pub fn container_length(length: Range) -> Constraint {
+        if !matches!(length, Range::IntegerNonNegative(_, _)) {
+            panic!("container_length constraint  must have a range of type IntegerNonNegative")
+        }
         Constraint::ContainerLength(ContainerLengthConstraint::new(length))
     }
 
     /// Creates a [Constraint::ByteLength] from a [Range] specifying a length range.
     pub fn byte_length(length: Range) -> Constraint {
+        if !matches!(length, Range::IntegerNonNegative(_, _)) {
+            panic!("byte_length constraint  must have a range of type IntegerNonNegative")
+        }
         Constraint::ByteLength(ByteLengthConstraint::new(length))
     }
 
     /// Creates a [Constraint::CodePointLength] from a [Range] specifying a length range.
     pub fn codepoint_length(length: Range) -> Constraint {
+        if !matches!(length, Range::IntegerNonNegative(_, _)) {
+            panic!("codepoint_length constraint  must have a range of type IntegerNonNegative")
+        }
         Constraint::CodepointLength(CodepointLengthConstraint::new(length))
     }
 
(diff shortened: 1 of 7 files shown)


src/constraint.rs
```diff
@@ -1406,3 +1420,67 @@ impl ConstraintValidator for ScaleConstraint {
...
+                };
+                return Err(Violation::new(
+                    "timestamp_precision",
+                    ViolationCode::TypeMismatched,
+                    &error_message,
+                ));
+            }
+        };
+
+        let value_precision = TimestampPrecisionValue::from_timestamp(&timestamp_value);
```
popematt, line 1463: issue—Clippy says "this expression creates a reference which is immediately dereferenced by the compiler".

src/isl/isl_constraint.rs
```diff
@@ -103,6 +104,11 @@ impl IslConstraint {
         IslConstraint::CodepointLength(length)
     }
 
+    /// Creates an [IslConstraint::TimestampPrecision] using the range specified in it
+    pub fn timestamp_precision(precision: Range) -> IslConstraint {
```
popematt, line 108: question—Is it possible to change this function signature so that you can't pass in the wrong type of `Range`?

popematt (commented):
Overall, it looks good, but I'm not sure how the Timestamp precision range is constructed. Would a user construct it as an int range? Should we provide an enum of `year`, `month`, etc.?

src/constraint.rs
```diff
@@ -1406,3 +1420,67 @@ impl ConstraintValidator for ScaleConstraint {
...
+
+        // get isl timestamp precision as a range
+        let precision_range: &Range = self.timestamp_precision();
+
+        // return a Violation if the value didn't follow timestamp precision constraint
+        if !precision_range
+            .contains(&(timestamp_value.to_owned()).into())
+            .unwrap()
+        {
```
zslayton, line 1472: Could you describe what's happening here? This struck me as a bit redundant:

```rust
//         v-- `timestamp_value` is a `&Timestamp`
    &(timestamp_value.to_owned()) // <-- that we turn into an owned `Timestamp`
//  ^-- before turning it back into a `&Timestamp`
```

Am I misunderstanding? Given the signature of [`Range::contains`](https://doc.rust-lang.org/std/ops/struct.Range.html#method.contains), I'm also not sure why the `unwrap()` is needed at the end.

src/isl/util.rs
```diff
@@ -188,6 +191,39 @@ impl Range {
...
+                let is_in_lower_bound = match start {
+                    Min => true,
+                    Value(start_value, boundary_type) => match start_value {
+                        TimestampPrecision(min_value) => match boundary_type {
+                            RangeBoundaryType::Inclusive => min_value <= &value,
+                            RangeBoundaryType::Exclusive => min_value < &value,
+                        },
+                        _ => unreachable!("TimestampPrecision range can only have timestamp precisions as lower and upper range boundary value"),
+                    },
+                    Max => unreachable!("Cannot have 'Max' as the lower range boundary")
```
zslayton, line 211: [`unreachable!`](https://doc.rust-lang.org/std/macro.unreachable.html) statements are meant to be used for branches that will never be encountered unless the library has a bug. Is that the case here, or are we using it to validate input? If it's the latter, we should either enforce these requirements earlier in the code (during construction of the values, say) or return an `IonSchemaError` here that explains what they did wrong.

src/isl/util.rs
```diff
@@ -188,6 +191,39 @@ impl Range {
...
+                    Value(end_value, boundary_type) => match end_value {
+                        TimestampPrecision(max_value) => match boundary_type {
+                            RangeBoundaryType::Inclusive => max_value >= &value,
+                            RangeBoundaryType::Exclusive => max_value > &value,
+                        },
+                        _ => unreachable!("TimestampPrecision range can only have timestamp precisions as lower and upper range boundary value"),
+                    }
+                };
+                Ok(is_in_upper_bound && is_in_lower_bound)
+            }
```
zslayton, line 226: I notice that this method is getting to be ~200 lines long. It might be nice to break some of the `match` arms' logic out into helper methods.

src/isl/util.rs
```diff
@@ -396,6 +450,41 @@ impl From<&IntegerValue> for Range {
...
 
+/// Provides `Range` for given `&str`
+impl TryFrom<&str> for Range {
+    type Error = IonSchemaError;
+
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        let timestamp_precision = match value {
+            "year" => TimestampPrecisionValue::Year,
+            "month" => TimestampPrecisionValue::Month,
+
```
zslayton, line 461:
```suggestion
```

src/isl/util.rs
```diff
@@ -446,10 +550,43 @@ impl RangeBoundaryValue {
 
         match value.ion_type() {
             IonType::Symbol => {
+                use TimestampPrecisionValue::*;
                 let sym = try_to!(try_to!(value.as_sym()).text());
                 match sym {
                     "min" => Ok(RangeBoundaryValue::Min),
                     "max" => Ok(RangeBoundaryValue::Max),
+                    "year" => Ok(RangeBoundaryValue::Value(
+                        RangeBoundaryValueType::TimestampPrecision(Year as i64),
```
zslayton, line 559: Having named `TimestampPrecisionValue` variants feels like a nice feature; why are we discarding the name and using the enum variant index? (And do we want that index to be signed?)

src/isl/util.rs
```diff
@@ -546,3 +687,30 @@ impl Annotation {
...
+pub enum TimestampPrecisionValue {
+    Year = -4,
+    Month = -3,
+    Day = -2,
+    Minute = -1,
+    Second = 0,
+    Millisecond = 3,
+    Microsecond = 6,
+    Nanosecond = 9,
+}
```
zslayton, line 701: This works, but I think it would be safer and a bit more self-describing to do something like this:

```rust
pub enum TimestampPrecisionValue {
    Year,
    Month,
    Day,
    Minute,
    Second,
    Millisecond,
    Microsecond,
    Nanosecond,
}

impl TimestampPrecisionValue {
  pub const fn exponent(&self) -> i64 {
    match self {
      Year => -4,
      Month = -3,
      // ... and so on.
    }
  }
}
```
(Maybe there's a better name than `exponent`?)

It's obvious to me what the subsecond precision values do, but I don't know what `Minute=-1`, `Day=-2`, etc means. If those aren't being used meaningfully anywhere, then `exponent` could return `Option<i64>` (or an unsigned type if we don't need negatives.)

src/isl/util.rs
```diff
@@ -546,3 +687,30 @@ impl Annotation {
         }
     }
 }
+
+#[derive(Debug, Clone, PartialEq, PartialOrd)]
+pub enum TimestampPrecisionValue {
```
zslayton, line 692: I like the name `TimestampPrecision` (dropping the `Value`). I realize this conflicts with the `Range` variant name, but I think that's ok. We do that in other places too.

src/isl/util.rs
desaikd: These checks are already performed while constructing the `Range` this method is for checking if a range contains the value so all these checks/errors here would be `unreachable!` as they were already verified at the time of range creation.

src/isl/isl_constraint.rs
desaikd: I agree that there should be a better representation to `Range` to resolve this issue. I will add a short-term solution that Zack suggested to panic for invalid range types for now.
I will create an issue to keep track of this.

desaikd: Reference issue: https://github.com/amzn/ion-schema-rust/issues/89

src/isl/util.rs
desaikd: I will rename this to be `scale` as that would make more sense than `from_timestamp`.

src/isl/util.rs
desaikd: I will handle this when changing the Range implementation with #89.

src/constraint.rs
desaikd: I was using `unwrap()` because contains would return an `IonSchemResult`. These `IonSchemaResult` returns  a single kind of `Error` which checks if the given input value(`OwnedElement`) has the same type as the range type to perform contains. Probably changing `Range` to be `Range<T>` could remove these type checks and also for all constraint the type of the data being passed to `contains` already checked before its passed so its safe to `unwrap` these.

zslayton: > Probably changing Range to be Range<T> could remove these type checks and also for all constraint the type of the data being passed to contains already checked before its passed so its safe to unwrap these.

👍

src/constraint.rs
```diff
@@ -82,16 +82,25 @@ impl Constraint {
 
     /// Creates a [Constraint::ContainerLength] from a [Range] specifying a length range.
     pub fn container_length(length: Range) -> Constraint {
+        if !matches!(length, Range::IntegerNonNegative(_, _)) {
+            panic!("container_length constraint  must have a range of type IntegerNonNegative")
```
zslayton, line 86:
```suggestion
            panic!("container_length constraint must have a range of type IntegerNonNegative")
```

src/constraint.rs
```diff
@@ -82,16 +82,25 @@ impl Constraint {
...
+        }
         Constraint::ContainerLength(ContainerLengthConstraint::new(length))
     }
 
     /// Creates a [Constraint::ByteLength] from a [Range] specifying a length range.
     pub fn byte_length(length: Range) -> Constraint {
+        if !matches!(length, Range::IntegerNonNegative(_, _)) {
+            panic!("byte_length constraint  must have a range of type IntegerNonNegative")
```
zslayton, line 94:
```suggestion
            panic!("byte_length constraint must have a range of type IntegerNonNegative")
```

I see this was copy/pasted for each function, so I'll refrain from commenting on each one.

src/isl/util.rs
zslayton: In the updated version you've added a `scale` method, but we're still using `Year = -4, Month = -3...` in the enum definition. I'd like to avoid that and instead define the `i64` values inside the body of the `scale` method. If we use the current system, it makes `Year = -4` part of the public API even though that value is arbitrary.

zslayton (commented):
Some superficial notes below. I'd like to change the definition of the `TimestampPrecision` enum before we merge this.

src/isl/util.rs
```diff
@@ -698,33 +699,85 @@ pub enum TimestampPrecision {
...
-            TimestampPrecision::Month => -3,
-            TimestampPrecision::Day => -2,
-            TimestampPrecision::Minute => -1,
-            TimestampPrecision::Second => 0,
-            TimestampPrecision::Millisecond => 3,
-            TimestampPrecision::Microsecond => 6,
-            TimestampPrecision::Nanosecond => 9,
-        }
+impl PartialOrd for TimestampPrecision {
+    fn partial_cmp(&self, other: &TimestampPrecision) -> Option<Ordering> {
```
zslayton, line 728: This arrangement works fine and is very explicit.

An alternative that's closer to your original implementation would be to map `self` and `other` to integers within this method and then use `.cmp` to produce the expected ordering. Something like:

```rust
let self_value = match self {
    Year => -4,
    Month => -3,
    // ...and so on
}
let other_value = match other {
    Year => -4,
    Month => -3,
    // ...and so on
}
self_value.cmp(other_value)
```

zslayton (approved):
One suggestion to consider below.

Issue #93, opened 2022-06-17
schema validation returns violation for self referencing types


desaikd:
All the types `in ion-schema-rust` a unique type_id. Currently, the self referencing types `type_id` is not calculated correctly resulting in a `Violation`.

PR #94, opened 2022-06-17
Fixes the issue with calculating self referencing type ids

desaikd:
*Issue #93*

*Description of changes:*
This PR fixes an issue with self referencing type id calculation.

### DIFF
--- a/src/system.rs
+++ b/src/system.rs
@@ -349,6 +349,10 @@ impl PendingTypes {
 
     /// Adds parent information storing the name and possible TypeId of the parent
     pub(crate) fn add_parent(&mut self, name: String, type_store: &mut TypeStore) {
+        // Parent information is used when resolving a self referencing type
+        // while we resolve a type using the PendingTypes (a temporary type store used while we resolve a type definition)
+        // the type id for any type definition should be the PendingType's types_by_id length in  + TypeStore's types_by_id length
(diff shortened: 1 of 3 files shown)


src/system.rs
```diff
@@ -348,8 +348,8 @@ impl PendingTypes {
     }
 
     /// Adds parent information storing the name and possible TypeId of the parent
-    pub(crate) fn add_parent(&mut self, name: String) {
-        self.parent = Some((name, self.types_by_id.len()))
+    pub(crate) fn add_parent(&mut self, name: String, type_store: &mut TypeStore) {
+        self.parent = Some((name, self.types_by_id.len() + type_store.types_by_id.len()))
```
popematt, line 352: Can you update either the documentation here or the PR description to explain _why_ this fixes it? I'm not sure I quite understand what's going on here.

desaikd: Previously `type_id` for a parent type was calculated incorrectly resulting to incorrect `Violation`. 
While we are resolving a type definition, the correct `type_id` for any type definition would be `PendingTypes`' `types_by_id` length + `TypeStore`'s `types_by_id` length.
`PendinTypes` is a temporary type storage used while resolving a type definition which then shifts all its contents to `TypeStore`.
`TypeStore` is the common place to store all the types.

popematt (approved):

Issue #89, opened 2022-06-02
Change enum `Range` implementation to generic types `Range<T>` 


desaikd:
Current implementation of `Range` doesn't allow to pass in a particular type of Range to a method. Having a generic implementation of `Range<T>` could resolve this issue.

```rust
// Generic range implementation
#[derive(Debug, Clone, PartialEq)]
pub struct Range<T> {
    start: RangeBoundaryValue<T>,
    end: RangeBoundaryValue<T>,
}

// Usage of this range in a method
fn timestamp_precision<TimestampPrecision>(precision: Range<TimestampPrecision>) {
     ...
}
```

This issue was raised while creating `timestamp_precision` constraint. 
More information can be found here: https://github.com/amzn/ion-schema-rust/pull/87#discussion_r887396581

desaikd:
This has been resolved with #95

PR #95, opened 2022-07-07
Generic Range implementation

desaikd:
*Issue #89*

*Description of changes:*
This PR works on generic implementation of ranges. 

*Grammar:*
```ANTLR

<RANGE_TYPE> ::= <DECIMAL>
               | <FLOAT>
               | <INT>
               | <NUMBER>
               | <TIMESTAMP>
               | <TIMESTAMP_PRECISION_VALUE>

<EXCLUSIVITY> ::= exclusive::
                | ""

<RANGE<RANGE_TYPE>> ::= range::[ <EXCLUSIVITY><RANGE_TYPE>, <EXCLUSIVITY><RANGE_TYPE> ]
                      | range::[ min, <EXCLUSIVITY><RANGE_TYPE> ]
                      | range::[ <EXCLUSIVITY><RANGE_TYPE>, max ]
```

*Ion Schema specification:*
https://amzn.github.io/ion-schema/docs/spec.html#grammar

*List of changes:*
* upgrades `ion-rs` dependency to `v0.12.0`
* Adds `RangeImpl<T>` generic range implementation
* Adds implementation of `Range` which is a wrapper around `RangeImpl<T>`
* Adds implementation of `TypedRangeBoundaryValue` which is a wrapper
around generic `RangeBoundaryVlaue<T>`
* all other changes are similar from previous  `Range` implementation but
are moved in this new module `isl_range`
* removes previous range implementation
* all the programmatic construction of `Constraint` and `IslConstraint`
now uses generic `RangeImpl<T>` where `T` is specific to the constraint
* now the new range implementation supports a check for `start > end`
hence some tests which were previously commented are now un-commented

*Tests:*
* removes some tests from skip-list (`start > end` kind of range tests)
* adds and modifies unit tests for ranges implementation.


src/isl/isl_range.rs
```diff
@@ -0,0 +1,850 @@
...
+            }
+            (TypedRangeBoundaryValue::Max, _) => {
+                invalid_schema_error("Lower range boundary value must not be max")
+            }
+            (_, TypedRangeBoundaryValue::Min) => {
+                invalid_schema_error("Upper range boundary value must not be min")
+            }
+            (_, TypedRangeBoundaryValue::Integer(_)) | (TypedRangeBoundaryValue::Integer(_), _) => {
+                Ok(Range::Integer(RangeImpl::<Integer>::int_range_from_typed_boundary_value(start, end)?))
+            }
```
zslayton, line 336: We can eliminate a lot of the boilerplate in this `match` statement if we check that `start.range_type() == end.range_type()` before we run it.

src/isl/isl_range.rs
```diff
@@ -0,0 +1,850 @@
...
+                RangeImpl::range(v1, v2)
+            }
+            (TypedRangeBoundaryValue::Min, TypedRangeBoundaryValue::Integer(v2)) => {
+                RangeImpl::range(RangeBoundaryValue::Min, v2)
+            }
+            (TypedRangeBoundaryValue::Integer(v1), TypedRangeBoundaryValue::Max) => {
+                RangeImpl::range(v1, RangeBoundaryValue::Max)
+            }
+            (TypedRangeBoundaryValue::Integer(Value(v1, _)), _) => {
+                invalid_schema_error("Range boundaries should have same types")
```
zslayton, line 432:
```suggestion
                invalid_schema_error("Range boundaries must have the same types")
```

src/isl/util.rs
```diff
@@ -1,809 +1,10 @@
-use crate::external::ion_rs::Integer;
```
zslayton, line 1: Thanks for breaking up `util.rs` 👍

src/isl/isl_range.rs
```diff
@@ -0,0 +1,850 @@
...
+                RangeImpl::range(v1, v2)
+            }
+            (TypedRangeBoundaryValue::Min, TypedRangeBoundaryValue::Decimal(v2)) => {
+                RangeImpl::range(RangeBoundaryValue::Min, v2)
+            }
+            (TypedRangeBoundaryValue::Decimal(v1), TypedRangeBoundaryValue::Max) => {
+                RangeImpl::range(v1, RangeBoundaryValue::Max)
+            }
+            (TypedRangeBoundaryValue::Decimal(Value(v1, _)), _) => {
+                invalid_schema_error("Range boundaries should have same types")
```
zslayton, line 532: It feels like we should be able to eliminate some of the repetitive boilerplate in these `match` statements.

src/constraint.rs
```diff
@@ -83,27 +84,22 @@ impl Constraint {
...
+        Constraint::ByteLength(ByteLengthConstraint::new(Range::IntegerNonNegative(length)))
     }
 
     /// Creates a [Constraint::CodePointLength] from a [Range] specifying a length range.
-    pub fn codepoint_length(length: Range) -> Constraint {
-        if !matches!(length, Range::IntegerNonNegative(_, _)) {
-            panic!("codepoint_length constraint must have a range of type IntegerNonNegative")
-        }
-        Constraint::CodepointLength(CodepointLengthConstraint::new(length))
+    pub fn codepoint_length(length: RangeImpl<usize>) -> Constraint {
```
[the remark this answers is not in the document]
desaikd: Oh yes, I had this in my TODO, but I was thinking to change `RangeImpl` to be `Range` and `Range` to be `TypedRange`. Thoughts?

src/isl/isl_range.rs
desaikd: The Problem with that would be that `TypedRangeBoundaryValue` has following definition which includes `Min`, `Max`:
```rust
pub(crate) enum TypedRangeBoundaryValue {
    Min,
    Max,
    Integer(RangeBoundaryValue<Integer>),
    IntegerNonNegative(RangeBoundaryValue<usize>),
    TimestampPrecision(RangeBoundaryValue<TimestampPrecision>),
    Float(RangeBoundaryValue<f64>),
    Decimal(RangeBoundaryValue<Decimal>),
    Number(RangeBoundaryValue<Number>),
    Timestamp(RangeBoundaryValue<Timestamp>),
}
```

src/isl/isl_range.rs
```diff
@@ -0,0 +1,850 @@
...
+    Number(RangeImpl<Number>),
+}
+
+impl Range {
+    /// Provides integer range with given start and end values
+    /// Returns an [IonSchemaError] if start value is less than end value
+    pub fn int_range(
+        start: RangeBoundaryValue<Integer>,
+        end: RangeBoundaryValue<Integer>,
+    ) -> IonSchemaResult<Range> {
```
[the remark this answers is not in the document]
desaikd: I like the second option you suggested here. Although we might need to shift to `TryFrom` implementation to just allow range types that re supported by Ion Schema.

src/isl/mod.rs
```diff
@@ -310,10 +317,10 @@ mod isl_tests {
                     "#),
         IslType::anonymous(
             [IslConstraint::valid_values_with_range(
-                Range::number_range(
-                    RangeBoundaryValue::Value((&IntegerValue::I64(1)).into(), RangeBoundaryType::Inclusive),
-                    RangeBoundaryValue::Value((&Decimal::new(55, -1)).try_into().unwrap(), RangeBoundaryType::Inclusive)
-                ).unwrap())
+                NumberRange::new(
+                    Number::from(&IntegerValue::I64(1)).into(),
+                    Number::try_from(&Decimal::new(55, -1)).unwrap().into()
```
zslayton, line 322: I was curious why `&Integer` could be converted to a `Number` infallibly while `&Decimal` had to use `try_from`. Looking at the source:

https://github.com/amzn/ion-schema-rust/blob/b5b643593dc1260e4048531f1ef61ff391d31066/src/isl/util.rs#L596-L604

`Decimal` cannot be converted to a `BigDecimal` if it's a negative zero, but will succeed for all other values. Since `Number` is being used in ranges and for ordering, I think its implementation could detect negative zero and return `BigDecimal::zero()` instead, making it infallible. What do you think?

src/isl/mod.rs
```diff
@@ -366,75 +373,75 @@ mod isl_tests {
...
+            ).unwrap().into()
         )
     )]
-    fn owned_struct_to_range(range1: IonSchemaResult<Range>, range2: IonSchemaResult<Range>) {
+    fn owned_struct_to_range(range1: Range, range2: Range) {
         // determine that both the ranges are created with no errors
-        assert!(range1.is_ok());
-        assert!(range2.is_ok());
+        // assert!(range1.is_ok());
+        // assert!(range2.is_ok());
```
zslayton, line 441:
```suggestion
```

src/isl/mod.rs
```diff
@@ -366,75 +373,75 @@ mod isl_tests {
...
-            )
+            ).unwrap(),
+            NumberRange::new(
+                Number::from(&IntegerValue::I64(1)).into(),
+                Number::try_from(&Decimal::new(55, -1)).unwrap().into()
+            ).unwrap().into()
         )
     )]
-    fn owned_struct_to_range(range1: IonSchemaResult<Range>, range2: IonSchemaResult<Range>) {
+    fn owned_struct_to_range(range1: Range, range2: Range) {
```
zslayton, line 438: I think you could change `range2` to be `impl Into<Range>`, letting you drop the `.into()` calls in each of the `case::`s above.

src/isl/isl_range.rs
```diff
@@ -14,6 +14,81 @@ use std::cmp::Ordering;
...
+pub type FloatRange = RangeImpl<f64>;
+
+impl FloatRange {
+    pub fn new(
+        start: RangeBoundaryValue<f64>,
+        end: RangeBoundaryValue<f64>,
+    ) -> IonSchemaResult<FloatRange> {
+        RangeImpl::range(start, end)
+    }
+}
```
zslayton, line 42: As a `TODO`, I think we should make `RangeImpl::range(...)` and each of these `new(...)` methods generic over `Into<RangeBoundaryValue<_>>`.

That way, calls like this:
```rust
FloatRange::new(
   2e1.into(),
   5e1.into()
)
```
can be reduced to:
```rust
FloatRange::new(2e1, 5e1)
```

src/isl/isl_range.rs
```diff
@@ -14,6 +14,81 @@ use std::cmp::Ordering;
...
+    pub fn new(
+        start: RangeBoundaryValue<Integer>,
+        end: RangeBoundaryValue<Integer>,
+    ) -> IonSchemaResult<IntegerRange> {
+        RangeImpl::range(start, end)
+    }
+}
+
+/// Provides IntegerNonNegativeRange type to be used to create integer non negative ranges.
+pub type IntegerNonNegativeRange = RangeImpl<usize>;
```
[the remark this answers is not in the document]
desaikd: That is a typo while adding the change.

src/isl/mod.rs
desaikd: Yes that makes sense

src/isl/isl_range.rs
desaikd: Yes, I will add this change.

src/isl/isl_range.rs
```diff
@@ -14,81 +14,27 @@ use std::cmp::Ordering;
...
 
-impl IntegerRange {
-    pub fn new(
-        start: RangeBoundaryValue<Integer>,
-        end: RangeBoundaryValue<Integer>,
-    ) -> IonSchemaResult<IntegerRange> {
-        RangeImpl::range(start, end)
-    }
-}
+/// Provides a type to be used to create integer non negative ranges.
```
zslayton, line 20:
```suggestion
/// Provides a type to be used to create non negative integer ranges.
```

zslayton (approved):
Nice!

Issue #12, opened 2021-08-03
Implementation of `nullable` type references


desaikd:
Current implementation of `TypeRef` doesn't include `nullable` type references. Add logic for nullable type references when converting type reference from given `OwnedElement` in the `TryFrom<&OwnedElement>`.
```rust
impl TryFrom<&OwnedElement> for TypeRef {
    type Error = IonSchemaError;

    fn try_from(value: &OwnedElement) -> Result<Self, Self::Error> {
         // add a check on the annotations for `nullable` type references
    }
}
```

desaikd:
closed with #100

PR #100, opened 2022-07-19
adds changes for 'nullable' annotated type references

desaikd:
*Issue #12*

*Description of changes:*
This PR works on implementation of type references that are annotated with `nullable`.

*List of changes:*
- Adds a check for `nullable` annotation on type references (only allows for built in type references e.g. `{ type: int }`)

*Test:*
Adds unit tests for nullable type references and removes skip list tests.

### DIFF
--- a/src/isl/isl_type_reference.rs
+++ b/src/isl/isl_type_reference.rs
@@ -6,7 +6,8 @@ use crate::result::{
 };
 use crate::system::{PendingTypes, TypeId, TypeStore};
 use crate::types::TypeDefinitionImpl;
-use ion_rs::value::owned::OwnedElement;
+use ion_rs::value::owned::{text_token, OwnedElement};
+use ion_rs::value::reader::{element_reader, ElementReader};
 use ion_rs::value::{Element, Struct, SymbolToken};
 use ion_rs::IonType;
 
@@ -23,7 +24,6 @@ pub enum IslTypeRef {
     Anonymous(IslTypeImpl),
 }
 
-// TODO: add a check for nullable type reference
 impl IslTypeRef {
     /// Creates a [IslTypeRef::Named] using the [IonType] referenced inside it
     pub fn named<A: Into<String>>(name: A) -> IslTypeRef {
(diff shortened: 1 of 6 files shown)


src/isl/isl_type_reference.rs
```diff
@@ -35,6 +35,31 @@ impl IslTypeRef {
         IslTypeRef::Anonymous(IslTypeImpl::new(None, constraints.into()))
     }
 
+    /// Provides a string representing a definition of a nullable built in type reference
+    // these are the built types that rae annotated with `nullable`
+    fn get_nullable_type_reference_definition(type_ref_name: &str) -> IonSchemaResult<&str> {
+        Ok(match type_ref_name {
```
popematt, line 41: There's a few more "core" types.

This should also include `lob`, `number`, `text`, and eventually `document`.

It should _maybe_ include `nothing` and `any` for completeness, but I'm not sure about them. What do you think?

src/isl/isl_type_reference.rs
```diff
@@ -47,13 +72,25 @@ impl IslTypeRef {
...
                     .text()
                     .ok_or_else(|| {
                         invalid_schema_error_raw(
                             "a base or alias type reference symbol doesn't have text",
                         )
-                    }).map(|type_name| IslTypeRef::Named(type_name.to_owned()))
+                    })?;
+
+                // check for nullable type reference
+                if value.annotations().any(|a| a == &text_token("nullable")) {
```
zslayton, line 85: I think this allocates a new `String` (`"nullable"`) on each pass, which is expensive. Instead, see if `a`'s text matches `"nullable"` (the `&str`).

src/isl/isl_type_reference.rs
```diff
@@ -47,13 +72,25 @@ impl IslTypeRef {
...
+                    let built_in_type_def = IslTypeRef::get_nullable_type_reference_definition(type_name)?;
+                    let value = &element_reader()
+                        .read_one(built_in_type_def.as_bytes())
+                        .map_err(|_| invalid_schema_error_raw(format!("Can not parse nullable type reference for given built in type: {}", type_name)))?;
+                    return IslTypeRef::from_ion_element(value, inline_imported_types);
```
zslayton, line 90: Not a blocker: it seems like we should be able to cache these instead of parsing/building them each time they're referenced in the schema.

src/schema.rs
```diff
@@ -309,6 +309,25 @@ mod schema_tests {
...
+                    null.decimal
+                    a
+                    "hello"
+                    false
+                "#),
+            load_schema_from_text(r#" // For a schema with named type and `nullable` annotation as below: 
+                    type:: { name: my_int, type: nullable::int }
+                "#),
+            "my_int"
+        ),
```
zslayton, line 330: Does `ion-schema-tests` cover `nullable::` on the other types?

desaikd: No, only on built in types.

src/isl/isl_type_reference.rs
desaikd: Sounds good. Will add it for `lob`, `number`, `text` and  a TODO for `document` (document type doesn't exist yet).
I will also add implementation of `nothing` with this PR and add it to this list for `nullable` annotation type references.

src/isl/isl_type_reference.rs
desaikd: Sounds good! I would just add the `OwnedElement` for them directly.

src/isl/isl_type_reference.rs
popematt: So, I don't think the spec is clear whether `nullable::nothing` _should_ be allowed. (That's why I said I'm not sure about it.) 

I don't think that `nullable::nothing` and `nullable::any` are actually useful. In theory, `nullable::nothing` is equivalent to `$null`, and `nullable::any` is equivalent to `$any`. However, if the spec requires it, then we should include it whether or not we think it's useful.

In practice, the Kotlin implementation breaks for both `nullable::nothing` and `nullable::any`, when it encounters any typed null.

src/schema.rs
zslayton: That's what I meant, the non-`int` built in types. I wanted to make sure our tests covered `float`, `list`, etc.

src/isl/isl_type_reference.rs
popematt: Maybe another reason to implement https://github.com/amzn/ion-rust/issues/400?

src/schema.rs
desaikd: Oh okay. Sure I will add some tests here for `float`, `list`, `string` etc.

src/isl/isl_type_reference.rs
desaikd: I agree. I think it can be used on many places in `ion-schema-rust`. I will create a PR for implementation of `has_annotation()`  along with `len()` for container type of `OwnedElement`s  because that is also used multiple times in `ion-schema-rust`.

src/isl/isl_type_reference.rs
desaikd: Got it! I think in that case as the spec is not clear on this we can probably just include those eventhough they just point to `$null` or `$any` just to support it for all built in/ derived types.

src/isl/isl_type_reference.rs
desaikd: Reference issue: https://github.com/amzn/ion-schema-rust/issues/101

src/isl/isl_type_reference.rs
popematt: Sure.

popematt (approved):

zslayton (approved):

Issue #103, opened 2022-07-28
adds `document` type implementation


desaikd:
As per [Ion schema spec](https://amzn.github.io/ion-schema/docs/spec.html#type-system) a `document` type definition:

> document: represents a series of top-level values

Add a wrapper around `OwnedElement`s that supports document type:
```rust
pub enum IonSchemaElement {
    Element(OwnedElement),
    Document(Vec<OwnedElement>),
}
```

PR #104, opened 2022-07-28
adds implementation for `document` type

desaikd:
*Issue #103*

*Description of changes:*
This PR adds implementation of `document` type.

*Ion Schema specification:*
https://amzn.github.io/ion-schema/docs/spec.html#type-system

> document: represents a series of top-level values

*List of changes:*
* adds `IonSchemaElement` implementation
* `TypeRef` validate method now uses `IonSchemaElement` as input
* `TypeDefinitionImpl`'s `parse_from_isl_type` method now returns a `TypeId`
instead of `TypeDefinitionImpl`
* adds `nullable` annotation check for non built in type references to
return an error
* Remove `document` type tests from skip list
* adds `id()` for `TypeRef` to get the `TypeId` of that `TypeRef`
* adds doc examples

*Tests:*
adds unit tests for document type (removes previously unsupported tests from skip list)

### DIFF
--- a/src/constraint.rs
+++ b/src/constraint.rs
@@ -679,7 +679,7 @@ impl OrderedElementsConstraint {
         while values_iter.peek() != None && occurs_range.contains(&(count + 1).into()) {
             // don't consume it until we know it's valid for the type
             if let Some(value) = values_iter.peek() {
-                let schema_element: IonSchemaElement = value.to_owned().into();
+                let schema_element: IonSchemaElement = (*value).into();
                 if type_def.is_valid(&schema_element, type_store) {
                     let _ = values_iter.next(); // consume it as it is valid
                     count += 1;
@@ -807,35 +807,11 @@ impl ConstraintValidator for FieldsConstraint {
     fn validate(&self, value: &IonSchemaElement, type_store: &TypeStore) -> ValidationResult {
         let mut violations: Vec<Violation> = vec![];
 
-        // Create a peekable iterator for given struct
-        let ion_struct = match &value {
-            IonSchemaElement::SingleElement(element) => match element.as_struct() {
-                None => {
-                    return Err(Violation::with_violations(
-                        "fields",
-                        ViolationCode::TypeMismatched,
-                        &format!(
-                            "expected struct ion found {}",
-                            if element.is_null() {
-                                format!("{:?}", element)
-                            } else {
-                                format!("{}", element.ion_type())
-                            }
-                        ),
-                        violations,
-                    ));
-                }
-                Some(ion_struct) => ion_struct,
-            },
-            IonSchemaElement::Document(document) => {
-                return Err(Violation::with_violations(
-                    "fields",
-                    ViolationCode::TypeMismatched,
-                    "expected struct ion found document type",
-                    violations,
-                ));
-            }
-        };
+        // get struct value
+        let ion_struct = value
+            .expect_element_of_type("fields", &[IonType::Struct])?
+            .as_struct()
+            .unwrap();
 
         // Verify if open content exists in the struct fields
         if !self.open_content() {
@@ -1054,36 +1030,11 @@ impl ByteLengthConstraint {
 impl ConstraintValidator for ByteLengthConstraint {
     fn validate(&self, value: &IonSchemaElement, type_store: &TypeStore) -> ValidationResult {
         // get the size of given bytes
-        let size = match value {
-            IonSchemaElement::SingleElement(element) => {
-                match element.as_bytes() {
-                    Some(bytes) => bytes.len(),
-                    _ => {
-                        // return Violation if value is not an clob/blob
-                        return Err(Violation::new(
-                            "byte_length",
-                            ViolationCode::TypeMismatched,
-                            &format!(
-                                "expected a clob/blob but found {}",
-                                if element.is_null() {
-                                    format!("{:?}", element)
-                                } else {
-                                    format!("{}", element.ion_type())
-                                }
-                            ),
-                        ));
-                    }
-                }
-            }
-            IonSchemaElement::Document(document) => {
-                // return Violation if value is not an clob/blob
-                return Err(Violation::new(
-                    "byte_length",
-                    ViolationCode::TypeMismatched,
-                    "expected a clob/blob but found document",
-                ));
-            }
-        };
+        let size = value
+            .expect_element_of_type("byte_length", &[IonType::Blob, IonType::Clob])?
+            .as_bytes()
+            .unwrap()
+            .len();
 
         // get isl length as a range
         let length_range: &Range = self.length();
@@ -1121,36 +1072,12 @@ impl CodepointLengthConstraint {
 impl ConstraintValidator for CodepointLengthConstraint {
     fn validate(&self, value: &IonSchemaElement, type_store: &TypeStore) -> ValidationResult {
         // get the size of given string/symbol Unicode codepoints
-        let size = match value {
-            IonSchemaElement::SingleElement(element) => {
-                match element.as_str() {
-                    Some(text) => text.chars().count(),
-                    _ => {
-                        // return Violation if value is not string/symbol
-                        return Err(Violation::new(
-                            "codepoint_length",
-                            ViolationCode::TypeMismatched,
-                            &format!(
-                                "expected a string/symbol but found {}",
-                                if element.is_null() {
-                                    format!("{:?}", element)
-                                } else {
-                                    format!("{}", element.ion_type())
-                                }
-                            ),
-                        ));
-                    }
-                }
-            }
-            IonSchemaElement::Document(document) => {
-                // return Violation if value is not string/symbol
-                return Err(Violation::new(
-                    "byte_length",
-                    ViolationCode::TypeMismatched,
-                    "expected a clob/blob but found document",
-                ));
-            }
-        };
+        let size = value
+            .expect_element_of_type("codepoint_length", &[IonType::String, IonType::Symbol])?
+            .as_str()
+            .unwrap()
+            .chars()
+            .count();
 
         // get isl length as a range
         let length_range: &Range = self.length();
@@ -1462,48 +1389,26 @@ impl PrecisionConstraint {
 
 impl ConstraintValidator for PrecisionConstraint {
     fn validate(&self, value: &IonSchemaElement, type_store: &TypeStore) -> ValidationResult {
-        match value {
-            IonSchemaElement::SingleElement(element) => {
-                // get the precision of given decimal
-                let value_precision = match element.as_decimal() {
-                    Some(decimal_value) => decimal_value.precision(),
-                    _ => {
-                        // return Violation if value is not decimal
-                        let error_message = if element.is_null() {
-                            format!("expected a decimal but found {:?}", element)
-                        } else {
-                            format!("expected a decimal but found {}", element.ion_type())
-                        };
-                        return Err(Violation::new(
-                            "precision",
-                            ViolationCode::TypeMismatched,
-                            &error_message,
-                        ));
-                    }
-                };
+        // get precision of decimal value
+        let value_precision = value
+            .expect_element_of_type("precision", &[IonType::Decimal])?
+            .as_decimal()
+            .unwrap()
+            .precision();
 
-                // get isl decimal precision as a range
-                let precision_range: &Range = self.precision();
+        // get isl decimal precision as a range
+        let precision_range: &Range = self.precision();
 
-                // return a Violation if the value didn't follow precision constraint
-                if !precision_range.contains(&(value_precision as i64).into()) {
-                    return Err(Violation::new(
-                        "precision",
-                        ViolationCode::InvalidLength,
-                        &format!(
-                            "expected precision {:?} found {}",
-                            precision_range, value_precision
-                        ),
-                    ));
-                }
-            }
-            IonSchemaElement::Document(document) => {
-                return Err(Violation::new(
-                    "precision",
-                    ViolationCode::TypeMismatched,
-                    "expected a decimal but found document",
-                ));
-            }
+        // return a Violation if the value didn't follow precision constraint
+        if !precision_range.contains(&(value_precision as i64).into()) {
+            return Err(Violation::new(
+                "precision",
+                ViolationCode::InvalidLength,
+                &format!(
+                    "expected precision {:?} found {}",
+                    precision_range, value_precision
+                ),
+            ));
         }
 
         Ok(())
@@ -1529,45 +1434,23 @@ impl ScaleConstraint {
 
 impl ConstraintValidator for ScaleConstraint {
     fn validate(&self, value: &IonSchemaElement, type_store: &TypeStore) -> ValidationResult {
-        match value {
-            IonSchemaElement::SingleElement(element) => {
-                // get the scale of given decimal
-                let value_scale = match element.as_decimal() {
-                    Some(decimal_value) => decimal_value.scale(),
-                    _ => {
-                        // return Violation if value is not decimal
-                        let error_message = if element.is_null() {
-                            format!("expected a decimal but found {:?}", element)
-                        } else {
-                            format!("expected a decimal but found {}", element.ion_type())
-                        };
-                        return Err(Violation::new(
-                            "scale",
-                            ViolationCode::TypeMismatched,
-                            &error_message,
-                        ));
-                    }
-                };
+        // get scale of decimal value
+        let value_scale = value
+            .expect_element_of_type("precision", &[IonType::Decimal])?
+            .as_decimal()
+            .unwrap()
+            .scale();
 
-                // get isl decimal scale as a range
-                let scale_range: &Range = self.scale();
+        // get isl decimal scale as a range
+        let scale_range: &Range = self.scale();
 
-                // return a Violation if the value didn't follow scale constraint
-                if !scale_range.contains(&(value_scale).into()) {
-                    return Err(Violation::new(
-                        "scale",
-                        ViolationCode::InvalidLength,
-                        &format!("expected scale {:?} found {}", scale_range, value_scale),
-                    ));
-                }
-            }
-            IonSchemaElement::Document(document) => {
-                return Err(Violation::new(
-                    "scale",
-                    ViolationCode::TypeMismatched,
-                    "expected a decimal but found document",
-                ));
-            }
+        // return a Violation if the value didn't follow scale constraint
+        if !scale_range.contains(&(value_scale).into()) {
+            return Err(Violation::new(
+                "scale",
+                ViolationCode::InvalidLength,
+                &format!("expected scale {:?} found {}", scale_range, value_scale),
+            ));
         }
 
         Ok(())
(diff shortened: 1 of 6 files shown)


src/types.rs
```diff
@@ -48,11 +96,23 @@ impl TypeRef {
...
         };
+
+        // convert given IonSchemaElement to an OwnedElement
+        // converts document type to SExpression
+        let value = match value.into() {
+            IonSchemaElement::Element(element) => element,
+            IonSchemaElement::Document(document) => {
+                OwnedElement::new_sexp(document.iter().cloned())
+                    .with_annotations(vec![text_token("document")])
+            }
```
popematt, line 107: issue—You are using `document::` as a meta-annotation, but it's completely valid for someone to have `document::(/*...*/)` within regular Ion data. You need some way to [escape](https://en.wikipedia.org/wiki/Escape_character) a "regular" usage of `document::(/*...*/)`. However, anything that you could use as an escape marker is also valid Ion, and would also need to be escaped. There are a few solutions to this:

1. Escape "regular" uses of `document::` with `#document::` or similar. This means that you also need to escape "regular" uses of `#document::` as `##document::` and so on.
2. Wrap all `document` _and_ `sexp` in another `sexp`. Eg. the document `a b c` becomes `document::(a b c)` and the sexp `(a b c)` becomes `sexp::((a b c))`.
3. Don't try to encode a document as a single Ion value. Document _can_ be a hard-coded type, even though it's not an Ion type.

I think option 3 will be the simplest in the long run, even if it means doing some significant refactoring right now. It's also the only option that doesn't start adding in a bunch of computational (and memory allocation) overhead for wrapping and unwrapping the documents as sexps.

src/types.rs
```diff
@@ -48,11 +96,23 @@ impl TypeRef {
...
+        };
+
         for constraint in type_def.constraints() {
-            if let Err(violation) = constraint.validate(value, &self.type_store) {
+            if let Err(violation) = constraint.validate(&value, &self.type_store) {
```
popematt, line 111: issue—as I mentioned in another comment, I think you have an escaping problem if you pass around documents as annotated s-expressions. I believe you can solve it by changing the signature of the validate functions to accept an `IonSchemaElement` instead of an `OwnedElement`.

src/types.rs
```diff
@@ -379,63 +443,63 @@ mod type_definition_tests {
...
-        TypeDefinition::named("my_int", [Constraint::type_constraint(34)])
+        TypeDefinition::named("my_int", [Constraint::type_constraint(35)])
     ),
     case::type_constraint_with_nested_self_reference_type(
         /* For a schema with nested self reference type as below:
             { name: my_int, type: { type: my_int } }
          */
         IslType::named("my_int", [IslConstraint::type_constraint(IslTypeRef::anonymous([IslConstraint::type_constraint(IslTypeRef::named("my_int"))]))]),
-        TypeDefinition::named("my_int", [Constraint::type_constraint(35)]) // 0-32 are built-in types which are preloaded to the type_store
+        TypeDefinition::named("my_int", [Constraint::type_constraint(36)]) // 0-32 are built-in types which are preloaded to the type_store
```
popematt, line 453: nitpick—Comment appears to be inaccurate now.

popematt (changes requested):
Supporting `document` is not an easy problem to solve. This is a good start, but there is a potential for subtle bugs if _user_ data actually contains any s-expression annotated with `document`. I've made some comments to explain and provide some suggestions. I spent quite a bit of time working on a nearly identical problem last year, and I'm more than happy to chat further about this if it would be of any help.

src/isl/isl_type_reference.rs
```diff
@@ -102,6 +102,11 @@ impl IslTypeRef {
                         "a base or alias type reference can not be null.struct",
                     )
                 }
+                if value.annotations().any(|a| a.text() == Some("nullable")) {
+                    return invalid_schema_error(
+                        "nullable type reference is only allowed for built in types",
+                    )
+                }
```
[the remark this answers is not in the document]
desaikd: Oh no. I just added this here because I realized the bug while working on `document` type. 
I will try to keep these type of changes separate for future PRs.

src/lib.rs
```diff
@@ -29,3 +33,57 @@ mod violation;
...
+                        .iter()
+                        .map(|oe| oe.to_owned())
+                        .collect();
+                    elements
+                }
+                _ => {
+                    panic!("invalid document")
+                }
+            };
+            // println!("{:?}", sequence);
```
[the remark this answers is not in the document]
desaikd: Oh yes. Thanks for catching that!

src/lib.rs
```diff
@@ -29,3 +33,57 @@ mod violation;
...
+    Element(OwnedElement),
+    Document(Vec<OwnedElement>),
+}
+
+impl From<&OwnedElement> for IonSchemaElement {
+    fn from(value: &OwnedElement) -> Self {
+        if value.annotations().any(|a| a.text() == Some("document")) {
+            let sequence = match value.ion_type() {
+                IonType::String => load(value.as_str().unwrap()),
+                IonType::List | IonType::SExpression => {
```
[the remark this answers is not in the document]
desaikd: I wanted to do something similar to what you mentioned here. But I am using this bit to convert an `OwnedElement` that was read through using `ion-rust`(which doesn't have a `document` type) and hence it would read it just as any other OwnedElement(an element with annotation as `document`). 
But I can still add an `impl From<&Vec<OwnedElement>> for IonSchemaElement` separately for the purpose of creating `document` for users.

src/system.rs
```diff
@@ -390,6 +390,7 @@ static DERIVED_ISL_TYPES: [&str; 9] = [
     "type::{ name: $number, one_of: [ $decimal, $float, $int ] }",
     "type::{ name: $text, one_of: [ $string, $symbol ] }",
     "type::{ name: nothing, not: $any }",
+    "type::{ name: document, annotations: required::closed::[\"document\"], type: any }",
```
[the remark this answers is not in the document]
desaikd: Hmm. That makes sense. Maybe here it should container type/string. 
```rust
     type::{ name: document, annotations: required::closed::[\"document\"], type: { any_of: [string, sexp, list] } }
```
I am not sure of a way to define this as top level value.

src/lib.rs
```diff
@@ -29,3 +33,57 @@ mod violation;
...
+/// use ion_schema::IonSchemaElement;
+///
+/// // create an IonSchemaElement from an OwnedElement
+/// let owned_element: OwnedElement = 4.into();
+/// let ion_schema_element: IonSchemaElement = (&owned_element).into();
+///
+/// // create an IonSchemaElement for document type based on vector of owned elements
+/// let document: IonSchemaElement = IonSchemaElement::Document(vec![owned_element]);
+/// ```
+pub enum IonSchemaElement {
```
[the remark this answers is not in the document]
desaikd: `Value` or `IonRustElement` ??

src/types.rs
desaikd: I think choosing option 3 like you said would need some significant refactoring including having a check for `document` type for each `validate` methods. But I do think this option is better and moves away from all these tricks or `meta-annotation` issues. I will try to refactor using option 3.

src/lib.rs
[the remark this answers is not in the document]
popematt: I would avoid including `Rust` in the name—that's implied by the fact that both libraries are Rust. I think `Value` is a little to... generic on its own, but `IonValue` could work.

src/types.rs
desaikd: I will update with correct comments.

src/lib.rs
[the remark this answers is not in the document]
popematt: Hey, what about something this?
```rust
pub enum IonData {
    // A single Ion value
    Value(OwnedElement),
    // A stream of top-level Ion values
    Document(Vec<OwnedElement>),
}
```

desaikd:
Suggested review changes:

* adds changes to use `IonSchemaElement` for validation
* changed all constraint and type validation logic to use
`IonSchemaElement` as input
* Added a Display implementation of `IonSchemaElement` to use it for
violation messages
* added changes for derived types that are preloaded into `TypeStore`
* modified tests according to above changes
* Renamed `IonSchemaElement` enum variant

src/types.rs
desaikd: I have now replaced all places where converting document to sexpression. Now `IonSchemaElement` is used everywhere in the validation logic.

src/system.rs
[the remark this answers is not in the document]
desaikd: Removed usage of converting `document` to an SExpression. `IonSchemaElement` is used everywhere in the validation logic now.

src/lib.rs
```diff
@@ -29,3 +35,118 @@ mod violation;
...
+}
+
+impl Display for IonSchemaElement {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), std::fmt::Error> {
+        match self {
+            IonSchemaElement::SingleElement(element) => {
+                write!(f, "{}", element)
+            }
+            IonSchemaElement::Document(document) => {
+                write!(f, "document::( ")?;
```
zslayton, line 108: The `document::(...)` syntax could be confusing. We could do something like:
```
/* Ion document */ ... /* end */
```
(Yuck.) Definitely open to better alternatives. This doesn't need to block the PR since it's just for error reporting.

src/system.rs
```diff
@@ -524,7 +527,7 @@ impl TypeStore {
 
     /// Provides the [Type] associated with given [TypeId] if it exists in the [TypeStore]  
     /// Otherwise returns None
-    pub(crate) fn get_type_by_id(&self, id: TypeId) -> Option<&TypeDefinition> {
+    pub fn get_type_by_id(&self, id: TypeId) -> Option<&TypeDefinition> {
```
zslayton, line 530: Why is this now fully public?

src/types.rs
```diff
@@ -166,34 +219,67 @@ impl TypeDefinition {
...
+                            ));
+                        }
                     }
-                    Ok(())
                 }
                 BuiltInTypeDefinition::Derived(other_type) => {
+                    if other_type.name() == &Some("document".to_owned()) {
+                        // Verify whether the given derived type is document
+                        // And check if it is using enum variant IonSchemaElement::Document
+                        if !matches!(value, IonSchemaElement::Document(_)) {
```
zslayton, line 270: Might be nice to have `is_document()` and `is_element()` convenience methods on the `IonSchemaElement` type.

src/types.rs
```diff
@@ -288,23 +374,27 @@ impl TypeDefinitionImpl {
...
             // clear parent information from type_store as the type is already added in the type_store now
             pending_types.clear_parent();
         } else {
             // update with this resolved type_def to context for type_id
-            pending_types.update_anonymous_type(type_id, type_def.to_owned(), type_store);
+            actual_type_id = pending_types.update_anonymous_type(type_id, type_def, type_store);
         }
 
-        Ok(type_def)
+        // let type_ref = TypeRef::new(final_type_id, Rc::from(type_store.to_owned()));
```
zslayton, line 396:
```suggestion
```

src/types.rs
desaikd: I already had `as_element()` defined as:
```rust
pub fn as_element(&self) -> Option<&OwnedElement> {
        match self {
            IonSchemaElement::SingleElement(element) => Some(element),
            IonSchemaElement::Document(_) => None,
        }
    }
```
I have similarly created `as_docuemnt()` and used it for type check as below:
```rust
if value.as_document() == None {
     // error if not a document
}
```

src/lib.rs
desaikd: Yes, I wad confused about this as this one is in particular an `SExpression` with `document` annotation. I can do this comment thing as I think that wouldn't conflict with any definition.

zslayton (approved):
👍

As discussed offline, there's room for even more elimination of boilerplate but we can take care of that in follow on PRs.

src/constraint.rs
```diff
@@ -1138,48 +1123,64 @@ impl ElementConstraint {
...
         // this type_id was validated while creating `ElementConstraint` hence the unwrap here is safe
         let type_def = type_store.get_type_by_id(self.type_id).unwrap();
 
-        // validate each element of the given value container
-        match value.ion_type() {
-            IonType::List | IonType::SExpression => {
-                for val in value.as_sequence().unwrap().iter() {
-                    if let Err(violation) = type_def.validate(val, type_store) {
-                        violations.push(violation);
+        // get the size of given string/symbol Unicode codepoints
```
popematt, line 1132: Looks like copy/paste error for this comment.

src/constraint.rs
```diff
@@ -1344,15 +1345,28 @@ impl AnnotationsConstraint {
...
+
+                // validate annotations that does not have list-level `ordered` annotation
+                self.validate_unordered_annotations(element, type_store, violations)
+            }
+            IonSchemaElement::Document(document) => {
+                // document type can not have annotations
+                Err(Violation::new(
+                    "annotations",
+                    ViolationCode::AnnotationMismatched,
+                    "annotations don't match expectations",
```
popematt, line 1366: question—Should this be a "not applicable for this type" sort of violation instead? (Would that be clearer, do you think?)

src/constraint.rs
```diff
@@ -1550,35 +1530,47 @@ impl ValidValuesConstraint {
...
+                    &format!(
+                        "expected valid_values to be from {:?}, found {:?}",
+                        &self, value
+                    ),
+                ))
+            }
+            IonSchemaElement::Document(document) => Err(Violation::new(
+                "valid_values",
+                ViolationCode::InvalidValue,
+                "valid_values constraint is not allowed to be used for document",
```
popematt, line 1571: minor suggestion—rather than having a different message (and having people wonder "why can't it be used for a document"), you could use the same message. `valid_values` will never match a `document` because there's no way to create a `valid_values` constraint with a `document` in it.

src/types.rs
```diff
@@ -379,144 +468,144 @@ mod type_definition_tests {
             { name: my_int, type: my_int }
          */
         IslType::named("my_int", [IslConstraint::type_constraint(IslTypeRef::named("my_int"))]),
-        TypeDefinition::named("my_int", [Constraint::type_constraint(34)])
+        TypeDefinition::named("my_int", [Constraint::type_constraint(35)])
```
popematt, line 471: We need to figure out a better solution for these tests than using the typeId numbers. It's not maintainable.

I don't expect it to be fixed in this PR, but it needs to be done soon. The more difficult it is to write/maintain tests, the less motivated someone will be to write/maintain tests.

src/lib.rs
```diff
@@ -29,3 +35,125 @@ mod violation;
...
+        match self {
+            IonSchemaElement::SingleElement(_) => None,
+            IonSchemaElement::Document(document) => Some(document),
+        }
+    }
+
+    fn expect_element_of_type(
+        &self,
+        constraint_name: &str,
+        types: &[IonType],
```
popematt, line 80: minor suggestion—it might make the code _slightly_ more readable to swap the order of these arguments. Consider the following:

```rust
expect_element_of_type("fields", &[IonType::Struct])?
```
vs.
```rust
expect_element_of_type(&[IonType::Struct], "fields")?
```

In my mind, the first one reads "expect element of type (fields) Struct", whereas the second one reads "expect element of type Struct (fields)".

src/lib.rs
```diff
@@ -29,3 +35,125 @@ mod violation;
...
+            IonSchemaElement::Document(_) => None,
+        }
+    }
+
+    pub fn as_document(&self) -> Option<&Vec<OwnedElement>> {
+        match self {
+            IonSchemaElement::SingleElement(_) => None,
+            IonSchemaElement::Document(document) => Some(document),
+        }
+    }
```
popematt, line 75: question—is there any particular reason for creating these functions rather than using `From`/`Into`?

src/lib.rs
```diff
@@ -29,3 +35,125 @@ mod violation;
...
+                write!(f, "/* Ion document */ ")?;
+                let mut peekable_itr = document.iter().peekable();
+                while peekable_itr.peek() != None {
+                    let value = peekable_itr.next().unwrap();
+                    write!(f, "{}", value)?;
+                    if peekable_itr.peek() != None {
+                        write!(f, " ")?;
+                    }
+                }
+                write!(f, " /* end */")
```
popematt, line 124: minor suggestion—you can eliminate the need for peeking by dropping the space before the "/* end */" comment, which will allow you to simplify the code to something like this:
```rust
write!(f, "/* Ion document */ ")?;
for value in document {
    write!(f, "{} ", value)?;
}
write!(f, "/* end */")
```
(You should double check if you choose to do this—I'm not sure off the top of my head whether it should be `document` or `&document`.)

src/types.rs
```diff
@@ -14,11 +15,11 @@ use std::rc::Rc;
 pub trait TypeValidator {
     /// If the specified value violates one or more of this type's constraints,
     /// returns `false`, otherwise `true`
-    fn is_valid(&self, value: &OwnedElement, type_store: &TypeStore) -> bool;
+    fn is_valid(&self, value: &IonSchemaElement, type_store: &TypeStore) -> bool;
```
popematt, line 18: question—Do you think it's worth exposing `IonSchemaElement` publicly here? Could we have two different functions (like this example), both of which convert to `IonSchemaElement` behind the scenes?
```rust
fn is_valid(&self, value: &OwnedElement, type_store: &TypeStore) -> bool;
fn is_valid(&self, value: &Vec<OwnedElement>, type_store: &TypeStore) -> bool;
```

Which option do you think is nicer for someone starting to use this library for the first time?

popematt:
Overall, this approach looks really good. Just a few minor questions/comments.

src/constraint.rs
desaikd: Yes, I think that is much precise. I will change it.

src/types.rs
desaikd: I will prioritize this issue.

src/types.rs
desaikd: This is an internal method and would not be exposed to user. The method that will exposed to user is:
```rust
     pub fn validate<I: Into<IonSchemaElement>>(&self, value: I) -> ValidationResult {
        ...
     }
```

This means that people can just pass on `&OwnedElement` and `&Vec<OwnedElement>` to this validate method and method takes care of the conversion to `IonSchemaElement`.

Currently it didn't have an implementation of `From<&Vec<OwnedElement>> for IonSchemaElement` but I will add it in following commit along with your suggested changes..

popematt: 🤦‍♂️ Right, we talked about this already, and I just forgot about it.

src/constraint.rs
desaikd: Are you saying that the violation message should just say:

> `valid_values` will never match a document because there's no way to create a `valid_values` constraint with a document in it.

popematt: I'm suggesting it should be:
```rust
&format!("expected valid_values to be from {:?}, found {:?}", &self, value)
```

desaikd: Oh okay, got it.

popematt (approved):

PR #116, opened 2022-09-24
adds changes to allow validating built-in types for TypeRef validation

desaikd:
*Description of changes:*
This PR works on adding changes for allowing validation for built-in types.

*Example of the issue:*
```rust
let schema = load_schema_from_text(r#" // No schema defined, uses built-in types"#);
// The following was only checking for types defined within schema but not built-in types
let type_to_validate = schema.get_type("int"); 
type_to_validate.validate(load_value("5"));
```

*List of changes:*
* adds implementation for validation of `BuilInTypeDefinition`
* removes redundant `TypeDefinition` checks for validation
* adds method in `TypeStore` to retrieve either a built-int type id or a defined type id based on the given type name

*Test:*
* adds test for built-in type validation

### DIFF
--- a/src/system.rs
+++ b/src/system.rs
@@ -517,7 +517,6 @@ impl TypeStore {
         };
         self.ids_by_name
             .get(name)
-            .or_else(|| self.imported_type_ids_by_name.get(name))
             .or_else(|| self.builtin_type_ids_by_name.get(type_name))
     }
 
--- a/src/system.rs
+++ b/src/system.rs
@@ -33,6 +33,7 @@ use ion_rs::value::owned::{text_token, OwnedElement, OwnedSymbolToken};
 use ion_rs::value::reader::{element_reader, ElementReader};
 use ion_rs::value::{Element, Sequence, Struct};
 use ion_rs::IonType;
+use regex::Regex;
 use std::collections::HashMap;
 use std::io::ErrorKind;
 use std::rc::Rc;
@@ -679,8 +680,24 @@ impl Resolver {
         let mut found_header = false;
         let mut found_footer = false;
 
+        // ISL version marker regex
+        let isl_version_marker = Regex::new(r"^\$ion_schema_\d.*$").unwrap();
+
         for value in elements {
+            // verify if value is an ISL version marker and if it has valid format
+            if value.ion_type() == IonType::Symbol
+                && isl_version_marker.is_match(value.as_str().unwrap())
+            {
+                // This implementation only supports Ion Schema 1.0
+                if value.as_str() != Some(r"$ion_schema_1_0") {
+                    return invalid_schema_error(
+                        "Unsupported Ion Schema version: ${it.stringValue()}",
+                    );
+                }
+            }
+
             let annotations: Vec<&OwnedSymbolToken> = value.annotations().collect();
+
             // load header for schema
             if annotations.contains(&&text_token("schema_header")) {
                 found_header = true;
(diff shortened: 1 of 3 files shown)


src/system.rs
```diff
@@ -502,8 +502,27 @@ impl TypeStore {
...
+        let type_name = match name {
+            "int" => "integer",
+            "bool" => "boolean",
+            "$int" => "$integer",
+            "$bool" => "$boolean",
+            _ => name,
+        };
+        self.ids_by_name
+            .get(name)
+            .or_else(|| self.imported_type_ids_by_name.get(name))
```
popematt, line 520: Suppose I have `foo.isl`:
```
type::{ name: foo_type }
```
and `bar.isl`:
```
schema_header::{ imports: [{ id: "foo.isl" }] }

type::{ name: bar_type }

schema_footer::{}
```
Assuming I have an authority that can resolve the import, what happens if I do this?
```rs
let schema = schema_system.load_schema("bar.isl");
let type_to_validate = schema.get_type("foo_type"); 
```
The result _should_ be that `type_to_validate` is `None`, but looking at this line, I _think_ `type_to_validate` will be (pseudocode) `Some(foo_type)`.

desaikd: Do you mean to say that imported type can not be used for validation? So whenever we try to `get_type` for an imported type it should result `None`?

I believe it should return the imported type when `get_type` is used.

popematt: Well, I was going to refer you to https://github.com/amzn/ion-schema/issues/39, but then I realized that using `get_type` in this way is not actually importing it, so I was incorrect to say that it should (as if the spec required it) be `None`.

However, I do still believe that it should be `None`. I think that if a user wants to use `foo_type`, they should be loading `foo.isl` rather than `bar.isl`. It comes back to naming conflicts. Suppose you have schema A which imports schema B and C. (These schemas are not necessarily owned by the same people or stored in the same location.) Schema B has a `d_type`, which you are using by loading A. Then Schema C introduces a different `d_type`. Now Schema A won't load anymore because of a name conflict. This problem isn't specific to the change that you're proposing, but I believe it illustrates why I prefer to be explicit and direct with dependencies, so I'd prefer not to enable people to create an indirect dependency like that. This is why I think that an imported type should not be returned by `get_type`.

Why do you think that the imported type _should_ be returned by `get_type`?

desaikd: Oh it does make sense if the imported type is coming from a schema that is not owned by the same people using the imported type. 

I didn't consider this case that you explained above. I thought what if someone wants that type which was actually part of the given schema as it was imported and used in this schema. But it does seem indirect dependency like this should be avoided.

Also, `ion-schema-kotlin` says its `getType` also includes imported types. So either way I feel both implementations should be consistent. 

I can go ahead and change it for `ion-schema-rust` and open an issue in `ion-schema-kotlin` for this. 
Although I am not sure if someone would actually be depending on this use case of `getType` from `ion-schema-kotlin`, we might need to verify this before changing it there.

popematt: > we might need to verify this before changing it there.

You are correct that it could be a backwards incompatible change, but there isn't any way to verify if someone is using it, which is why the fixed behavior was added as an optional configuration in  https://github.com/amzn/ion-schema-kotlin/pull/180.  Whenever we release a new major version of `ion-schema-kotlin`, then the correct behavior will be the default.

desaikd: Understood. Thanks!

popematt (approved):

Issue #114, opened 2022-09-15
Issue with loading type definition that uses type references from the following part of schema


desaikd:
There is a bug in loading schema that has a definition which uses a type reference from the following part of the schema.

Example: 
```
type:: {
  name: a,
  type: b // <--- this is defined in the following part of schema
},
type::{
  name: b,
  type: int
}
```

The above schema throws following error:
```
UnresolvableSchemaError { description: "Could not resolve type reference: \"b\" does not exist" }
``` 
This is because it doesn't keep track of following part of schema where other type definitions are defined.

PR #118, opened 2022-10-04
Fixes loading type definition that uses type references from the following part of schema

desaikd:
*Issue #114*

*Description of changes:*
This PR works on resolving issue when loading a type definition that uses a type reference from following part of schema.

*Example of the issue:*
```
type:: {
  name: a,
  type: b // <--- this is defined in the following part of schema
},
type::{
  name: b,
  type: int
}
```

The above schema throws following error:
```
UnresolvableSchemaError { description: "Could not resolve type reference: \"b\" does not exist" }
```

*Solution:*
As per above example whenever we discover an unknown type reference like `b` (which might be defined in following part of schema), 
1. add that type definition using the `TypdefinitionImpl#new_deferred_type_def()` to the `PendingTypes`. 
2. when updating `TypeStore` with the `PendingTypes`, verify if that deferred type definition actually exists in the schema,
  a. if type reference does not exist in the schema, return an error for this undefined type reference.
  b. if type reference exists in the schema, then whenever we resolve this type reference update its type definition in the `TypeStore`.

*List of changes:*
* `isl_type_reference` module
  * adds changes for adding a deferred type definition whenever an unknown type reference is found while loading type definition.
  * renames `get_type_id_from_built_in_type_name` correctly to `get_type_id_from_type_name` as per its functionality
* `system`  module
  * adds changes in `PendingTypes#update_type_store()` for correctly while verifying if the deferred type definitions actually exists in the schema
  * removes unused methods from `PendingTypes`
  * adds `PendingTypes#add_deferred_type_with_name()` to add deferred type definition for a type reference that is not yet resolved
  * adds `PendingTypes#update_deferred_type_def` to update deferred type definition with its resolved type definition
  * all `Resolver` methods now use `isl_type_names` which keeps names of all the types defined within given schema (this is created from first phase of loading schema using `IslSchema#types()` )
* `types` module
  * adds changes in `TypdefinitionImpl` to store a deferred type (which will resolved later while loading schema)

*Test*
* adds unit tests to fix the issue with loading a type reference that is defined in the following part of schema.

### DIFF
--- a/src/system.rs
+++ b/src/system.rs
@@ -34,7 +34,7 @@ use ion_rs::value::reader::{element_reader, ElementReader};
 use ion_rs::value::{Element, Sequence, Struct};
 use ion_rs::IonType;
 use regex::Regex;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::io::ErrorKind;
 use std::rc::Rc;
 
@@ -46,6 +46,8 @@ use std::rc::Rc;
 /// * A reference to itself. This could happen in a recursive structure like a
 ///   linked list or binary tree.
 /// * A nested anonymous type definition.
+/// * A reference to type definition that is followed by current type definition. These type references
+///   will be deferred to later check if a type definition with that name exists in the schema.
 /// Because the [`SchemaSystem`] does not yet know the complete definition
 /// of these types, it cannot find them in the [`TypeStore`].
 /// An instance of [`PendingTypes`] is used to track information about types
@@ -68,18 +70,25 @@ impl PendingTypes {
     /// * `load_isl_import` - If this argument is Some(isl_import), then we are not within an import process of schema.
     ///                       Based on given enum variant isl_import we will add the types to type_store.
     ///                       Otherwise we will add all the types from this PendingTypes to TypeStore.
+    /// * `isl_type_names` - The isl type names defined within the schema. This will be used to determine
+    ///                      if a type definition actually exists within the schema.
     /// Returns true, if this update is not for an isl import type or it is for an isl import type but it is added to the type_store
     /// Otherwise, returns false if this update is for an isl import type and it is not yet added to the type_store.
     pub fn update_type_store(
         &mut self,
         type_store: &mut TypeStore,
         load_isl_import: Option<&IslImport>,
+        isl_type_names: &HashSet<&str>,
     ) -> IonSchemaResult<bool> {
         // if load_isl_import is not None, then match the enum variant and update type store with import types accordingly
         if let Some(import) = load_isl_import {
             match import {
                 IslImport::Schema(_) => {
-                    self.update_type_store_with_all_isl_imported_types(None, type_store)?;
+                    self.update_type_store_with_all_isl_imported_types(
+                        None,
+                        type_store,
+                        isl_type_names,
+                    )?;
                 }
                 IslImport::Type(isl_import) => {
                     // if import has a specified type to import then only add that type
@@ -91,6 +100,7 @@ impl PendingTypes {
                         self.update_type_store_with_all_isl_imported_types(
                             Some(isl_import.type_name()),
                             type_store,
+                            isl_type_names,
                         )?;
                     } else {
                         // if the named_type_def appears as None then it means we haven't reached
@@ -117,6 +127,7 @@ impl PendingTypes {
                         self.update_type_store_with_all_isl_imported_types(
                             Some(isl_import.type_name()),
                             type_store,
+                            isl_type_names,
                         )?;
                     } else {
                         // if the named_type_def appears as None then it means we haven't reached
@@ -128,7 +139,7 @@ impl PendingTypes {
             }
         } else {
             // if load_isl_import is None i.e. it is the root schema, then update type_store with all the types inside this PendingTypes
-            self.update_type_store_with_all_types(type_store)?;
+            self.update_type_store_with_all_types(type_store, isl_type_names)?;
         }
         self.types_by_id.clear();
         self.ids_by_name.clear();
@@ -182,7 +193,11 @@ impl PendingTypes {
     }
 
     // helper method to update type store with all the types from this PendingTypes
-    fn update_type_store_with_all_types(&self, type_store: &mut TypeStore) -> IonSchemaResult<()> {
+    fn update_type_store_with_all_types(
+        &self,
+        type_store: &mut TypeStore,
+        isl_type_names: &HashSet<&str>,
+    ) -> IonSchemaResult<()> {
         for optional_type in &self.types_by_id {
             // return an error if any of the type in types_by_id vector is None/Unresolved
             let type_def = optional_type.to_owned().ok_or_else(|| {
@@ -190,7 +205,19 @@ impl PendingTypes {
             })?;
 
             match type_def {
-                TypeDefinition::Named(named_type_def) => type_store.add_named_type(named_type_def),
+                TypeDefinition::Named(named_type_def) => {
+                    // check if the type definitions that are not yet resolved actually exists within the schema
+                    // we can use the isl_type_names to make sure if they exists, otherwise return error.
+                    if named_type_def.is_deferred_type_def()
+                        && !isl_type_names
+                            .contains(named_type_def.name().as_ref().unwrap().as_str())
+                    {
+                        return unresolvable_schema_error(
+                            "Unable to load schema due to unresolvable type",
+                        );
+                    }
+                    type_store.add_named_type(named_type_def)
+                }
                 TypeDefinition::Anonymous(anonymous_type_def) => {
                     type_store.add_anonymous_type(anonymous_type_def)
                 }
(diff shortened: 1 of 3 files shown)


src/system.rs
```diff
@@ -182,15 +193,31 @@ impl PendingTypes {
...
-                TypeDefinition::Named(named_type_def) => type_store.add_named_type(named_type_def),
+                TypeDefinition::Named(named_type_def) => {
+                    // check if the type definitions that are not yet resolved actually exists within the schema
+                    // we can use the isl_type_names to make sure if they exists, otherwise return error.
+                    if named_type_def.is_deferred_type_def()
+                        && !isl_type_names
+                            .contains(named_type_def.name().as_ref().unwrap().as_str())
+                    {
+                        return unresolvable_schema_error(
+                            "Unable to load schema due to unresolvable type",
```
popematt, line 216: suggestion–in the error message, include the name of the type that cannot be resolved

src/system.rs
```diff
@@ -220,10 +248,27 @@ impl PendingTypes {
...
                     match isl_imported_type_name {
                         None => {
+                            // check if the type definitions that are not yet resolved actually exists within the schema
+                            // we can use the isl_type_names to make sure if they exists, otherwise return error.
+                            if named_type_def.is_deferred_type_def()
+                                && !isl_type_names
+                                    .contains(named_type_def.name().as_ref().unwrap().as_str())
+                            {
+                                return unresolvable_schema_error(
+                                    "Unable to load schema due to unresolvable type",
```
popematt, line 258: suggestion–in the error message, include the name of the type (and _possibly_ the ID of the schema that we're attempting to import from) that cannot be resolved

src/system.rs
```diff
@@ -220,10 +248,27 @@ impl PendingTypes {
...
+                            }
                             // imports all types into imported_type_ids_by_name section of type_store
                             type_store.add_isl_imported_type(None, named_type_def);
                         }
                         Some(import_type_name) => {
+                            // if there are any deferred type definitions for this import
+                            // then return an error
+                            if named_type_def.is_deferred_type_def() {
+                                return unresolvable_schema_error(
+                                    "Type definition being imported has dependency to a type reference which is not being imported from import schema",
```
popematt, line 269: I don't understand what this message means. Can you explain to me how would an `ion-schema-rust` user cause this error to occur?

src/system.rs
```diff
@@ -1357,6 +1426,93 @@ mod schema_system_tests {
...
+                "sample_builtin_nullable_types.isl",
+                r#"
+                    schema_header::{
+                      imports: [],
+                    }
+                    
+                    type::{
+                      name: my_text,
+                      one_of: [
+                        my_string, // this type reference was not improted in sample.isl
```
popematt, line 1459:
```suggestion
                        my_string, // this type reference was not imported in sample.isl
```

src/system.rs
```diff
@@ -1357,6 +1426,93 @@ mod schema_system_tests {
         assert!(schema.is_ok());
     }
 
+    #[test]
+    fn schema_system_map_authority_with_invalid_inline_import_type_test() {
```
popematt, line 1430: Can you explain to me how this is invalid? There are no inline imports here. All of the imports that do exist appear to be perfectly fine.

src/system.rs
```diff
@@ -1357,6 +1426,93 @@ mod schema_system_tests {
...
+                    }
+                "#,
+        )];
+        let mut schema_system =
+            SchemaSystem::new(vec![Box::new(MapDocumentAuthority::new(map_authority))]);
+        // verify if the schema loads without any errors
+        let schema = schema_system.load_schema("sample.isl");
+        assert!(schema.is_ok());
+    }
+
```
popematt, line 1515: Here are two more tests that I think cover important use cases for recursively defined data structures. I know that they are a little bit contrived (why would anyone have data that can only be alternately-nested lists and sexps?), but they are representative of recursive data types. The alternating recursive definition is a lot like ISL types and constraints, and the self-referencing recursion is basically a tree or heap.

```suggestion

    #[test]
    fn schema_system_map_authority_with_multiple_codependent_type_definitions() {
        // map with (id, ion content)
        let map_authority = [(
            "sample.isl",
            r#"
                    type::{
                       name: node_a,
                       type: list,
                       element: node_b,
                    }
                    
                    type::{
                       name: node_b,
                       type: sexp,
                       element: node_a,
                    }
                "#,
        )];
        let mut schema_system =
            SchemaSystem::new(vec![Box::new(MapDocumentAuthority::new(map_authority))]);
        // verify if the schema loads without any errors
        let schema = schema_system.load_schema("sample.isl");
        assert!(schema.is_ok());
    }

    #[test]
    fn schema_system_map_authority_with_self_referencing_type_definition() {
        // map with (id, ion content)
        let map_authority = [(
            "sample.isl",
            r#"
                    type::{
                       name: binary_heap_node,
                       type: sexp,
                       element: { one_of: [ binary_heap_node, int ] },
                       container_length: range::[0, 2],
                    }
                "#,
        )];
        let mut schema_system =
            SchemaSystem::new(vec![Box::new(MapDocumentAuthority::new(map_authority))]);
        // verify if the schema loads without any errors
        let schema = schema_system.load_schema("sample.isl");
        assert!(schema.is_ok());
    }

```

popematt (changes requested):
Can you add some documentation that explains the difference between a pending type and a deferred type? When will something be pending and not deferred vs pending and deferred? Is it possible for something to be deferred and not pending?

src/system.rs
desaikd: I meant say that if ta type definition that is being imported has a type reference that is not imported then that should return an error. 
Example: 
```rust
             (
                "sample.isl",
                r#"
                    schema_header::{
                      imports: [ { id: "sample_builtin_nullable_types.isl", type: my_text } ],
                    }
                    
                    type::{
                      name: my_type,
                      type: my_text
                    }
                    
                    schema_footer::{
                    }
                "#,
            ),
            (
                "sample_builtin_nullable_types.isl",
                r#"
                    schema_header::{
                      imports: [],
                    }
                    
                    type::{
                      name: my_text,
                      one_of: [
                        my_string, // <--- this type reference was not improted in sample.isl
                        symbol,
                      ],
                    }
                    
                    type::{
                        name: my_string,
                        type: string,
                    }
                    
                    schema_footer::{
                    }
                "#,
            ),
 ```

File: src/system.rs (hunk 5 above)
Comment (desaikd) [reply src/system.rs:1430]: My bad. Test name should be changed to `schema_system_map_authority_with_invalid_type_import_test`.
The reason for this is we are importing `my_text` which has a type reference `my_string`. Here only `my_text` is being imported. 
This is related to an error I explained above.

**Comment by desaikd** [comment]
> Can you add some documentation that explains the difference between a pending type and a deferred type? When will something be pending and not deferred vs pending and deferred? 
Yes, I will add some documentation for that. Explaining here as well. There is no such thing is pending type, it is just a name for a single type definition which is currently being parsed. This uses `PendingTypes` because an ISL type definition can include types that are not yet fully defined (self-reference, nested anonymous types or deferred type). Deferred type is basically a type reference that is not resolved because the type definition for it has not yet been parsed. 
e.g.
```
type::{
  name: a,
  type: b
}
type::{
  name: b,
  type: int
}
```
For above, when parsing type definition `a`, type reference `b` would not have yet resolved as it is defined following this type definition. Hence while a is being parsed `b` will be called deferred type. Once we parse type reference `b` its deferred type definition will be updated.

Both deferred type and pending type(current type) are defined using `TypeDefinitionImpl`.
>Is it possible for something to be deferred and not pending? 
Pending type is just a name for a type currently being resolved so it is possible that a type is deferred but not pending. but it will be in a pending type soon when we will parse it.  
for example, when parsing type definition `a` type reference `b` will  be added to `TypeStore` as deferred type. But when we will parse `b` it will be a pending type or a resolved type and will replace its deferred type definition.

> Also, Clippy has a couple of warnings. Can you address those?
Sure, I see only one warning  which I think got added because of my change after running `cargo clippy`. I will correct that.

src/system.rs
desaikd: I will add these tests with next commit

src/system.rs
desaikd: I will add type definition name for the type import and transitive type def in this error message.

popematt: Ah, but this example you provided should be valid. The `sample.isl` schema should not have to import `my_string` because `my_string` is abstracted behind `my_text`. If ISL required you to import even types that you were using indirectly, it would quickly become unmaintainable to logically organize types across multiple files.

src/system.rs
popematt: Okay, so `sample.isl` should not be able to use `my_string` by name, but it should be able to use `my_text` without importing all of `my_text`'s dependencies. The "transitivity" that we're concern about actually has to do with whether a type can be referenced directly within the scope of a particular schema document. (Essentially, it's about each schema having a namespace for type names instead of having a single, global namespace.)

popematt:
Okay, let me make sure I understand. Would a good TLDR about pending vs deferred be something like this?
* "pending" is a state related to **parsing** a _type_
* "deferred" is a state related to **resolving** a _type reference_

src/system.rs
desaikd: Thanks for the explanation! Makes sense. I will change accordingly and add both valid and invalid tests to make sure the behavior is correct.

popematt (approved):

PR #124, opened 2022-11-14
Adds an error check for top level type definition's `name` field

desaikd:
*Description of changes:*
This PR adds an error check for a top level type definition `name` field.

*List of changes:*
- modifies `IslSchema#types()` to return `IslType` instead of `IslTypeImpl`
- adds `name()` method for `IslType`
- adds an error check for top level type def `name` field
- adds a unit test for this error check

*Test:*
Adds a unit test for the error check

### DIFF
--- a/src/system.rs
+++ b/src/system.rs
@@ -720,14 +720,7 @@ impl Resolver {
             isl_types
                 .iter()
                 .filter(|t| matches!(t, IslType::Named(_)))
-                .map(|t| match t {
-                    IslType::Named(named_isl_type) => {
-                        named_isl_type.name().as_ref().unwrap().as_str()
-                    }
-                    IslType::Anonymous(_) => {
-                        unreachable!("already filtered named types")
-                    }
-                }),
+                .map(|t| t.name().as_ref().unwrap().as_str()),
         );
 
         for isl_type in &isl_types {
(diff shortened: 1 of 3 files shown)


src/system.rs
```diff
@@ -1640,6 +1648,24 @@ mod schema_system_tests {
         assert!(schema.is_ok());
     }
 
+    #[test]
+    fn top_level_type_def_without_name_field() {
```
popematt, line 1652: question—what happens if more than one `name` field is present, or if the `name` field is not a symbol?

popematt (approved):

src/system.rs
desaikd: If more than one `name` field are supplied it would select the last one. If the `name` field is not a `string` or `symbol` it will return an error.

popematt: >  If more than one name field are supplied it would select the last one.

Structs are unordered, so this is undefined behavior. It needs to be an error if more than one name field exists.

> If the name field is not a string or symbol it will return an error.

I'm pretty sure the spec says that type names may only symbols, so it should also be an error if the type name is a string.

desaikd: Reference PR: #125

PR #127, opened 2022-12-02
Convert project to Cargo workspace

popematt:
**Issue #, if available:**

None

**Description of changes:**

This change moves the main library into `ion-schema/` directory, and makes the root of the repo into a cargo workspace. Most files are moved without any modifications, but there are a few that had to be modified because of file paths that are hard-coded in the tests.

I believe this refactor will make it easier to maintain in the long term since there is a single `Cargo.lock` file for the whole workspace as opposed to one for each crate. It will also make it easier if we ever decide we want to separate Ion Schema models and read/write logic from the validation logic, or add sub-projects/crates for any other reason.

_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice._


ion-schema/tests/ion_schema_tests_runner.rs
```diff
@@ -88,14 +88,24 @@ const SKIP_LIST: &[&str] = &[
...
-fn validation_tests(path: &str) {
-    print!("{}...", path);
+fn validation_tests(path_string: &str) {
+    // The #[test_resources] macro provides file paths that are relative to the cargo workspace root
+    // but when this test is run, it seems the current directory is the "ion-schema/" directory, so
+    // we need to adjust the path to ensure that they are relative to the correct location.
+    let mut path_buf = PathBuf::from_str(TEST_ROOT_DIR).unwrap();
+    if path_string.starts_with("ion-schema-tests") {
+        path_buf.pop();
+    }
```
zslayton, line 98: The `if` here suggests that `path_string` will sometimes _not_ start with `ion-schema-tests`. Is that the case, or can we simply call `path_buf.pop()` no matter what? The working directory of a call to `cargo test` is wherever the `Cargo.toml` file for that crate lives, so I think it should always be the same unless we reorganize the directory structure again.

zslayton (approved):
One superficial comment to consider below.

PR #130, opened 2022-12-06
adds changes for `utf8_byte_length` constraint

desaikd:
*Issue #9 #10*

*Description of changes:*
This PR works on adding implementation of `utf8_byte_length` constraint.

*Grammar:*
```ANTLR
utf8_byte_length: <INT>
utf8_byte_length: <RANGE<INT>>
```

*Ion Schema specification:*
https://amzn.github.io/ion-schema/docs/isl-1-0/spec#utf8_byte_length

*List of changes:*
* adds changes for `utf8_byte_length` constraint implementation
* adds `Utf8ByteLength` enum variants for `IslConstraint` and `Constraint`
* adds `Utf8ByteLengthConstraint` implementations
* adds validation logic for `utf8
_byte_length`
* adds unit tests for `utf8_byte_length`

*Tests:*
added unit tests for `utf8_byte_length` implementation.
- adds tests for programmatically creating `utf8_byte_length` constraint
- adds tests for loading a schema using `utf8_byte_length` constraint 
- adds validation tests for `utf8_byte_length` constraint

### DIFF
--- a/src/constraint.rs
+++ b/src/constraint.rs
@@ -687,7 +687,7 @@ impl OrderedElementsConstraint {
 
         // greedily take as many values as we can of this type without going out
         // of the maximum of the range
-        while values_iter.peek() != None && occurs_range.contains(&(count + 1).into()) {
+        while Option::is_some(&values_iter.peek()) && occurs_range.contains(&(count + 1).into()) {
             // don't consume it until we know it's valid for the type
             if let Some(value) = values_iter.peek() {
                 let schema_element: IonSchemaElement = (*value).into();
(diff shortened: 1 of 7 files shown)


popematt (approved):
The `utf8_byte_length` changes look good.

I think clippy was suggesting that you replace `foo != None` with `foo.is_some()` rather than `Option::is_some(&foo)`. I don't think it makes much difference though.

desaikd:
> I think clippy was suggesting that you replace `foo != None` with `foo.is_some()` rather than `Option::is_some(&foo)`. I don't think it makes much difference though.

Yes, I have already changed according to clippy warnings in https://github.com/amzn/ion-schema-rust/pull/130/commits/02285169a1c986f3acbf0198b4dec90f4667ae4b so it seems to be good now.

PR #131, opened 2022-12-14
adds changes for `timestamp_offset` constraint

desaikd:
*Issue #9 #10*

*Description of changes:*
This PR works on adding implementation of `timestamp_offset` constraint.

*Grammar:*
```ANTLR
timestamp_offset: [ "[+|-]hh:mm"... ]
```

*Ion Schema specification:*
https://amzn.github.io/ion-schema/docs/isl-1-0/spec#timestamp_offset

*List of changes:*
* adds changes for `timestamp_offset` constraint implementation
* adds a util implementation of `TimestampOffset` to create `timestamp_offset` constraint with valid  timetsamp offset values
* adds `TimestampOffsetLength` enum variants for `IslConstraint` and `Constraint`
* adds `TimestampOffsetConstraint` implementations
* adds validation logic for `timestamp_offset`
* adds unit tests for `timestamp_offset`

*Tests:*
added unit tests for `timestamp_offset` implementation.
- adds tests for programmatically creating `timestamp_offset` constraint
- adds tests for loading a schema using `timestamp_offset` constraint 
- adds validation tests for `timestamp_offset` constraint

### DIFF
--- a/ion-schema/src/constraint.rs
+++ b/ion-schema/src/constraint.rs
@@ -1,7 +1,7 @@
 use crate::isl::isl_constraint::{IslConstraint, IslRegexConstraint};
 use crate::isl::isl_range::{Range, RangeImpl};
 use crate::isl::isl_type_reference::IslTypeRef;
-use crate::isl::util::{Annotation, TimestampPrecision, ValidValue};
+use crate::isl::util::{Annotation, TimestampOffset, TimestampPrecision, ValidValue};
 use crate::result::{
     invalid_schema_error, invalid_schema_error_raw, IonSchemaError, IonSchemaResult,
     ValidationResult,
@@ -50,6 +50,7 @@ pub enum Constraint {
     Precision(PrecisionConstraint),
     Regex(RegexConstraint),
     Scale(ScaleConstraint),
+    TimestampOffset(TimestampOffsetConstraint),
     TimestampPrecision(TimestampPrecisionConstraint),
     Type(TypeConstraint),
     Utf8ByteLength(Utf8ByteLengthConstraint),
@@ -158,6 +159,11 @@ impl Constraint {
         Constraint::TimestampPrecision(TimestampPrecisionConstraint::new(
             Range::TimestampPrecision(precision),
         ))
+    }
+
+    /// Creates an [Constraint::TimestampOffset] using the offset list specified in it
+    pub fn timestamp_offset(offsets: Vec<TimestampOffset>) -> Constraint {
+        Constraint::TimestampOffset(TimestampOffsetConstraint::new(offsets))
     }
 
     /// Creates a [Constraint::Utf8ByteLength] from a [Range] specifying a length range.
@@ -312,6 +318,9 @@ impl Constraint {
             IslConstraint::Scale(scale_range) => Ok(Constraint::Scale(ScaleConstraint::new(
                 scale_range.to_owned(),
             ))),
+            IslConstraint::TimestampOffset(timestamp_offset) => Ok(Constraint::TimestampOffset(
+                TimestampOffsetConstraint::new(timestamp_offset.valid_offsets().to_vec()),
+            )),
             IslConstraint::TimestampPrecision(timestamp_precision_range) => {
                 Ok(Constraint::TimestampPrecision(
                     TimestampPrecisionConstraint::new(timestamp_precision_range.to_owned()),
@@ -360,6 +369,9 @@ impl Constraint {
             Constraint::Precision(precision) => precision.validate(value, type_store),
             Constraint::Regex(regex) => regex.validate(value, type_store),
             Constraint::Scale(scale) => scale.validate(value, type_store),
+            Constraint::TimestampOffset(timestamp_offset) => {
+                timestamp_offset.validate(value, type_store)
+            }
             Constraint::TimestampPrecision(timestamp_precision) => {
                 timestamp_precision.validate(value, type_store)
             }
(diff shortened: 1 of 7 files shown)


ion-schema/src/isl/util.rs
```diff
@@ -192,3 +194,81 @@ impl Display for ValidValue {
...
+            let hours = TimestampOffset::offset_minutes(
+                string_value.get(1..3).unwrap(),
+                Range::NonNegativeInteger(RangeImpl::new(
+                    RangeBoundaryValue::Value(0, RangeBoundaryType::Inclusive),
+                    RangeBoundaryValue::Value(23, RangeBoundaryType::Inclusive),
+                )?),
+            )?;
+            let minutes = TimestampOffset::offset_minutes(
+                string_value.get(4..6).unwrap(),
+                Range::NonNegativeInteger(RangeImpl::new(
```
popematt, line 260: We should not be using a ISL range for this. We should just use the ranges in the std lib.

ion-schema/src/isl/util.rs
```diff
@@ -192,3 +194,81 @@ impl Display for ValidValue {
...
+                    RangeBoundaryValue::Value(0, RangeBoundaryType::Inclusive),
+                    RangeBoundaryValue::Value(59, RangeBoundaryType::Inclusive),
+                )?),
+            )?;
+            Ok(TimestampOffset::new(Some(sign * (hours * 60 + minutes))))
```
popematt, line 265: Could make it more concise something like this and eliminate a lot of the bespoke parsing logic.
```suggestion
            let (h, _) = string_value.split_at(3);
            let (_, m) = string_value.split_at(4);

            match (h.parse::<i32>(), m.parse::<i32>()) {
                (Ok(hours), Ok(minutes))
                    if (-23..24).contains(&hours) && (0..60).contains(&minutes) =>
                {
                    Ok(TimestampOffset::new(Some(hours * 60 + minutes)))
                }
                _ => invalid_schema_error(
                    "`timestamp_offset` values must be of the form \"[+|-]hh:mm\"",
                ),
            }
```

ion-schema/src/isl/isl_constraint.rs
```diff
@@ -377,6 +383,69 @@ impl IslConstraint {
...
+
+                                // convert the string to TimestampOffset which stores offset in minutes
+                                string_value.try_into()
+                            })
+                            .collect();
+                        list_vec?
+                    }
+                    _ => {
+                        return invalid_schema_error(format!(
+                            "ion type: {:?} is not supported with occurs constraint",
```
popematt, line 440: `IonType` implements `Display`. Constraint should be `timestamp_offset`.
```suggestion
                            "ion type: {} is not supported with timestamp_offset constraint",
```

That being said, how about a message that describes what the correct format is so that the user knows how to correct their mistake, like `"timestamp_offset requires a list of offset strings, but found: {value}"`?

ion-schema/src/isl/isl_constraint.rs
```diff
@@ -392,6 +461,17 @@ impl IslConstraint {
         }
     }
 
+    // helper method to convert from a string to offset in minutes
+    fn offset_minutes(s: &str, range: Range) -> IonSchemaResult<i32> {
```
popematt, line 465: Isn't this duplicated in `util.rs`? Is it necessary to have it in both places?

desaikd: I have removed this method from both the places with the new compact solution you suggested

ion-schema/src/isl/util.rs
desaikd: I still have to consider the sign so I have kept that part otherwise using this solution.

ion-schema/src/constraint.rs
```diff
@@ -1860,3 +1872,48 @@ impl ConstraintValidator for Utf8ByteLengthConstraint {
...
+        // get isl timestamp precision as a range
+        let valid_offsets: &Vec<TimestampOffset> = self.valid_offsets();
+
+        // return a Violation if the value didn't follow timestamp precision constraint
+        if !valid_offsets.contains(&timestamp_value.offset().into()) {
+            return Err(Violation::new(
+                "timestamp_offset",
+                ViolationCode::InvalidLength,
+                &format!(
+                    "expected timestamp offset from {:?} found {:?}",
```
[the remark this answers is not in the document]
desaikd: I am adding Display for `TimestampOffset` and will use that for this violation message.

ion-schema/src/isl/util.rs
```diff
@@ -192,3 +194,79 @@ impl Display for ValidValue {
...
+                _ => {
+                    return invalid_schema_error(format!(
+                        "unrecognized `timestamp_offset` sign '{}'",
+                        &string_value[..1]
+                    ))
+                }
+            };
+            match (h.parse::<i32>(), m.parse::<i32>()) {
+                (Ok(hours), Ok(minutes))
+                    if (-23..24).contains(&hours) && (0..60).contains(&minutes) =>
```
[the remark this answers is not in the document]
desaikd: Oh. Thanks for catching that.

ion-schema/src/isl/util.rs
```diff
@@ -192,3 +194,79 @@ impl Display for ValidValue {
...
+
+impl TimestampOffset {
+    pub fn new(offset: Option<i32>) -> Self {
+        Self { offset }
+    }
+
+    pub fn offset(&self) -> &Option<i32> {
+        &self.offset
+    }
+}
```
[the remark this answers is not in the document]
desaikd: Yes, I think that sounds like a right to me.

popematt (approved):
Looks good!

PR #134, opened 2023-01-05
Updates ion-schema-tests runner

popematt:
### Issue #, if available:

N/A

### Description of changes:

* Updates `ion-schema-tests` to the latest commit, including the tests for Ion Schema 2.0
* Adds a new crate in the workspace with a proc macro that generates a Rust `#[test]` function for test cases in `ion-schema-tests`.
  * `lib.rs` contains the macro function and the structs that define the args to the macro
  * `model.rs` is a model for the test cases in an `.isl` test file and the logic to convert from a `$test::{ /* ... */ }` `Element` to the model.
  * `generator.rs` contains the code that walks the file tree and generates a `mod` for each directory and file.
* Adds an integration test file for ISL 1.0 and 2.0.

I've chosen to keep the tests in the `ion-schema-tests-runner` crate rather than put the tests in the `ion-schema` crate so that `ion-schema` doesn't have to have a dependency on `ion-schema-tests-runner` and we don't have to publish `ion-schema-tests-runner` on crates.io. This is a two-way door, and we can always choose to publish it later if a need arises, but I don't foresee any need at this time.

----


desaikd (approved):
Thanks for working on this. Looks good!

PR #145, opened 2023-02-01
adds changes to allow open content within schema and adds APIs to retrieve open content

desaikd:
### Description of changes:
This PR works on adding changes for allowing open content within schema and adds APIs to retrieve open content.

### List of changes:
* adds `open_content()` in `IslType` as well `IslSchema` to retrieve open content
* adds `load_isl_schema()` that load a schema content as an internal ISL model which can be programmatically manipulated
* adds related tests for open content
_Note_: `IslSchema#open_content()` only provides top level open content. For open content in type definition  added `IslType#open_content()`

*Test:*
- adds unit tests inside `system` and `isl` modules
- allows test files from `ion-schema-tests` to verify open content is allowed

----

### DIFF
--- a/ion-schema/src/isl/isl_constraint.rs
+++ b/ion-schema/src/isl/isl_constraint.rs
@@ -35,6 +35,7 @@ pub enum IslConstraint {
     TimestampOffset(IslTimestampOffsetConstraint),
     TimestampPrecision(Range),
     Type(IslTypeRef),
+    Unknown(String, Element), // Unknown constraint is used to store open contents
     Utf8ByteLength(Range),
     ValidValues(IslValidValuesConstraint),
 }
@@ -446,12 +447,9 @@ impl IslConstraint {
                 RangeType::NonNegativeInteger,
             )?)),
             "valid_values" => Ok(IslConstraint::ValidValues(value.try_into()?)),
-            _ => Err(invalid_schema_error_raw(
-                "Type: ".to_owned()
-                    + type_name
-                    + " can not be built as constraint: "
-                    + constraint_name
-                    + " does not exist",
+            _ => Ok(IslConstraint::Unknown(
+                constraint_name.to_string(),
+                value.to_owned(),
             )),
         }
     }
--- a/ion-schema/src/isl/isl_type.rs
+++ b/ion-schema/src/isl/isl_type.rs
@@ -103,7 +103,7 @@ impl IslTypeImpl {
         open_content
     }
 
-    pub fn is_open_content_allowed(&self) -> bool {
+    pub(crate) fn is_open_content_allowed(&self) -> bool {
         let mut open_content = true;
         if self.constraints.contains(&IslConstraint::ContentClosed) {
             open_content = false;
--- a/ion-schema/src/isl/isl_type.rs
+++ b/ion-schema/src/isl/isl_type.rs
@@ -96,13 +96,8 @@ impl IslTypeImpl {
     pub fn open_content(&self) -> Vec<(String, Element)> {
         let mut open_content = vec![];
         for constraint in &self.constraints {
-            match constraint {
-                IslConstraint::Unknown(constraint_name, element) => {
-                    open_content.push((constraint_name.to_owned(), element.to_owned()))
-                }
-                _ => {
-                    // no op
-                }
+            if let IslConstraint::Unknown(constraint_name, element) = constraint {
+                open_content.push((constraint_name.to_owned(), element.to_owned()))
             }
         }
         open_content
--- a/ion-schema/src/isl/isl_type.rs
+++ b/ion-schema/src/isl/isl_type.rs
@@ -44,7 +44,15 @@ impl IslType {
     }
 
     /// Verifies if the [IslType] allows open content or not
-    pub fn open_content(&self) -> bool {
+    pub fn is_open_content_allowed(&self) -> bool {
+        match &self {
+            IslType::Named(named_type) => named_type.is_open_content_allowed(),
+            IslType::Anonymous(anonymous_type) => anonymous_type.is_open_content_allowed(),
+        }
+    }
+
+    /// Provides open content that is there in the type definition
+    pub fn open_content(&self) -> Vec<(String, Element)> {
         match &self {
             IslType::Named(named_type) => named_type.open_content(),
             IslType::Anonymous(anonymous_type) => anonymous_type.open_content(),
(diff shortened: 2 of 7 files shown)


ion-schema/src/isl/isl_constraint.rs
```diff
@@ -446,12 +447,9 @@ impl IslConstraint {
                 RangeType::NonNegativeInteger,
             )?)),
             "valid_values" => Ok(IslConstraint::ValidValues(value.try_into()?)),
-            _ => Err(invalid_schema_error_raw(
-                "Type: ".to_owned()
-                    + type_name
-                    + " can not be built as constraint: "
-                    + constraint_name
-                    + " does not exist",
+            _ => Ok(IslConstraint::Unknown(
```
popematt, line 450: Is this going to match against `name`? I believe a name should already be captured elsewhere, so it should not be captured here, but can you double check that?

ion-schema/src/isl/isl_type.rs
```diff
@@ -85,7 +93,22 @@ impl IslTypeImpl {
...
+        let mut open_content = vec![];
+        for constraint in &self.constraints {
+            match constraint {
+                IslConstraint::Unknown(constraint_name, element) => {
+                    open_content.push((constraint_name.to_owned(), element.to_owned()))
+                }
+                _ => {
+                    // no op
+                }
+            }
```
popematt, line 106: You should follow Clippy's suggestion here.

ion-schema/src/isl/isl_type.rs
```diff
@@ -44,7 +44,15 @@ impl IslType {
     }
 
     /// Verifies if the [IslType] allows open content or not
-    pub fn open_content(&self) -> bool {
+    pub fn is_open_content_allowed(&self) -> bool {
+        match &self {
+            IslType::Named(named_type) => named_type.is_open_content_allowed(),
+            IslType::Anonymous(anonymous_type) => anonymous_type.is_open_content_allowed(),
+        }
+    }
```
popematt, line 52: Why wouldn't it allow open content? ISL 1.0 always allows open content _in a type definition_. The `content: closed` constraint stops the `fields` constraint from allowing open content in the data that is being validated.

I guess a better question is how/where is this function used?

ion-schema/src/system.rs
```diff
@@ -930,6 +938,34 @@ impl Resolver {
...
+            return match authority.elements(id) {
+                Err(error) => match error {
+                    IonSchemaError::IoError { source } => match source.kind() {
+                        ErrorKind::NotFound => continue,
+                        _ => Err(IonSchemaError::IoError { source }),
+                    },
+                    _ => Err(error),
+                },
+                Ok(schema_content) => self.isl_schema_from_elements(schema_content.into_iter(), id),
+            };
```
popematt, line 965: Is it possible to do something like this so that we don't have so many nested `match`s?
```suggestion
            return match authority.elements(id) {
                Ok(schema_content) => self.isl_schema_from_elements(schema_content.into_iter(), id),
                Err(IonSchemaError::IoError { source: { kind: ErrorKind::NotFound, .. } }) => continue,
                err => err,
            };
```

ion-schema/src/isl/mod.rs
```diff
@@ -93,18 +94,22 @@ pub struct IslSchema {
     types: Vec<IslType>,
     // Represents all the inline IslImportTypes in this schema file.
     inline_imported_types: Vec<IslImportType>,
+    // Represents open content as `Element`s
+    open_content: Vec<Element>,
```
popematt, line 98: This strategy means that we can't tell the difference between, e.g., 
```
$foo
$bar
type::{ name: foo, codepoint_length: 1 }
```
and
```
type::{ name: foo, codepoint_length: 1 }
$foo
$bar
```

If we're going to go with this strategy, we should document the limitations (in the rust doc) and the reason for the decision.

ion-schema/src/system.rs
```diff
@@ -953,6 +989,13 @@ impl SchemaSystem {
             .load_schema(id, &mut TypeStore::default(), None)
     }
 
+    /// Requests each of the provided [`DocumentAuthority`]s, in order, to get ISL model for the
+    /// requested schema id until one successfully resolves it.
+    /// If an authority throws an exception, resolution silently proceeds to the next authority.
+    pub fn load_isl_schema<A: AsRef<str>>(&mut self, id: A) -> IonSchemaResult<IslSchema> {
+        self.resolver.load_isl_schema(id, None)
+    }
```
popematt, line 997: Suppose I already have a `Rc<Schema>`, is there any way for me to get the `IslSchema` without having to reload it from the `Authority`s?

If I have an `IslSchema`, is there any way to go straight to `Rc<Schema>`?

ion-schema/src/isl/isl_type.rs
desaikd: In this context its just used to mark if `content:closed` is used or not?

ion-schema/src/system.rs
desaikd: Currently there is no way to get `IslSchema` from `Rc<Schema>` or get Rc<Schema> from `IslSchema`. 
1. I will create another API that gets `Rc<Schema>` from an `IslSchema`
2. I will have to store `IslSchema` with the resolved `Rc<Schema>`  or create a store where the conversion is mapped. So that one can retrieve `IslSchema` from `Rc<Schema>`

ion-schema/src/isl/mod.rs
desaikd: Added a note for that in the doc comments.

ion-schema/src/isl/isl_type.rs
popematt: I don't think we should expose it as a `pub fn`, so that users don't get confused the way I was. If a user wants to inspect a type and see if it has `content: closed`, they can check the type for `IslConstraint::ContentClosed`, right?

desaikd: Thats correct. I will change it.

ion-schema/src/isl/mod.rs
```diff
@@ -85,26 +86,43 @@ pub mod util;
...
-    // Represents all the IslType defined in this schema file.
-    // For more information: https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#type-definitions
+    /// Represents all the IslType defined in this schema file.
+    /// For more information: https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#type-definitions
     types: Vec<IslType>,
-    // Represents all the inline IslImportTypes in this schema file.
+    /// Represents all the inline IslImportTypes in this schema file.
     inline_imported_types: Vec<IslImportType>,
+    /// Represents open content as `Element`s
+    /// Note: This doesn't preserve the information about where the open content is placed in the schema file.
```
popematt, line 98: Can you document _why_ you have chosen to do it this way? It's possible that in a year, there will be a use case that requires the relative ordering of ISL and non-ISL. If that comes up, we need to understand why the decision was made so that we can make an informed decision about whether to change it.

desaikd: As of now I do not know of any use case for preserving this information and it requires storing the order for non ISL and ISL which could complicate the current `IslSchema` model by having to store the order and add a model to represent non ISL data. If a use case arrives we can change it.

popematt (approved):

PR #148, opened 2023-02-16
adds `IonPath` implementation for `Violation`

desaikd:
### Description of changes:
This PR works on adding changes to add path information for an Ion value that was invalid and resulted into violation.

### List of changes:
* adds implementation of `IonPath` along with `IonPathElement`
* adds `Debug` implementation for `IonPath` as that will be sued for schema sandbox
* `Into<Element>` implementation helps serialize `IonPath` as an SExpression

## Schema
```
type::{
  name: customer,
  type: struct,
  fields: {
    addresses: {
      type: list,
      element: {
        type: struct,
        fields: {
          city: { type: string, codepoint_length: range::[8, max] },
          state: symbol
        }
      }
    }
  }
}
```

## Ion Value to be validated:
```
{
  addresses: [
    { city: "Seattle", state: WA },
    { city: "Portland", state: OR },
    { city: "San Francisco", state: CA }
  ]
}
```

## Output Violation (Using the deeply nested violation for example here):
```
Violation {
    constraint: "{type: string, codepoint_length: range::[8, max]}",
    code: TypeConstraintsUnsatisfied,
    message: "value didn't satisfy type constraint(s)",
    ion_path: ./.addresses[0].city,
    violations: [
        Violation {
            constraint: "codepoint_length",
            code: InvalidLength,
            message: "expected codepoint length range::[ 8, max ] found 7",
            ion_path: ./.addresses[0].city,
            violations: [],
        },
    ],
}
```
----

### DIFF
--- a/ion-schema/src/constraint.rs
+++ b/ion-schema/src/constraint.rs
@@ -1,3 +1,4 @@
+use crate::ion_path::{IonPath, IonPathElement};
 use crate::isl::isl_constraint::{IslConstraint, IslRegexConstraint};
 use crate::isl::isl_range::{Range, RangeImpl};
 use crate::isl::isl_type_reference::IslTypeRef;
@@ -26,7 +27,12 @@ pub trait ConstraintValidator {
     /// adding [Violation]s and/or [ViolationChild]ren to `Err(violation)`
     /// if the constraint is violated.
     /// Otherwise, if the value passes the validation against the constraint then returns `Ok(())`.
-    fn validate(&self, value: &IonSchemaElement, type_store: &TypeStore) -> ValidationResult;
+    fn validate(
+        &self,
+        value: &IonSchemaElement,
+        type_store: &TypeStore,
+        ion_path: &mut IonPath,
+    ) -> ValidationResult;
 }
 
 /// Defines schema Constraints
@@ -342,16 +348,25 @@ impl Constraint {
         }
     }
 
-    pub fn validate(&self, value: &IonSchemaElement, type_store: &TypeStore) -> ValidationResult {
+    pub fn validate(
+        &self,
+        value: &IonSchemaElement,
+        type_store: &TypeStore,
+        ion_path: &mut IonPath,
+    ) -> ValidationResult {
         match self {
-            Constraint::AllOf(all_of) => all_of.validate(value, type_store),
-            Constraint::Annotations(annotations) => annotations.validate(value, type_store),
-            Constraint::AnyOf(any_of) => any_of.validate(value, type_store),
-            Constraint::ByteLength(byte_length) => byte_length.validate(value, type_store),
-            Constraint::CodepointLength(codepoint_length) => {
-                codepoint_length.validate(value, type_store)
+            Constraint::AllOf(all_of) => all_of.validate(value, type_store, ion_path),
+            Constraint::Annotations(annotations) => {
+                annotations.validate(value, type_store, ion_path)
             }
-            Constraint::Contains(contains) => contains.validate(value, type_store),
+            Constraint::AnyOf(any_of) => any_of.validate(value, type_store, ion_path),
+            Constraint::ByteLength(byte_length) => {
+                byte_length.validate(value, type_store, ion_path)
+            }
+            Constraint::CodepointLength(codepoint_length) => {
+                codepoint_length.validate(value, type_store, ion_path)
+            }
+            Constraint::Contains(contains) => contains.validate(value, type_store, ion_path),
             Constraint::ContentClosed => {
                 // No op
                 // `content: closed` does not work as a constraint by its own, it needs to be used with other container constraints
@@ -360,30 +375,34 @@ impl Constraint {
                 Ok(())
             }
             Constraint::ContainerLength(container_length) => {
-                container_length.validate(value, type_store)
+                container_length.validate(value, type_store, ion_path)
             }
-            Constraint::Element(element) => element.validate(value, type_store),
-            Constraint::Fields(fields) => fields.validate(value, type_store),
-            Constraint::Not(not) => not.validate(value, type_store),
-            Constraint::OneOf(one_of) => one_of.validate(value, type_store),
-            Constraint::Type(type_constraint) => type_constraint.validate(value, type_store),
-            Constraint::Occurs(occurs) => occurs.validate(value, type_store),
+            Constraint::Element(element) => element.validate(value, type_store, ion_path),
+            Constraint::Fields(fields) => fields.validate(value, type_store, ion_path),
+            Constraint::Not(not) => not.validate(value, type_store, ion_path),
+            Constraint::OneOf(one_of) => one_of.validate(value, type_store, ion_path),
+            Constraint::Type(type_constraint) => {
+                type_constraint.validate(value, type_store, ion_path)
+            }
+            Constraint::Occurs(occurs) => occurs.validate(value, type_store, ion_path),
             Constraint::OrderedElements(ordered_elements) => {
-                ordered_elements.validate(value, type_store)
+                ordered_elements.validate(value, type_store, ion_path)
             }
-            Constraint::Precision(precision) => precision.validate(value, type_store),
-            Constraint::Regex(regex) => regex.validate(value, type_store),
-            Constraint::Scale(scale) => scale.validate(value, type_store),
+            Constraint::Precision(precision) => precision.validate(value, type_store, ion_path),
+            Constraint::Regex(regex) => regex.validate(value, type_store, ion_path),
+            Constraint::Scale(scale) => scale.validate(value, type_store, ion_path),
             Constraint::TimestampOffset(timestamp_offset) => {
-                timestamp_offset.validate(value, type_store)
+                timestamp_offset.validate(value, type_store, ion_path)
             }
             Constraint::TimestampPrecision(timestamp_precision) => {
-                timestamp_precision.validate(value, type_store)
+                timestamp_precision.validate(value, type_store, ion_path)
             }
             Constraint::Utf8ByteLength(utf8_byte_length) => {
-                utf8_byte_length.validate(value, type_store)
+                utf8_byte_length.validate(value, type_store, ion_path)
             }
-            Constraint::ValidValues(valid_values) => valid_values.validate(value, type_store),
+            Constraint::ValidValues(valid_values) => {
+                valid_values.validate(value, type_store, ion_path)
+            }
             Constraint::Unknown(_, _) => {
                 // No op
                 // `Unknown` represents open content which can be ignored for validation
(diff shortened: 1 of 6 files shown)


ion-schema/src/ion_path.rs
```diff
@@ -0,0 +1,90 @@
...
+                write!(f, "[{index}]")
+            }
+            IonPathElement::Field { name } => {
+                write!(f, ".{name}")
+            }
+        }
+    }
+}
+
+#[derive(Default, Clone, PartialEq)]
```
popematt, line 41: I don't know if we can derive it, but I think implementing `PartialOrd` might be useful for customers who want to extract all of the violation paths to create a flattened list of violations.
* if you want to confirm that a particular piece of data has exactly violations X, Y, and Z;  instead of having to move the violations into a set, you could just compare the sorted violations
* If you want to display a list of violations to a user, it will be easier for a human to read them if they have some logical ordering

popematt (commented):
I have some questions about the path string representation.
```
ion_path: ./.addresses[0].city,
```
I'm assuming that `./` means "root". Since the `ion_path` field is not arbitrary paths, do we really need this?

What if the data being validated is a document rather than a single value? Does the path still start with `./`?

How would it handle a field name like `foo[0]`, `/`, or `foo.bar`?

If we're defining a path expression syntax, it needs to be rigorous enough to handle all of these edge cases.
I would like to suggest an alternative that doesn't have any of these edge cases.

We can leverage Ion to do the heavy lifting for us here by making the ion_path a list or sexp. (I'm going to use a sexp for my examples, but a list would also work.) 
Here are some examples:
```
// Violation at the "root"
()

// Violation at the field "com.amazon.foo" in the second value of a document
// Using current syntax: ./[2].com.amazon.foo
(2 'com.amazon.foo')

// Violation at the field "foo" in the field "amazon" in the field "com" in the second value of a document
// Using current syntax: ./[2].com.amazon.foo
(2 com amazon foo)

// Violation at ./addresses[0].city
(addresses 0 city)

// Violation at the [0] field when validating a top level struct.
('[0]')
```

Benefits of this approach:
* Because we're leveraging Ion, we can be sure that field names, etc. are clearly delimited and properly escaped.
* It's always unambiguous (except when there's duplicate field names, but that's caused by the Ion data model, and no syntax can solve that).
* We can use the `Display` implementation for `Element` rather than having to write our own.
* The format is human-readable and machine-readable.

What do you think?

ion-schema/src/ion_path.rs
```diff
@@ -0,0 +1,90 @@
...
+pub struct IonPath {
+    parents: Vec<IonPathElement>,
+}
+
+impl IonPath {
+    pub fn add_parent(&mut self, parent: IonPathElement) {
```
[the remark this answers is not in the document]
desaikd: Yes, this happened due to a rename that I did before pushing the code. I will change this probably `pop/push` sounds better here.

desaikd:
> We can leverage Ion to do the heavy lifting for us here by making the ion_path a list or sexp. (I'm going to use a sexp for my examples, but a list would also work.) Here are some examples:
> 
> ```
> // Violation at the "root"
> ()
> 
> // Violation at the field "com.amazon.foo" in the second value of a document
> // Using current syntax: ./[2].com.amazon.foo
> (2 'com.amazon.foo')
> 
> // Violation at the field "foo" in the field "amazon" in the field "com" in the second value of a document
> // Using current syntax: ./[2].com.amazon.foo
> (2 com amazon foo)
> 
> // Violation at ./addresses[0].city
> (addresses 0 city)
> 
> // Violation at the [0] field when validating a top level struct.
> ('[0]')
> ```
> 
> What do you think?

Are you suggesting something similar to https://github.com/amazon-ion/ion-java-path-extraction#search-paths. I think that sounds good to me. That way there will not be any ambiguity with the path representation.

popematt:
> Are you suggesting something similar to https://github.com/amazon-ion/ion-java-path-extraction#search-paths. I think that sounds good to me. That way there will not be any ambiguity with the path representation.

There is some similarity, but I would avoid explicitly tying it to the Ion Path Extractor because Ion Path Extractor treats each top-level value in a stream as a separate thing to match against, whereas Ion Schema needs some way to indicate "the whole stream/document" or "the Nth element of the stream/document". In other words, Ion Path Extractor treats every top-level value as a separate root for matching, but Ion Schema treats a value or a stream of values as a single root for validation.

ion-schema/src/constraint.rs
```diff
@@ -682,26 +736,35 @@ impl OrderedElementsConstraint {
...
         values_iter: &mut Peekable<Box<dyn Iterator<Item = &Element> + 'a>>,
         type_store: &TypeStore,
+        ion_path: &mut IonPath,
     ) -> ValidationResult {
         let occurs_range: Range = type_def.get_occurs_constraint("ordered_elements");
 
         // use this counter to keep track of valid values for given type_def
         let mut count: i64 = 0;
+        // use this index to keep track of Ion path for violation
+        let mut idx = 0;
```
zslayton, line 746: Subjective: If you rename this `index`, you can simplify constructing an `IndexedElement` on line 752 from
```rust
IonPathElement::IndexedElement {index: idx}
```
to
```rust
IonPathElement::IndexedElement {index}
```
It also has the nice side effect of using whole-word names, which I happen to prefer. 🙂

ion-schema/src/constraint.rs
```diff
@@ -861,6 +939,11 @@ impl ConstraintValidator for FieldsConstraint {
             let type_def = type_store.get_type_by_id(*type_id).unwrap();
             let values: Vec<&Element> = ion_struct.get_all(field_name).collect();
 
+            // add parent value for current field in ion path
+            ion_path.add_parent(IonPathElement::Field {
```
zslayton, line 943: Low-priority observation: you could implement `From<usize>` and `From<&str>` for `IonPathElement` to trim some boilerplate.

```rust
ion_path.push(7);     // IndexedElement
ion_path.push("foo"); // Field

File: ion-schema/src/constraint.rs [hunk 5]
```diff
@@ -1158,48 +1282,63 @@ impl ConstraintValidator for ElementConstraint {
...
                                 element.ion_type()
                             ),
+                            ion_path,
                         ));
                     }
                 }
             }
             IonSchemaElement::Document(document) => {
-                for val in document {
+                for (idx, val) in document.iter().enumerate() {
```
zslayton, line 1333: Same thought here RE: whole-word names.

ion-schema/src/ion_path.rs
```diff
@@ -0,0 +1,90 @@
...
+        write!(f, "./",)?;
+        for parent in &self.parents {
+            write!(f, "{parent}")?;
+        }
+        Ok(())
+    }
+}
+
+impl fmt::Display for IonPath {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
```
zslayton, line 83: Are these using the same implementation? Could one call the other?

ion-schema/src/ion_path.rs
```diff
@@ -0,0 +1,90 @@
+use ion_rs::value::owned::Element;
+use ion_rs::value::Builder;
+use ion_rs::Symbol;
+use std::fmt;
+use std::fmt::Formatter;
+
+#[derive(Clone, PartialEq)]
+pub enum IonPathElement {
+    IndexedElement { index: usize },
```
zslayton, line 9: I think we can trim the `IndexedElement` variant name since the enum itself already says `Element`.

```suggestion
    Index { index: usize },
```

ion-schema/src/ion_path.rs
```diff
@@ -0,0 +1,90 @@
...
+    Field { name: String },
```
zslayton, line 10: What's the rationale for using
```rust
pub enum IonPathElement {
  Index { index: usize },
  Field { name: String },
}
```
over
```rust
pub enum IonPathElement {
  Index(usize),
  Field(String),
}
```
?

ion-schema/src/ion_path.rs
desaikd: Sure. I will change that.

ion-schema/src/ion_path.rs
desaikd: Usually with 2nd option, I would add a document comment explaining what the parameterized enum variant means but 1st one is self explanatory so I chose the second one. But it ended up having just one parameter(index or field name) so there's not really any benefit from using 1st option now.

desaikd:
New representation of `IonPath` according to suggested changes is as following:

```ion
Violation {
    constraint: "{type: string, codepoint_length: range::[8, max]}",
    code: TypeConstraintsUnsatisfied,
    message: "value didn't satisfy type constraint(s)",
    ion_path: (addresses 0 city),
    violations: [
        Violation {
            constraint: "codepoint_length",
            code: InvalidLength,
            message: "expected codepoint length range::[ 8, max ] found 7",
            ion_path: (addresses 0 city),
            violations: [],
        },
    ],
}
```

ion-schema/src/ion_path.rs
```diff
@@ -2,64 +2,80 @@ use ion_rs::value::owned::Element;
...
     }
 }
 
-#[derive(Default, Clone, PartialEq)]
+/// Represents an IonPath associated with given Ion value for which the violation occurred.
+/// IonPath consists of IonPathElements where each element could either be an index value or a field name.
+///
+/// Note: IonPath is Displayed as an SExpression where each element of SExpression is either an index value or a field name.
+///
+/// Example of IonPath for an Ion value is as following:
```
zslayton, line 40: Minor thing in the docs: I think we need to avoid suggesting that `IonPath` is a standard syntax akin to `JsonPath` or other projects. It's fine for the type name to be `IonPath`, but we should try to use "Ion path" anywhere where we're referring to the abstract idea of a collection of "step ins".

```suggestion
/// Represents the path to the nested Ion value for which the violation occurred.
/// An [IonPath] consists of [IonPathElement]s where each element could either be an index integer or a field name string. [IonPath]s are serialized as s-expressions.
///
/// Here are some examples of serialized paths:
```

ion-schema/src/ion_path.rs
```diff
@@ -2,64 +2,80 @@ use ion_rs::value::owned::Element;
...
+/// ```ion
+/// {
+///     greetings: [ "hello", "hi", "hey" ]
+/// }
+/// ```
+///
+/// For an Ion value given as above if the violation occurred for Ion value "hi" then the associated
```
zslayton, line 47: I suggest starting with an example of the empty path (`()`) referring to the top level struct, then making subsequent examples show increasingly nested paths.

zslayton (approved):
Some superficial documentation comments below.

ion-schema/src/ion_path.rs
desaikd: Got it. I will add that.

PR #149, opened 2023-02-22
adds a state machine implementation for `ordered_elements` constraint

desaikd:
### Description of changes:
This PR works on adding an NFA state machine for validating `ordered_elements` constraint.

### List of changes:
* adds implementation of a state machine to use for validation of `ordered_elements`
* adds changes for using nfa for validation
  * builds nfa using `NfaBuilder` and uses `NfaRun` to validates
`ordered_elements` constraint
  * adds a function to retrieve non negative range boundaries in `Range`
  * modifies to correctly check for overlapping types in
`ordered_elements` constraint (non greedy)

----


zslayton (commented):
I'm realizing that there's some weirdness in the way we're modeling our NFA. In particular, we're asking whether we can enter or exit states, but I think it makes more sense to be asking whether we can take any given transition. That would require shuffling where we're storing some information.

ion-schema/src/constraint.rs
```diff
@@ -731,72 +733,90 @@ impl OrderedElementsConstraint {
...
 
-    /// Validates a type_def for occurs constraint using values_iter
-    fn type_def_occurs_validation<'a>(
-        type_def: &TypeDefinition,
-        values_iter: &mut Peekable<Box<dyn Iterator<Item = &Element> + 'a>>,
-        type_store: &TypeStore,
-        ion_path: &mut IonPath,
-    ) -> ValidationResult {
-        let occurs_range: Range = type_def.get_occurs_constraint("ordered_elements");
+    pub fn build_nfa_from_type_ids(type_ids: &Vec<TypeId>, type_store: &TypeStore) -> NfaRun {
```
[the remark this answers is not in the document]
desaikd: Already changed this.

ion-schema/src/constraint.rs
```diff
@@ -733,84 +733,94 @@ impl OrderedElementsConstraint {
         Ok(OrderedElementsConstraint::new(resolved_types))
     }
 
-    pub fn build_nfa_from_type_ids(type_ids: &Vec<TypeId>, type_store: &TypeStore) -> NfaRun {
+    // Builds an NFA state machine based on given type_ids
```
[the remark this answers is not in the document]
desaikd: Sure. I will add this in the doc comments.

ion-schema/src/nfa.rs
```diff
@@ -0,0 +1,229 @@
...
+    pub fn can_enter(
+        &self,
+        event: Option<Event>,
+        type_id: Option<TypeId>,
+        type_store: &TypeStore,
+    ) -> bool {
+        match &self {
+            State::Initial => false,
+            State::Intermediate { min, max } => {
+                if event.is_none() {
```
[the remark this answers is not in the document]
desaikd: Yes, `END_OF_STREAM` always represents `None`. I like the idea you suggested to use an enum here.

ion-schema/src/nfa.rs
```diff
@@ -0,0 +1,229 @@
...
+            .any(|(id, _)| self.nfa.states[*id] == State::Final)
+    }
+
+    pub fn transition(
+        &mut self,
+        event: Option<Event>,
+        type_store: &TypeStore,
+    ) -> HashMap<usize, usize> {
+        let mut new_states = HashMap::new();
+        for (transition_from_state_id, visits) in self.visits.iter() {
```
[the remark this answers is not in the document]
desaikd: Yes, I think having 'transition' in the name is confusing.

ion-schema/src/nfa.rs
```diff
@@ -0,0 +1,229 @@
...
+            .iter()
+            .any(|(id, _)| self.nfa.states[*id] == State::Final)
+    }
+
+    pub fn transition(
+        &mut self,
+        event: Option<Event>,
+        type_store: &TypeStore,
+    ) -> HashMap<usize, usize> {
+        let mut new_states = HashMap::new();
```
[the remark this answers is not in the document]
desaikd: Correct. It creates new set of state visits for the `NfaRun`.

ion-schema/src/nfa.rs
```diff
@@ -0,0 +1,229 @@
...
+
+#[derive(Default, Debug, Clone, PartialEq)]
+pub struct Nfa {
+    pub(crate) states: Vec<State>,
+    pub(crate) transitions: HashMap<usize, HashSet<Transition>>,
+}
+
+#[derive(Debug, Clone)]
+pub struct NfaRun {
+    pub(crate) visits: HashMap<usize, usize>,
```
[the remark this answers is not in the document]
desaikd: I think in order to keep track of all possible states for transition we will need to keep track for all state visits not just current state.

ion-schema/src/nfa.rs
```diff
@@ -0,0 +1,258 @@
...
+#[derive(Default, Debug, Clone, PartialEq)]
+pub struct Nfa {
+    pub(crate) states: Vec<State>, // represents states with state_id
+    pub(crate) transitions: HashMap<usize, HashSet<Transition>>, // represents transitions between states
+}
+
+/// This is a context which will be used while validating an Ion value for `ordered_elements` constraint using its NFA
+#[derive(Debug, Clone)]
+pub struct NfaRun {
+    pub(crate) visits: HashMap<usize, usize>, // represents a map of (stat_id, visit_count)
```
popematt, line 52:
```suggestion
    pub(crate) visits: HashMap<usize, usize>, // represents a map of (state_id, visit_count)
```

Also, I don't think a hash map is correct in this situation.

What if we have this:
```
ordered_elements: [
  { type: any, occurs: optional },     // S0
  { type: int, occurs: range::[1, 2] } // S1
]
```
And this input:
```
[1, 2, 3]
```

After seeing the first two elements, the current NFA state should include `(S1, 1)` and `(S1, 2)`, but using a hash map means that one of these will clobber the other.

ion-schema/src/nfa.rs
```diff
@@ -0,0 +1,258 @@
...
+    Initial,
+    Intermediate { min: usize, max: usize },
+    Final,
+}
+
+impl State {
+    pub fn allows_to_enter_for_ion_value(
+        &self,
+        event: Event,
+        type_id: Option<TypeId>,
```
popematt, line 218: question—can the `TypeId` be added as a field in `State::Intermediate`?

ion-schema/src/nfa.rs
```diff
@@ -0,0 +1,258 @@
...
+
+    fn evaluate_transitions(
+        &self,
+        source_state_id: usize,
+        transitions: &HashSet<Transition>,
+        can_exit: bool,
+        visits: usize,
+        event: Event,
+        type_store: &TypeStore,
+        new_states: &mut HashMap<usize, usize>,
```
popematt, line 101: Clippy says there are too many arguments. At minimum, I don't think we need to pass around the set of transitions or the type store.

The set of transitions is unnecessary since it can be derived using the source state id:
```
let transitions: &HashSet<Transition> = self.nfa.transitions.get(source_state_id).unwrap();
```

I think the typestore could be store by reference in the `Nfa` struct (unless there's a circular reference problem). I don't see how there would be a circular reference problem storing it in `NfaRun` though, so that could be another option.

ion-schema/src/nfa.rs
```diff
@@ -0,0 +1,258 @@
...
+            if transition.to_optional_state {
+                let mut transitions = self
+                    .nfa
+                    .transitions
+                    .get(&destination_state_id)
+                    .unwrap()
+                    .to_owned();
+
+                // skip the optional state itself
+                transitions.remove(transition);
```
popematt, line 140: This looks like a code smell to me. NFA graph should be in a usable form when you build it, and shouldn't require any modification when evaluating some input against the NFA. Can you explain why this is necessary?

ion-schema/src/nfa.rs
```diff
@@ -0,0 +1,258 @@
...
+impl NfaBuilder {
+    pub fn new() -> NfaBuilder {
+        NfaBuilder {
+            nfa: Nfa::default(),
+        }
+    }
+
+    pub fn build(self) -> Nfa {
+        self.nfa
+    }
```
popematt, line 172: suggestion—this should check some of the invariants of the NFA graph.
- do all of the transitions start at a valid state id?
- do all of the transitions end at a valid state id?

ion-schema/src/nfa.rs
```diff
@@ -0,0 +1,258 @@
...
+            State::Intermediate { min, .. } => *min <= visits,
+            State::Final => false,
+        }
+    }
+
+    pub fn allows_n_visits(&self, visits: usize) -> bool {
+        match self {
+            State::Initial => false,
+            State::Intermediate { max, .. } => *max >= visits,
+            State::Final => false,
```
popematt, line 255: Shouldn't this always allow 1 visit?

ion-schema/src/nfa.rs
```diff
@@ -0,0 +1,258 @@
...
+impl From<Element> for Event {
+    fn from(value: Element) -> Self {
+        Event::Value(value)
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct Transition {
+    destination: usize,      // represents destination state
+    to_optional_state: bool, // represents if destination state is an optional state or not
```
popematt, line 38: I don't understand why `to_optional_state` is necessary. Shouldn't all of that be handled by the min occurs for the state itself?

ion-schema/src/nfa.rs
```diff
@@ -0,0 +1,258 @@
...
+pub struct Nfa {
+    pub(crate) states: Vec<State>, // represents states with state_id
+    pub(crate) transitions: HashMap<usize, HashSet<Transition>>, // represents transitions between states
+}
+
+/// This is a context which will be used while validating an Ion value for `ordered_elements` constraint using its NFA
+#[derive(Debug, Clone)]
+pub struct NfaRun {
+    pub(crate) visits: HashMap<usize, usize>, // represents a map of (stat_id, visit_count)
+    pub(crate) nfa: Rc<Nfa>,                  // reference to the NFA
```
popematt, line 53: nitpick—this comment doesn't actually say anything useful (since we already know from the type name that `Rc<Nfa>` is a reference to the NFA)

ion-schema/src/constraint.rs
```diff
@@ -731,72 +733,97 @@ impl OrderedElementsConstraint {
...
+                nfa_builder.total_states() - 1,
+                nfa_builder.total_states(),
+                Some(*type_id),
+                min == 0,
+            );
+
+            nfa_builder.with_state(State::Intermediate { min, max });
+
+            if max >= 2 {
+                // add a a transition to self for states that have  max >= 2
```
popematt, line 805: nitpick—I think that `> 1` makes the intention a little more clear.
```suggestion
            if max > 1 {
                // add a a transition to self for states that have  max > 1
```

ion-schema/src/constraint.rs
```diff
@@ -731,72 +733,97 @@ impl OrderedElementsConstraint {
...
+    // |   END  |  S3: 1              |
+    // +------------------------------+
+    // As shown above visit count for `END` doesn't have final state in it which means the value resulted to be invalid.
+    //
+    fn build_nfa_from_type_ids(type_ids: &Vec<TypeId>, type_store: &TypeStore) -> NfaRun {
+        let mut nfa_builder = NfaBuilder::new();
+        // add initial state to nfa
+        nfa_builder.with_state(State::Initial);
+
+        for type_id in type_ids {
```
popematt, line 787: Might be useful to have the index of the current element instead of trying to compute it based on the total number of states in the `nfa_builder`.

```suggestion
        for (i, type_id) in type_ids.iter().enumerate() {
```

ion-schema/src/constraint.rs
```diff
@@ -731,72 +733,97 @@ impl OrderedElementsConstraint {
...
-                }
-            }
-        }
+        // add final state transition
+        nfa_builder.with_transition(
+            nfa_builder.total_states() - 1,
+            nfa_builder.total_states(),
+            None,
+            false,
+        );
```
popematt, line 821: What if there's more than one state that can transition to `Final`? How is that handled?

E.g.:
```
ordered_elements: [
  symbol,
  { occurs: optional: type: text },
  { occurs: optional: type: number },
]
```

ion-schema/src/nfa.rs
desaikd: Yes, I can remove this.

ion-schema/src/nfa.rs
desaikd: I am using this to simp0lify building NFA. The NFa that is built also contains transitions with `0` occurrences. Meaning for optional state I will have to traverse through all possible states and them in the Visit-count map that is created.
Example,
```
 NFA:
                                 +------------ 0 -------------+
                                 |                            |
                                 |                            V
     I(INITIAL) ----> S1(INTERMEDIATE(0, 1)) -- 1 --> S2(INTERMEDIATE(1, 1)) -- 1 --> S3(INTERMEDIATE(1, 1)) ----> F(FINAL)
```

ion-schema/src/nfa.rs
desaikd: Makes sense. Thanks for catching that.

ion-schema/src/nfa.rs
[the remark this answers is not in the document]
zslayton: > I think in order to keep track of all possible states for transition

I had been thinking of an `NfaRun` as a single "run" through the state machine (which might get cloned):

```rust
struct NfaRun {
  current_state: usize,
  visits_to_current_state: usize,
}
```
and that an evaluation of the NFA would produce many `NfaRun`s that would be tracked/updated. I was picturing something like this:

```rust
struct NfaEvaluation {
  runs: Vec<NfaRun>
}
```

I think you're achieving the same thing, but with `NfaRun` as the overarching data structure that stores all of the individual runs. If so, that works but I think we should tweak the name and add a doc comment to explain the type's intent.

desaikd:
As per an offline discussion, moving all the information to `Transition` instead of `State` resolves the weirdness around entry/exit condition checks on states instead having those checks to see if we can take a given transition or not.

List of change:
* adds changes to store all information in `Transition` instead of states.
* removes `State` as all states are now intermediate states which
removes complexity invoved with types of states
* moves all entry/exit methods for `State` to `Transition`
* states are now simply `usize` which its represents states id
* NFa build process now creates a set of states that are final states or
accepted states of NFA

ion-schema/src/nfa.rs
```diff
@@ -4,135 +4,219 @@ use crate::types::TypeValidator;
...
-                    if can_enter {
-                        new_states.insert(transition_to_state_id, visits + 1);
+        let mut element = element;
+        let mut destination_states_for_optional_state: HashSet<Transition> = HashSet::new();
+
+        let source_state_id = nfa_run.state_id;
+        let visits = nfa_run.state_visits;
+
+        // traverse all transitions for source state and verify if we can take the given transition
+        for transition in transitions {
```
zslayton, line 169: This method is relatively complex (which is fine, this is the heart of the NFA logic!), but remember that future readers may not know what an NFA is or does. I think it'd be good to break this method into some smaller helper methods so it's easier to see the three main cases it tries to address:
1. Transition loops back to the same state
2. Transition moves to the next state
3. Transition moves to an optional state

ion-schema/src/nfa.rs
```diff
@@ -4,135 +4,219 @@ use crate::types::TypeValidator;
...
+            let destination_state_id = transition.destination;
+
+            if destination_state_id == source_state_id {
```
zslayton, line 172: I think each case needs a high-level description comment explaining it.

ion-schema/src/nfa.rs
```diff
@@ -4,135 +4,219 @@ use crate::types::TypeValidator;
...
+                let mut visit_count = visits + 1;
+                if transition.allows_n_visits(visit_count) {
+                    if transition.is_valid_for_ion_value(element, type_store) {
+                        // if transition is valid, add destination state to next states
+                        next_states.insert(NfaRun::new(destination_state_id, visit_count));
+                    }
+                    // consume elements for at least minimum n visits
```
zslayton, line 179: What does 'consume' mean in this context? What does "minimum n visits" refer to?

To make the comments higher level, consider walking through what the function does at the beginning. Then you can refer to these concepts more tersely later in the function body.

ion-schema/src/nfa.rs
```diff
@@ -4,135 +4,219 @@ use crate::types::TypeValidator;
...
+        // use nfa_evaluation for validation
+        while let Some(element) = elements_iter.next() {
+            let mut next_states = HashSet::new();
+            for nfa_run in self.visits.iter() {
+                let transitions: HashSet<Transition> = self.nfa.get_transitions(nfa_run.state_id);
+                self.evaluate_transitions(
+                    nfa_run,
+                    transitions,
+                    element,
+                    &mut elements_iter,
```
zslayton, line 144: Curious why we're passing through a `Peekable<Iter<Element>>` instead of just the `&[Element]`.

zslayton (commented):
Looking good! I think it could use some more explanatory comments for future readers, though.

ion-schema/src/constraint.rs
```diff
@@ -782,51 +779,41 @@ impl OrderedElementsConstraint {
...
-                nfa_builder.with_transition(
-                    nfa_builder.total_states() - 1,
-                    nfa_builder.total_states() - 1,
-                    Some(*type_id),
-                    min == 0,
-                );
+            // if the current state is required then that is the only final state till now
+            if min > 0 {
+                // remove all previous final states
+                final_states.clear();
```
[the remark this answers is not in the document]
desaikd: The problem with that would be I will have two traverse through all possible transitions as the states or node itself doesn't have any information about min, max occurrence count to decide whether it is optional or non optional.

ion-schema/src/nfa.rs
```diff
@@ -4,135 +4,219 @@ use crate::types::TypeValidator;
...
+    pub fn has_final_state(&self, type_store: &TypeStore) -> bool {
+        if self.nfa.get_final_states().is_empty() && self.visits.is_empty() {
+            return true;
+        }
+
+        self.visits.iter().any(|nfa_run| {
+            self.nfa.get_final_states().iter().any(|fs| {
+                fs.state_id == nfa_run.state_id
+                    && nfa_run.state_visits >= fs.min
+                    && nfa_run.state_visits <= fs.max
```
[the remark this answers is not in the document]
desaikd: It is validated at each transition but when an invalid Ion value is passed through transitions its possible that it reaches final state but with incorrect count. Meaning if the invalid Ion value reaches final state but doesn't satisfy minimum occurrence constraint then it would lead to an incorrect visit count. 

Example from `ion-schema-tests`:
```
// for the following schema:
type::{
  name: ordered_elements_occurs_1,
  ordered_elements: [
    { type: int, occurs: 2 }, // S0
    { type: string, occurs: range::[0, 3] }, // S1
    { type: symbol, occurs: range::[0, max] }, // S2
  ],
}

// Invalid Ion value:
(1)
(1 "a" w)

// nfa_run results fro `(1)`
1 --> {NfaRun { state_id: 0, state_visits: 1 }}

// nfa_run results fro `(1 "a" w)`
1 --> {NfaRun { state_id: 0, state_visits: 1 }}
"a" --> {NfaRun { state_id: 0, state_visits: 1 }}
w --> {NfaRun { state_id: 0, state_visits: 1 }}
```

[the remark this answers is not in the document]
desaikd: Alternate would be to remove final state from next set of `NfaRun`s when a minimum occurrence isn't met.

ion-schema/src/nfa.rs
```diff
@@ -4,135 +4,219 @@ use crate::types::TypeValidator;
...
             nfa,
         }
     }
 
-    pub fn has_final_state(&self) -> bool {
-        self.visits
-            .iter()
-            .any(|(id, _)| self.nfa.states[*id] == State::Final)
+    pub fn has_final_state(&self, type_store: &TypeStore) -> bool {
+        if self.nfa.get_final_states().is_empty() && self.visits.is_empty() {
```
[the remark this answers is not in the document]
desaikd: It is possible that the `ordered_elements` constraint was empty meaning there were no states or transitions added and no final states created for it. Hence added this check to ensure we do have some final states to to verify with last set of visits or `NfaRun`s.

ion-schema/src/nfa.rs
desaikd: In order to move all information to transitions, I had to change the way can_exit check was working. Previously with can_exit check I would have to verify it for each element for given Ion Elements. But now I am using a peekable iterator to just satisfy the minimum occurrence at once and then exit that state. This removes the need to check for can_exit everytime for an element.

ion-schema/src/nfa.rs
```diff
@@ -4,135 +4,219 @@ use crate::types::TypeValidator;
...
-            self.to_states_transitions(
-                transition_from_state_id,
-                transition_to_states,
-                can_exit,
-                visits,
-                event.to_owned(),
-                type_store,
-                &mut new_states,
-            );
+    pub fn validate_ordered_elements(&mut self, elements: Vec<Element>, type_store: &TypeStore) {
```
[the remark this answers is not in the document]
desaikd: Sure. I will change that.

ion-schema/src/nfa.rs
```diff
@@ -4,135 +4,219 @@ use crate::types::TypeValidator;
...
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct Transition {
-    destination: usize,      // represents destination state
-    to_optional_state: bool, // represents if destination state is an optional state or not
-    type_id: Option<TypeId>, // None is used for final state transition
+    destination: StateId, // represents destination state
+    type_id: TypeId,
+    min: usize, // minimum occurrence allowed for destination state
+    max: usize, // maximum occurrence allowed for destination state
```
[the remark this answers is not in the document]
desaikd: Yes, we only care about that when destination state is equal to source state. And helps identify optional states when `min == 0`.

ion-schema/src/constraint.rs
```diff
@@ -733,17 +733,14 @@ impl OrderedElementsConstraint {
...
+    // Builds an NFA state machine based on given type_ids. This is a limited form of NFA where state machine is linear and every transition either leads to itself or the next state.
     //
     // All the states has some transitions between them leading from one state to another or back to itself.
-    // All the intermediate states that have minimum occurrence as 0 are optional states meaning those states can lead to another state with 0 occurrence event transitions.
-    // There two special case in transitions as following:
-    // * max >= 2 -This represents transitions to self for a state. Any intermediate states with max >= 2 can transition back to itself.
-    // * max == 0 -This represents transitions skipping the optional state. Any intermediate state that has min == 0 can be skipped to transition to its next state.
+    // All the intermediate states that have a minimum occurrence of 0 are optional states, meaning those states can lead to another state with 0 occurrence event transitions.
+    // There are two special cases of transition that need to be handled.
+    // For any state whose corresponding `type_id` has an `occurs` where:
```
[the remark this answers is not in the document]
desaikd: Yes. Only `intermediate state` will now be just `state`.

ion-schema/src/nfa.rs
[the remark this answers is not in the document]
zslayton: Ok, cool. Can you add a comment explaining that?

ion-schema/src/nfa.rs
```diff
@@ -79,7 +91,9 @@ impl Nfa {
     }
 }
 
-/// A context that will be used for a single transition
+/// A context that will be used to store state-visit counts.
+/// This will be used by [NfaEvaluation] which uses [Nfa] to evaluate given Ion value.
+/// With each element in the given ordered elements there will be a set of [NfaRun]s created for all possible transitions to next states.
```
zslayton, line 96:
```suggestion
/// For each step in the NFA's evalution, each existing `NfaRun` will be cloned once for each possible transition.
```
Is this accurate? I'm hoping to capture the idea that an NFA run is one of the things moving through the state machine and that it gets cloned a potentially large number of times during evaluation.

zslayton (approved):
Some superficial notes below.

ion-schema/src/nfa.rs
desaikd: That is correct. For each possible and correct transition a new `NfaRun` gets created.

PR #152, opened 2023-03-17
adds initial implementation to support ISL 2.0

desaikd:
### Description of changes:
This PR works on adding version configuration changes to allow ISL 2.0 support.

### List of changes:
- fixes an issue with range construction which had a missing type check for range construction (95e97c973dfa06083227bfabda4813cceddaaefe)
- adds schema versioning configuration for all internal methods that were used to load schema or ISL model using an enum `IonSchemaLanguageVersion`(529fa2d02425d65bbda3062cda87ae0b5756c89e)
- `isl_type`, `isl_type_reference` and `isl_constraint` contains two modules with specific implementation of types/type references and constraints based on ISL version(`v_1_0` and `v_2_0`). This will help with easier API modifications with new major/minor version release.
- All the previously defined schema model structs/enums are now hidden inside wrapper classes which helps with hiding implementation details based on ISL versioning.
  -  e.g. `IslType` which was previously used to model an ISL type definition is now renamed as `IslTypeKind`. `IslTypeKind` has all the implementation details required to support both/all ISL version with a single struct. A new wrapper class that takes in this `IslTypeKind` to hide its implementation details. Along with that now `IslTypeKind` will be private to crate and users can only construct for `IslType` and use its accessor methods. In conclusion, all the wrapper classes are added to separate out implementation details  from public facing APIs/structs.
  - Similar to `IslType` following structures are changed to add on a layer of abstraction from implementation layer.
    -  `IslConstraint` enum has now changed to be called `IslConstraintImpl`. A wrapper class `IslCosntraint` is added on top of it that takes `IslConstraintImpl` to create the abstraction layer. 
    - Similarly `IslTypeRef` is now called `IslTypeRefImpl`. And it has wrapper class that is called `IslTypeRef`.
- added `UserReservedFields` struct that stores user reserved fields for ISL 2.0. 
- added validation for user reserved fields for ISL 2.0. 
- adds methods to `schema_system` that converts ISL model to a resolved schema called `schema_from_isl_schema`. 
   - `schema_from_isl_schema` makes sure that the given ISL model types contains the same version of ISL for their cosntraints. This check is not done while constructing the ISL model.
   - `v_1_0` and `v_2_0` modules does give user an idea on which ISL version they are using to construct the model but it does not guarantee that the model that is created has all the constraints from same version of ISL.  
- `IslSchema` now has two methods to construct schema for a particular version `schema_v_1_0` and `schema_v_2_0`.

### Example construction of ISL model programmatically using the new signature:
Using a single ISL type to construct ISL schema model as below:
```rust
// an ISL 1.0 type that contains ISL 1.0 related constraints
let isl_type = isl_type::v_1_0::named_type(
    "my_type",
    [isl_constraint::v_1_0::element(
        isl_type_reference::v_1_0::named_type_ref("int"),
    )],
);
let isl = IslSchema::schema_v_1_0("sample.isl", vec![], vec![isl_type], vec![], vec![]);
```

### DIFF
--- a/ion-schema-tests-runner/src/generator.rs
+++ b/ion-schema-tests-runner/src/generator.rs
@@ -177,26 +177,32 @@ fn generate_test_cases_for_file(ctx: Context) -> TokenStream {
 
 /// find ISL version from schema content
 fn find_isl_version(schema_content: &[Element]) -> IonSchemaLanguageVersion {
-    let isl_version_marker = Regex::new(r"^\$ion_schema_\d.*$").unwrap();
-    let mut isl_version = IonSchemaLanguageVersion::V1_0;
+    // ISL version marker regex
+    let isl_version_marker: Regex = Regex::new(r"^\$ion_schema_\d.*$").unwrap();
 
-    // find the ISL version
     for value in schema_content {
+        // if find a type definition or a schema header before finding any version marker then this is ISL 1.0
+        if value.ion_type() == IonType::Struct
+            && (value.has_annotation("type") || value.has_annotation("schema_header"))
+        {
+            // default ISL 1.0 version will be returned
+            break;
+        }
         // verify if value is an ISL version marker and if it has valid format
         if value.ion_type() == IonType::Symbol
             && isl_version_marker.is_match(value.as_str().unwrap())
         {
             // This implementation supports Ion Schema 1.0 and Ion Schema 2.0
-            isl_version = match value.as_str().unwrap() {
+            return match value.as_str().unwrap() {
                 "$ion_schema_1_0" => IonSchemaLanguageVersion::V1_0,
                 "$ion_schema_2_0" => IonSchemaLanguageVersion::V2_0,
-                _ => {
-                    unimplemented!("Unsupported Ion Schema Language version: {}", value)
-                }
+                _ => unimplemented!("Unsupported Ion Schema Language version: {}", value),
             };
         }
     }
-    isl_version
+
+    // default ISL version 1.0 if no version marker is found
+    IonSchemaLanguageVersion::V1_0
 }
 
 /// Generates a test case to assert that some Ion text is or is not a valid ISL schema document.
@@ -243,7 +249,11 @@ fn generate_invalid_type_case(
 ) -> TokenStream {
     let invalid_type_text_token = TokenTree::from(Literal::string(invalid_type_text));
     let test_name = format_ident!("{}", util::escape_to_ident(description));
-    let isl_version = TokenTree::from(Literal::string(&format!("{isl_version}")));
+    let version = match isl_version {
+        IonSchemaLanguageVersion::V1_0 => "$ion_schema_1_0",
+        IonSchemaLanguageVersion::V2_0 => "$ion_schema_2_0",
+    };
+    let isl_version = TokenTree::from(Literal::string(version));
     quote! {
         #[test]
         fn #test_name() -> Result<(), String> {
--- a/ion-schema-tests-runner/src/generator.rs
+++ b/ion-schema-tests-runner/src/generator.rs
@@ -5,7 +5,7 @@ use ion_rs::value::reader::element_reader;
 use ion_rs::value::reader::ElementReader;
 use ion_rs::value::IonElement;
 use ion_rs::IonType;
-use ion_schema::isl::IonSchemaLanguageVersion;
+use ion_schema::isl::IslVersion;
 use proc_macro2::{Literal, TokenStream, TokenTree};
 use quote::{format_ident, quote};
 use regex::Regex;
@@ -176,7 +176,7 @@ fn generate_test_cases_for_file(ctx: Context) -> TokenStream {
 }
 
 /// find ISL version from schema content
-fn find_isl_version(schema_content: &[Element]) -> IonSchemaLanguageVersion {
+fn find_isl_version(schema_content: &[Element]) -> IslVersion {
     // ISL version marker regex
     let isl_version_marker: Regex = Regex::new(r"^\$ion_schema_\d.*$").unwrap();
 
@@ -194,15 +194,15 @@ fn find_isl_version(schema_content: &[Element]) -> IonSchemaLanguageVersion {
         {
             // This implementation supports Ion Schema 1.0 and Ion Schema 2.0
             return match value.as_str().unwrap() {
-                "$ion_schema_1_0" => IonSchemaLanguageVersion::V1_0,
-                "$ion_schema_2_0" => IonSchemaLanguageVersion::V2_0,
+                "$ion_schema_1_0" => IslVersion::V1_0,
+                "$ion_schema_2_0" => IslVersion::V2_0,
                 _ => unimplemented!("Unsupported Ion Schema Language version: {}", value),
             };
         }
     }
 
     // default ISL version 1.0 if no version marker is found
-    IonSchemaLanguageVersion::V1_0
+    IslVersion::V1_0
 }
 
 /// Generates a test case to assert that some Ion text is or is not a valid ISL schema document.
@@ -245,13 +245,13 @@ fn generate_value_test_case(
 fn generate_invalid_type_case(
     description: &str,
     invalid_type_text: &str,
-    isl_version: &IonSchemaLanguageVersion,
+    isl_version: &IslVersion,
 ) -> TokenStream {
     let invalid_type_text_token = TokenTree::from(Literal::string(invalid_type_text));
     let test_name = format_ident!("{}", util::escape_to_ident(description));
     let version = match isl_version {
-        IonSchemaLanguageVersion::V1_0 => "$ion_schema_1_0",
-        IonSchemaLanguageVersion::V2_0 => "$ion_schema_2_0",
+        IslVersion::V1_0 => "$ion_schema_1_0",
+        IslVersion::V2_0 => "$ion_schema_2_0",
     };
     let isl_version = TokenTree::from(Literal::string(version));
     quote! {
--- a/ion-schema-tests-runner/src/generator.rs
+++ b/ion-schema-tests-runner/src/generator.rs
@@ -1,10 +1,14 @@
 use crate::generator::util::*;
 use crate::model::{TestCaseDetails, TestCaseVec};
+use ion_rs::value::owned::Element;
 use ion_rs::value::reader::element_reader;
 use ion_rs::value::reader::ElementReader;
 use ion_rs::value::IonElement;
+use ion_rs::IonType;
+use ion_schema::isl::IonSchemaLanguageVersion;
 use proc_macro2::{Literal, TokenStream, TokenTree};
 use quote::{format_ident, quote};
+use regex::Regex;
 use std::fs;
 use std::path::{Path, PathBuf};
 use std::rc::Rc;
@@ -123,6 +127,8 @@ fn generate_test_cases_for_file(ctx: Context) -> TokenStream {
         .read_all(&ion_content)
         .unwrap_or_else(|e| panic!("Error in {path_string} – {e:?}"));
 
+    let isl_version = find_isl_version(&schema_content);
+
     for element in schema_content {
         if element.has_annotation("$test") {
             let test_cases: TestCaseVec = element
@@ -143,9 +149,9 @@ fn generate_test_cases_for_file(ctx: Context) -> TokenStream {
                         &schema_text,
                         expect_valid,
                     )),
-                    TestCaseDetails::InvalidType { type_text } => {
-                        test_case_tokens.push(generate_invalid_type_case(&description, &type_text))
-                    }
+                    TestCaseDetails::InvalidType { type_text } => test_case_tokens.push(
+                        generate_invalid_type_case(&description, &type_text, &isl_version),
+                    ),
                     TestCaseDetails::Value {
                         type_id,
                         value_text,
@@ -167,6 +173,30 @@ fn generate_test_cases_for_file(ctx: Context) -> TokenStream {
     }
 
     join_token_streams(test_case_tokens)
+}
+
+/// find ISL version from schema content
+fn find_isl_version(schema_content: &[Element]) -> IonSchemaLanguageVersion {
+    let isl_version_marker = Regex::new(r"^\$ion_schema_\d.*$").unwrap();
+    let mut isl_version = IonSchemaLanguageVersion::V1_0;
+
+    // find the ISL version
+    for value in schema_content {
+        // verify if value is an ISL version marker and if it has valid format
+        if value.ion_type() == IonType::Symbol
+            && isl_version_marker.is_match(value.as_str().unwrap())
+        {
+            // This implementation supports Ion Schema 1.0 and Ion Schema 2.0
+            isl_version = match value.as_str().unwrap() {
+                "$ion_schema_1_0" => IonSchemaLanguageVersion::V1_0,
+                "$ion_schema_2_0" => IonSchemaLanguageVersion::V2_0,
+                _ => {
+                    unimplemented!("Unsupported Ion Schema Language version: {}", value)
+                }
+            };
+        }
+    }
+    isl_version
 }
 
 /// Generates a test case to assert that some Ion text is or is not a valid ISL schema document.
@@ -206,13 +236,18 @@ fn generate_value_test_case(
 }
 
 /// Generates a test case for an invalid ISL type.
-fn generate_invalid_type_case(description: &str, invalid_type_text: &str) -> TokenStream {
+fn generate_invalid_type_case(
+    description: &str,
+    invalid_type_text: &str,
+    isl_version: &IonSchemaLanguageVersion,
+) -> TokenStream {
     let invalid_type_text_token = TokenTree::from(Literal::string(invalid_type_text));
     let test_name = format_ident!("{}", util::escape_to_ident(description));
+    let isl_version = TokenTree::from(Literal::string(&format!("{isl_version}")));
     quote! {
         #[test]
         fn #test_name() -> Result<(), String> {
-            __assert_type_is_invalid(#invalid_type_text_token)
+            __assert_type_is_invalid(#invalid_type_text_token, #isl_version)
         }
     }
 }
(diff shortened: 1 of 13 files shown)


ion-schema/src/constraint.rs
```diff
@@ -218,132 +219,170 @@ impl Constraint {
         Ok(Constraint::Regex(regex.try_into()?))
     }
 
+    /// Resolves all ISL type references to corresponding [TypeId]s
+    fn resolve_type_references_to_type_ids(
```
desaikd, line 223: Previously each constraint had its own method to parse type ids. This is just a reusable method that can be used in multiple places.

ion-schema-tests-runner/src/generator.rs
```diff
@@ -123,6 +127,8 @@ fn generate_test_cases_for_file(ctx: Context) -> TokenStream {
         .read_all(&ion_content)
         .unwrap_or_else(|e| panic!("Error in {path_string} – {e:?}"));
 
+    let isl_version = find_isl_version(&schema_content);
```
desaikd, line 130: This file contains changes for ion-schema-tests-runner to consider version information when testing for invalid types.

ion-schema/src/isl/isl_type.rs
```diff
@@ -1,61 +1,130 @@
...
+
 /// Represents a type in an ISL schema.
 #[derive(Debug, Clone, PartialEq)]
-pub enum IslType {
-    Named(IslTypeImpl),
-    Anonymous(IslTypeImpl),
+pub struct IslType {
+    pub(crate) kind: IslTypeKind,
+    constraints: Vec<IslConstraint>,
 }
```
popematt, line 67: I am confused about this change.

We now have `IslType` which contains a `Vec<IslConstraint>` and an `IslTypeKind`. The `IslTypeKind` is an enum where both variants contain `IslTypeImpl`, and then `IslTypeImpl` contains an optional name and a `Vec<IslConstraintImpl>`.

- What is the point of `IslTypeKind`? Why not just have `name: Option<String>` in either `IslType` or `IslTypeImpl` and get rid of the extra indirection?
- We already have the separate internal model. What exactly is gained by having both `IslType` and `IslTypeImpl`? How does it help hide version information?

ion-schema/src/system.rs
```diff
@@ -862,27 +929,65 @@ impl Resolver {
...
+                        }
+                        IonSchemaLanguageVersion::V2_0 => {
+                            // verify if there are any constraints with ISL 1.0 for this isl_type
+                            let has_isl_1_0_constraints =
+                                isl_type.constraints().iter().any(|c| match c.constraint {
+                                    IslConstraintImpl::Regex(_)
+                                    | IslConstraintImpl::Element(_)
+                                    | IslConstraintImpl::Annotations(_)
+                                    | IslConstraintImpl::Scale(_) => {
+                                        c.version == IonSchemaLanguageVersion::V2_0
```
popematt, line 970: Should this be `IonSchemaLanguageVersion::V1_0`?

ion-schema/src/system.rs
```diff
@@ -862,27 +929,65 @@ impl Resolver {
...
+                            let has_isl_2_0_constraints =
+                                isl_type.constraints().iter().any(|c| match c.constraint {
+                                    IslConstraintImpl::Regex(_)
+                                    | IslConstraintImpl::Element(_)
+                                    | IslConstraintImpl::Annotations(_)
+                                    | IslConstraintImpl::Scale(_) => {
+                                        c.version == IonSchemaLanguageVersion::V2_0
+                                    }
+                                    _ => false,
+                                });
```
popematt, line 956:
```suggestion
                                isl_type.constraints().iter().any(|c| c.version != IonSchemaLanguageVersion::V1_0);
```
It's much more readable if you just check the version of all of the constraints. (It's also fewer operations.)

ion-schema/src/system.rs
```diff
@@ -931,30 +1039,57 @@ impl Resolver {
                     _ => Err(error),
                 },
                 Ok(schema_content) => {
-                    let isl = self.isl_schema_from_elements(schema_content.into_iter(), id)?;
-                    self.schema_from_isl_schema(isl, id, type_store, load_isl_import)
+                    // ISL version marker regex
+                    let isl_version_marker = Regex::new(r"^\$ion_schema_\d.*$").unwrap();
```
popematt, line 1043: Compiling a `Regex` can be expensive. Can you move this out of the `authorities` loop?

ion-schema/src/system.rs
```diff
@@ -931,30 +1039,57 @@ impl Resolver {
...
+                    // find the ISL version
+                    for value in &schema_content {
+                        // verify if value is an ISL version marker and if it has valid format
+                        if value.ion_type() == IonType::Symbol
+                            && isl_version_marker.is_match(value.as_str().unwrap())
+                        {
+                            // This implementation supports Ion Schema 1.0 and Ion Schema 2.0
+                            isl_version = match value.as_str().unwrap() {
+                                "$ion_schema_1_0" => IonSchemaLanguageVersion::V1_0,
+                                "$ion_schema_2_0" => IonSchemaLanguageVersion::V2_0,
```
popematt, line 1055: It should probably break out of the loop immediately after finding either of these.

ion-schema/src/system.rs
```diff
@@ -931,30 +1039,57 @@ impl Resolver {
...
-                    let isl = self.isl_schema_from_elements(schema_content.into_iter(), id)?;
-                    self.schema_from_isl_schema(isl, id, type_store, load_isl_import)
+                    // ISL version marker regex
+                    let isl_version_marker = Regex::new(r"^\$ion_schema_\d.*$").unwrap();
+                    let mut isl_version = IonSchemaLanguageVersion::V1_0;
+
+                    // find the ISL version
+                    for value in &schema_content {
+                        // verify if value is an ISL version marker and if it has valid format
+                        if value.ion_type() == IonType::Symbol
```
popematt, line 1049: If `value` is an Ion struct that is annotated with `type::` or `schema_header::` (and no version marker has been encountered yet), then we know it's ISL 1.0, and we can break out of the loop.

ion-schema/src/system.rs
```diff
@@ -987,10 +1122,49 @@ impl SchemaSystem {
     }
 
     /// Requests each of the provided [`DocumentAuthority`]s, in order, to get ISL model for the
-    /// requested schema id until one successfully resolves it.
+    /// requested schema id until one successfully resolves it using ISL 1.0.
+    /// If an authority throws an exception, resolution silently proceeds to the next authority.
+    /// If the given ISL model has any ISL 2.0 related types/constraints, resolution returns an error.
+    pub fn load_isl_schema_v1_0<A: AsRef<str>>(&mut self, id: A) -> IonSchemaResult<IslSchema> {
```
popematt, line 1128: I'm a little confused by the function signature here. Does this mean that you need to know the ISL version of e.g. `foo.isl` before loading `foo.isl` so that you call the correct function? If so, we need to change it—it's not reasonable for a user to have to determine the ISL version _before_ loading the schema.

ion-schema/src/system.rs
```diff
@@ -1835,4 +1986,108 @@ mod schema_system_tests {
...
+                }
+                
+                type::{
+                  name: my_type,
+                  type: string,
+                }
+                
+                schema_footer::{
+                    baz: "this is an unexpected field in schema header"
+                }
```
popematt, line 2038: According to the spec, this is an ISL 1.0 schema document (since there is no `$ion_schema_2_0` marker), and so it should be valid to have all of these extra fields.

ion-schema/src/system.rs
```diff
@@ -862,27 +929,65 @@ impl Resolver {
...
 
         // Resolve all ISL types and constraints
         for isl_type in isl.types() {
             let pending_types = &mut PendingTypes::default();
 
-            match isl_type {
-                IslType::Named(named_isl_type) => {
+            match &isl_type.kind {
+                IslTypeKind::Named(named_isl_type) => {
+                    match isl_version {
```
popematt, line 944: The match arms here are nearly identical. It should be possible to get rid of the match entirely using something like this: 
```
isl_type.constraints().iter().any(|c| c.version != isl_version)
```

ion-schema/src/system.rs
```diff
@@ -862,27 +929,65 @@ impl Resolver {
...
+                            isl_version,
                             named_isl_type,
                             type_store,
                             pending_types,
                         )?;
                 }
-                IslType::Anonymous(_) => {
+                IslTypeKind::Anonymous(_) => {
                     unreachable!("Top level ISL type definitions are always named type definitions")
                 }
```
popematt, line 992: If this is truly unreachable, then I don't think you need to match on `kind` at all.

ion-schema-tests-runner/src/generator.rs
```diff
@@ -169,6 +175,30 @@ fn generate_test_cases_for_file(ctx: Context) -> TokenStream {
     join_token_streams(test_case_tokens)
 }
 
+/// find ISL version from schema content
+fn find_isl_version(schema_content: &[Element]) -> IonSchemaLanguageVersion {
```
popematt, line 179: The logic of this function is not correct. See my comments on the other copy of this function.

ion-schema-tests-runner/tests/ion-schema-tests-1-0.rs
```diff
@@ -9,8 +9,6 @@ ion_schema_tests!(
         "constraints::occurs::invalid::occurs_for_a_field_must_be_a_positive_integer_or_a_non_empty__non_negative_integer_range__7_",
         "constraints::occurs::invalid::occurs_must_be_a_positive_integer_or_non_empty__non_negative_integer_range__06_",
         "constraints::occurs::invalid::occurs_range_must_be_a_valid__satisfiable_range__1_",
-        "constraints::scale::invalid::scale_must_be_an_integer_or_range__10_",
-        "constraints::scale::invalid::scale_must_be_an_integer_or_range__11_",
```
popematt, line 13: Just curious—is this something that is fixed in this PR or are these just tests that were incorrectly being skipped?

desaikd: No this was due to an issue in `Range`s. All the changes in the `isl_range` module are related to that.   

> fixes an issue with range construction which had a missing type check for range construction (https://github.com/amazon-ion/ion-schema-rust/commit/95e97c973dfa06083227bfabda4813cceddaaefe)

ion-schema/src/isl/isl_range.rs
```diff
@@ -476,7 +476,8 @@ impl<T: std::cmp::PartialOrd> RangeImpl<T> {
             (TypedRangeBoundaryValue::Integer(v1), TypedRangeBoundaryValue::Max) => {
                 RangeImpl::range(v1, RangeBoundaryValue::Max)
             }
-            (TypedRangeBoundaryValue::Integer(Value(v1, _)), _) => {
+            (TypedRangeBoundaryValue::Integer(Value(v1, _)), _)
+            | (_, TypedRangeBoundaryValue::Integer(Value(v1, _))) => {
```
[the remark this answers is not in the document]
desaikd: Related to removing the skip list files for `occurs` tests.

ion-schema/src/isl/isl_type.rs
desaikd: > What is the point of IslTypeKind? Why not just have name: Option<String> in either IslType or IslTypeImpl and get rid of the extra indirection?
I just tried to add wrapper (`IslType`) around what was already available previously. But It does seem that we can get rid of `IslTypeKind` in this case.

> We already have the separate internal model. What exactly is gained by having both IslType and IslTypeImpl? How does it help hide version information?
I am adding `IslTypeImpl` is using `IslConstraintImpl` to be store its constraints. Similar to `IslConstraint`, I have added `IslType` to provide methods based on versions through modules `v_1_0` and `v_2_0`.

ion-schema/src/isl/isl_type_reference.rs
```diff
@@ -11,11 +11,64 @@ use ion_rs::value::reader::{element_reader, ElementReader};
 use ion_rs::value::{IonElement, IonStruct};
 use ion_rs::IonType;
 
+/// Provides public facing APIs for constructing ISL type references programmatically for ISL 1.0
+pub mod v_1_0 {
```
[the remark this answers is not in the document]
desaikd: I don't think it is possible to provide something like `use ion-schema-rust::isl::v_1_0::*;` because this means that all the structs `IslType`, `IslConstraint` are defined within separate module `v_1_0` or `v_2_0`. This leads to 2 different implementations of `IslType` and `IslConstraint`.  I was trying not to create duplicate structures which are doing the same things.

```rust
mod v_1_0 {
  struct IslType {
     ...
  }
}

mod v_2_0 {
  struct IslType {
     ...
  }
}
```

ion-schema/src/isl/isl_constraint.rs
```diff
@@ -148,54 +209,235 @@ impl IslConstraint {
...
+    /// Creates an [IslConstraint::Type] using the [IslTypeRef] referenced inside it
+    // type is rust keyword hence this method is named type_constraint unlike other ISL constraint methods
+    pub fn type_constraint(isl_type: IslTypeRef) -> IslConstraint {
+        v_1_0::type_constraint(isl_type)
+    }
+
+    /// Creates an [IslConstraint::AllOf] using the [IslTypeRef] referenced inside it
+    pub fn all_of<A: Into<Vec<IslTypeRef>>>(isl_types: A) -> IslConstraint {
+        v_1_0::all_of(isl_types)
+    }
```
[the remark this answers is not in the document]
desaikd: It is true that we can still mix and match which is why I am checking the versions in the method that resolve an ISL model schema to a `Schema` which can be used for validation.

ion-schema/src/system.rs
```diff
@@ -518,14 +521,19 @@ impl TypeStore {
...
                 &element_reader()
                     .read_one(text.as_bytes())
                     .expect("parsing failed unexpectedly"),
                 &mut vec![],
             )
             .unwrap();
-            let type_def =
-                BuiltInTypeDefinition::parse_from_isl_type(&isl_type, self, pending_types)?;
+            let type_def = BuiltInTypeDefinition::parse_from_isl_type(
+                isl_version.to_owned(),
```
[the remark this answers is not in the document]
desaikd: No I can remove this.

ion-schema/src/system.rs
desaikd: The reason I was doing this is because I am using v_1_0 constraints methods for cosntraints that are working identically in both versions. But I see how this can get complex to check when we have more ISL versions in future. 

I will change the places where ISL constraints for ISL 2.0 are using ISL 1.0 construction methods and rather use its own which passes correct version.

```rust
   /// current approach
    pub fn all_of<A: Into<Vec<IslTypeRef>>>(isl_types: A) -> IslConstraint {
        v_1_0::all_of(isl_types)
    }
    
    // new approach
     pub fn all_of<A: Into<Vec<IslTypeRef>>>(isl_types: A) -> IslConstraint {
     // the following is similar to what v_!_0 is doign except the version detail
        IslConstraint::new(
            IonSchemaLanguageVersion::V2_0,
            IslConstraintImpl::AllOf(
                isl_types
                    .into()
                    .into_iter()
                    .map(|t| t.type_reference)
                    .collect(),
            ),
        )
    }


```

ion-schema/src/system.rs
desaikd: True. I will need to change the way constraints are constructed but I agree that this should be simpler.

ion-schema/src/system.rs
desaikd: That is correct. Happened while modularization the code. I will add some tests for this once we have some implementation of constraints from ISL 2.0. Currently those are all `todo!()`.

ion-schema/src/system.rs
desaikd: Sure,I will change this.

ion-schema/src/system.rs
desaikd: Oh yes! Thanks for catching this. I can probably add the check inside `load_isl_schema` for version marker one doesn't need to know what version it is using.

ion-schema/src/system.rs
desaikd: Yes, That is correct. This should refactored with the change in `load_isl_schema` function that has version marker check.

> Oh yes! Thanks for catching this. I can probably add the check inside `load_isl_schema` for version marker one doesn't need to know what version it is using.

ion-schema-tests-runner/src/generator.rs
desaikd: Changed with a correct implementation.

desaikd:
Updates with latest changes:
* Makes simpler check for the ISL constraint version
* `load_isl_schema` doesn't need an argument for ISL verison, isntead
checking the version marker
* Added mixed ISL constraints from different version tests
* ISL constraints `v_1_0` and `v_2_0` has its own implementation now to
pass the proper ISL version.
* change in the `Display` for `IonSchemaLanguageVersion` enum

ion-schema-tests-runner/src/generator.rs
```diff
@@ -1,10 +1,14 @@
 use crate::generator::util::*;
 use crate::model::{TestCaseDetails, TestCaseVec};
+use ion_rs::value::owned::Element;
```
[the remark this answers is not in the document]
desaikd: I will work on updating `ion-rs` in the following PR.

ion-schema/src/isl/isl_type.rs
desaikd: I have  removed `IslTypeKind` with https://github.com/amazon-ion/ion-schema-rust/pull/152/commits/24c20fb7ebeb6c68d48a9daba28b040aafe16b49

popematt (approved):

ion-schema/src/isl/isl_constraint.rs
```diff
@@ -148,54 +209,317 @@ impl IslConstraint {
...
+    pub(crate) fn new(version: IonSchemaLanguageVersion, constraint: IslConstraintImpl) -> Self {
+        Self {
+            constraint,
+            version,
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub(crate) enum IslConstraintImpl {
```
zslayton, line 477: I wish there were a more descriptive name for this, but I'm having difficulty thinking of one. Fortunately, since this is a private type we can rename it later if inspiration strikes.

zslayton (commented):
I haven't had a chance to read through this entirely, but here're some initial comments.

ion-schema/src/constraint.rs
```diff
@@ -218,132 +219,170 @@ impl Constraint {
...
+        isl_version: IonSchemaLanguageVersion,
+        type_references: &[IslTypeRefImpl],
+        type_store: &mut TypeStore,
+        pending_types: &mut PendingTypes,
+    ) -> IonSchemaResult<Vec<TypeId>> {
+        type_references
+            .iter()
+            .map(|t| {
+                IslTypeRefImpl::resolve_type_reference(
+                    isl_version.to_owned(),
```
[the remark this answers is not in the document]
desaikd: Oh yes, it does implement `Copy`. I will remove this `to_owned()` here though.

ion-schema/src/constraint.rs
```diff
@@ -218,132 +219,170 @@ impl Constraint {
...
+                )
+            })
+            .collect::<IonSchemaResult<Vec<TypeId>>>()
+    }
+
     /// Parse an [IslConstraint] to a [Constraint]
-    pub fn resolve_from_isl_constraint(
-        isl_constraint: &IslConstraint,
+    pub(crate) fn resolve_from_isl_constraint(
+        isl_version: IonSchemaLanguageVersion,
```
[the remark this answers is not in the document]
desaikd: Agree 👍🏻

PR #165, opened 2023-04-12
adds `TypeReference` implementation and `$null_or` ISL 2.0 modifier support

desaikd:
### Description of changes:
This PR works on adding `$null_or` modifier support for ISL 2.0 and adds an implementation of `TypeReference`.

### List of changes:
* add implementation of `IslTypeReferenceModifier` which stores modifier information for given type reference
* add utility methods to create `nullable_built_in_type_ref` and ``null_or_named_type_ref`.
* only allows built in types to use `nullable` modifier for ISL 1.0
* adds implementation of `TypeReference` which stores type id and modifier information for givne type reference
* all constraints now uses TypeReference instead of `TypeId`s to store type reference information
* adds document type to defintioon of `$any`
* adds `validate_for_base_nullable_type` to get the underlying ion type for a type
definition
* adds default type `$any` for ISL 2.0
* adds unit tests for `$null_or` modifier
* renames `TypeDefinition` to `TypeDefinitionKind`
* renames `TypeRef` to `TypeDefinition`
----


ion-schema/src/system.rs
```diff
@@ -181,14 +181,14 @@ impl PendingTypes {
             Some(id) => self.types_by_id[*id]
                 .to_owned()
                 .and_then(|type_def| match type_def {
-                    TypeDefinition::Named(named_type_def) => Some(Ok(named_type_def)),
-                    TypeDefinition::Anonymous(_) => {
+                    TypeDefinitionKind::Named(named_type_def) => Some(Ok(named_type_def)),
```
desaikd, line 184: All the changes in this file are related to renaming of `TypeDefinition` to `TypeDefinitionKind`.

ion-schema/src/constraint.rs
```diff
@@ -71,32 +72,58 @@ pub enum Constraint {
 impl Constraint {
     /// Creates a [Constraint::Type] referring to the type represented by the provided [TypeId].
     pub fn type_constraint(type_id: TypeId) -> Constraint {
-        Constraint::Type(TypeConstraint::new(type_id))
+        Constraint::Type(TypeConstraint::new(TypeReference::new(
```
desaikd, line 75: All the changes in this file are due to storing `TypeReference`s instead of `TypeId`s for all the constraints

ion-schema/src/system.rs
```diff
@@ -437,12 +439,12 @@ static DERIVED_ISL_TYPES: [&str; 10] = [
...
     "type::{ name: $text, one_of: [ $string, $symbol ] }",
-    "type::{ name: $any, one_of: [ $blob, $bool, $clob, $decimal,
-                                    $float, $int, $string, $symbol, $timestamp,
-                                    $list, $sexp, $struct, $null ] }",
     // this is just a place holder for document type,
     // IonSchemaElement::Document(_) type is used to verify the correctness on the validation side
     "type::{ name: document }",
+    "type::{ name: $any, one_of: [ $blob, $bool, $clob, $decimal,
+                                    $float, $int, $string, $symbol, $timestamp,
+                                    $list, $sexp, $struct, $null, document ] }",
```
desaikd, line 447: Corrects the definition for `$any` type which should also include `document`  type.

ion-schema/src/type_reference.rs
```diff
@@ -0,0 +1,72 @@
...
+use crate::isl::isl_type_reference::IslTypeRefModifier;
+use crate::result::ValidationResult;
+use crate::system::{TypeId, TypeStore};
+use crate::types::TypeValidator;
+use crate::IonSchemaElement;
+use ion_rs::IonType;
+
+/// Provides an internal representation of a schema type reference.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TypeReference {
```
desaikd, line 11: Adds a new implementation for type references which can be used to store type references in constraints. This also contains modifier related information (e.g. `nullable`, `$null_or`).

ion-schema/src/types.rs
```diff
@@ -461,13 +507,29 @@ impl TypeDefinitionImpl {
...
             };
 
-            let isl_constraint = IslConstraintImpl::from_ion_element(
-                isl_version,
-                "type",
-                &Element::symbol(Symbol::from("any")),
-                &isl_type_name,
-                &mut vec![],
-            )?;
+            let isl_constraint: IslConstraintImpl = match isl_version {
```
desaikd, line 510: adds default type based on ISL version (`$any` for ISL 2.0 and `any` for ISL 1.0)

ion-schema/src/types.rs
```diff
@@ -241,41 +242,84 @@ impl Display for BuiltInTypeDefinition {
...
+                        Some(format!("${type_name}"))
+                    } else {
+                        Some(type_name)
+                    }
+                }
+            },
+        }
+        .unwrap(); // safe unwrap() because nullable type references are only used for built in types
+
+        // unwrap() here are safe because it has been verified while constructing the schema that built type definition exist in the type store
```
popematt, line 311: This might not be safe if someone has tried to use `nullable::` on `document`, `nothing`, or anything derived from those types since there are no `$document` or `$nothing` types. Is there any way to construct a `nullable::document` or `nullable::nothing` type reference?

I took a look at the old function, and it did handle `nullable::nothing`. Is it intentional that we are getting rid of that?

ion-schema/src/isl/isl_type_reference.rs
```diff
@@ -66,53 +86,31 @@ impl IslTypeRef {
     }
 }
 
+//TODO: can add occurs here to store occurs range information for variably occurring type
```
popematt, line 89: I recommend that we keep `occurs` and nullability separate from each other. It's better to structure it like this: 
```
struct VariableOccuringTypeRef {
    occurs: SomeRange,
    type: RegularTypeRefThatCouldBeNullable
}
```
We don't want to have any confusion about the precedence between `Nullable` and `occurs`.

Here's an example to help illustrate what I mean.
The following is illegal in ISL 2.0, and for good reason. Should `$null_or` count towards the `occurs` requirement?
```
// ...
fields: {
  foo: $null_or::{ type: string, occurs: 2 }
}
// ...
```

If yes, then we should just write `{ type: $null_or::string, occurs: 2 }`. If `null` does not count towards the occurs limit, then we have a mess of surprising behavior on our hands. (Does it mean two strings plus any number of `null`? Maybe `$null_or` has the default occurs. Does it mean optional `null` _xor_ two strings? Or maybe it means we can have optional `null` _and_ two strings?)

Unless we have a very compelling performance reason, I think we're better off structuring the data in Rust in a way that helps facilitate correct understanding of the data.

ion-schema/src/type_reference.rs
```diff
@@ -0,0 +1,72 @@
+use crate::ion_path::IonPath;
+use crate::isl::isl_type_reference::IslTypeRefModifier;
+use crate::result::ValidationResult;
+use crate::system::{TypeId, TypeStore};
+use crate::types::TypeValidator;
+use crate::IonSchemaElement;
+use ion_rs::IonType;
+
+/// Provides an internal representation of a schema type reference.
```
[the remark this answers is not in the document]
desaikd: I meant to ISL type reference but I realized there is no definition of that in the spec. Does something like following sound right to you:
```suggestion
/// Provides reference to a type definition.
/// This will be used by constraints to store type references.
```

ion-schema/src/types.rs
desaikd: Programmatically only nullable types will be possible to generate based on Ion types for now using [`nullable_built_in_type_ref()`](https://github.com/amazon-ion/ion-schema-rust/pull/165/files#diff-6cf6ec7800b96bbfd4a54d935d0c1c4fd685aaa4322e866b059739d73ce32ce3R30)
Otherwise like you mentioned there is a check that happens while loading the schema to make sure nullable:document isn't constructed.
I will add `nullable:nothing` that was missed while working on this new change.

ion-schema/src/types.rs
desaikd: > Can you explain why this is necessary? `$any` is the top type, so adding `type: $any` in Ion Schema 2.0 should be a no-op.
I did not understand this. What did you mean by adding `type: $any` should be a no op?
IMO we are adding a default type constraint when there are none present in a top level named type definition. Hence I am adding `type: $any` for ISL 2.0 here.

desaikd:
Updates after last review:
* [adds suggested changes](https://github.com/amazon-ion/ion-schema-rust/pull/165/commits/ead132e58570d1b01608e56020aeb3b19aacc3da)
  * renames `IslTypeRefModifier` to `NullabilityModifier`
  * adds doc comment changes
  * changes default type addition to be only applied for ISL 1.0
  * changes nfa to use TypeReference instead of TypeId to consider
nullability
  * `$null_or` modifier should not be allowed on a type reference that has
`occurs` field in its definition

* [adds implementation of VariablyOccurringTypeRef](https://github.com/amazon-ion/ion-schema-rust/pull/165/commits/4f10fe0d0fd74ef921f95e029475edaaa09438d0)
  * removes unused method `TypeDefinitionKind::get_occurs_constraint()`
now that we have `VariablyOccurringTypeRef`
  * modifies `IslTypeRefImpl::get_occurs_constraint()` to return
`Option<Range>` instead of `bool`
  * renames `IslTypeRefImpl::get_occurs_constraint()` to
`IslTypeRefImpl::get_occurs_range()`
  * removes implementation of `OccursConstraint` as it is treated as no-op
and is handled by `VariablyOccurringTypeRef`

ion-schema/src/isl/isl_type_reference.rs
desaikd: Modified with: [adds implementation of VariablyOccurringTypeRef](https://github.com/amazon-ion/ion-schema-rust/pull/165/commits/4f10fe0d0fd74ef921f95e029475edaaa09438d0)

ion-schema/src/isl/isl_type_reference.rs
```diff
@@ -150,8 +157,8 @@ impl IslTypeRefImpl {
                     }
 
                     return match type_name {
-                        // TODO: currently it only allows for built in types to be defined with `nullable` annotation for all other type references it return an error
-                        "int" | "float" | "bool" | "decimal" | "string" | "symbol" | "blob" | "clob" | "timestamp" | "struct" | "sexp" | "list" | "text" | "lob" | "number" | "any" => {
+                        // TODO: currently it only allows for built in types to be defined with `nullable` annotation for `document` and all other type references it returns an error
```
popematt, line 160: nitpick:
```suggestion
                        // TODO: currently it only allows for built in types (other than `document`) to be defined with `nullable` annotation. For `document` and all other type references it returns an error
```

ion-schema/src/isl/isl_type_reference.rs
```diff
@@ -126,6 +131,7 @@ impl IslTypeRefImpl {
         value: &Element,
         inline_imported_types: &mut Vec<IslImportType>,
     ) -> IonSchemaResult<Self> {
+        use crate::isl::isl_type_reference::NullabilityModifier::*;
```
popematt, line 134: minor suggestion—for readability, it might be helpful to refactor something like this:
```suggestion
        use crate::isl::isl_type_reference::NullabilityModifier::*;
        let nullability = if value.has_annotation("nullable") {
            if isl_version == IslVersion::V1_0 { Nullable } else {
                return invalid_schema_error("`nullable::` modifier is not supported since ISL 2.0")
            }
        } else if value.has_annotation("$null_or") {
            if isl_version != IslVersion::V1_0 { NullOr } else {
                return invalid_schema_error("`$null_or::` modifier is not supported before ISL 2.0")
            }
        } else {
            Nothing
        }
```

This way, we can eliminate some of the duplicate code in the match arms. See comment below, for example.

ion-schema/src/isl/isl_type_reference.rs
```diff
@@ -143,29 +149,82 @@ impl IslTypeRefImpl {
...
+                // check for `$null_or` type reference
+                if value.has_annotation("$null_or") {
+                    if isl_version == IslVersion::V1_0 {
+                        return invalid_schema_error(
+                            "`$null_or::` modifier is not supported before ISL 2.0",
+                        )
+                    }
+                    isl_type_ref_modifier = NullOr
+                }
+
```
popematt, line 214: If you take my suggestion above, then I think you can replace all of this with:
```suggestion
                if nullability == Nullable {
                    // TODO: currently it only allows for built in types to be defined with `nullable` annotation for `document` and all other type references it returns an error
                    return invalid_schema_error(
                        "`nullable` modifier is only supported for built-in types",
                    )
                }
```
(Beware—committing this this "suggestion" block _might_ not work the way you expect it should because it covers some "deleted lines".)

popematt (commented):
Looking pretty good. I like how the introduction of `TypeReference` and `VariablyOccurringTypeRef` have changed the factoring of the code. I've left a few comments about the "nullability" code regarding DRY-ing it up a little bit, and a possible correctness issue (easily fixable) that I missed last time I reviewed this (sorry!).

ion-schema/src/types.rs
```diff
@@ -241,79 +241,102 @@ impl Display for BuiltInTypeDefinition {
...
 
-/// Represents a [`TypeDefinition`] which stores a resolved ISL type using [`TypeStore`]
+/// Represents a [`TypeDefinitionKind`] which stores a resolved ISL type using [`TypeStore`]
 #[derive(Debug, Clone, PartialEq)]
-pub enum TypeDefinition {
+pub enum TypeDefinitionKind {
     Named(TypeDefinitionImpl),
     Anonymous(TypeDefinitionImpl),
     BuiltIn(BuiltInTypeDefinition),
 }
```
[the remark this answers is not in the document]
desaikd: Yes,  I was thinking to do something similar because the `TypeDefinitionImpl` here is repetitive and uses `Option<String>` for naming. I will add an issue for this into the milestone.

ion-schema/src/isl/isl_type_reference.rs
```diff
@@ -143,29 +149,82 @@ impl IslTypeRefImpl {
...
+                if value.has_annotation("nullable") {
+                    if isl_version == IslVersion::V2_0 {
+                        return invalid_schema_error(
+                            "`nullable::` modifier is not supported since ISL 2.0",
+                        )
+                    }
+
+                    return match type_name {
+                        // TODO: currently it only allows for built in types to be defined with `nullable` annotation for `document` and all other type references it returns an error
+                        "int" | "float" | "bool" | "decimal" | "string" | "symbol" | "blob" | "clob" | "timestamp" | "struct" | "sexp" | "list" | "text" | "lob" | "number" | "any" | "nothing" => {
```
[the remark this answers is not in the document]
desaikd: Yes, as you mentioned both `$ION_TYPE` and `ION_TYPE` will actually be identical in terms of validation because we have an explicit check for `IonType::Null`.

ion-schema/src/types.rs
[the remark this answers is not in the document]
desaikd: https://github.com/amazon-ion/ion-schema-rust/issues/166

ion-schema/src/isl/isl_type_reference.rs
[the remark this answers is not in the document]
popematt: > Yes, as you mentioned both $ION_TYPE and ION_TYPE will actually be identical in terms of validation because we have an explicit check for IonType::Null.

Right, but it's not a question of the validation. It's a matter of whether `nullable::$null` is _legal_ in Ion Schema 1.0. If you have `nullable::$null` should that result in an invalid schema error?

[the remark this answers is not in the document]
desaikd: Yes. I meant we should support all the `$ION_TYPES` because AFAIK we haven't specified these to be illegal anywhere in the Specification which would make people expect it as legal.

popematt (approved):

PR #168, opened 2023-04-18
adds struct fields related constraints for ISL 2.0

desaikd:
### Description of changes:
This PR works on adding struct fields related constraint `field_names` and `closed::` modifier for `fields` constraint.

### List of changes: 
-  adds changes for `field_names` constraint implementation
-  adds `FieldNames` enum variants for `IslConstraintImpl` and `Constraint`
-  adds `FieldNamesConstraint` implementations
-  adds validation logic for `field_names`
-  adds changes for `closed::` modifier for `fields` constraint as per ISL 2.0
-  adds unit tests for `field_names` constraint and `closed::` modifier.

----

### DIFF
--- a/ion-schema/src/constraint.rs
+++ b/ion-schema/src/constraint.rs
@@ -15,7 +15,7 @@ use crate::types::TypeValidator;
 use crate::violation::{Violation, ViolationCode};
 use crate::IonSchemaElement;
 use ion_rs::element::Element;
-use ion_rs::{Int, IonType};
+use ion_rs::{Int, IonType, Symbol};
 use regex::{Regex, RegexBuilder};
 use std::collections::{HashMap, HashSet};
 use std::convert::TryInto;
@@ -53,6 +53,7 @@ pub enum Constraint {
     ContainerLength(ContainerLengthConstraint),
     Element(ElementConstraint),
     Exponent(ExponentConstraint),
+    FieldNames(FieldNamesConstraint),
     Fields(FieldsConstraint),
     Not(NotConstraint),
     OneOf(OneOfConstraint),
@@ -252,6 +253,22 @@ impl Constraint {
         Constraint::Fields(FieldsConstraint::new(fields, true))
     }
 
+    /// Creates a [Constraint::FieldNames] referring to the type represented by the provided [TypeId] where each field name is distinct.
+    pub fn distinct_field_names(type_id: TypeId) -> Constraint {
+        Constraint::FieldNames(FieldNamesConstraint::new(
+            TypeReference::new(type_id, NullabilityModifier::Nothing),
+            true,
+        ))
+    }
+
+    /// Creates a [Constraint::FieldNames] referring to the type represented by the provided [TypeId] where each field name.
+    pub fn field_names(type_id: TypeId) -> Constraint {
+        Constraint::FieldNames(FieldNamesConstraint::new(
+            TypeReference::new(type_id, NullabilityModifier::Nothing),
+            false,
+        ))
+    }
+
     /// Creates a [Constraint::ValidValues] using the [Element]s specified inside it
     /// Returns an IonSchemaError if any of the Elements have an annotation other than `range`
     pub fn valid_values_with_values(values: Vec<Element>) -> IonSchemaResult<Constraint> {
@@ -356,6 +373,18 @@ impl Constraint {
                 Ok(Constraint::Element(ElementConstraint::new(
                     type_id,
                     *require_distinct_elements,
+                )))
+            }
+            IslConstraintImpl::FieldNames(isl_type_reference, distinct) => {
+                let type_reference = IslTypeRefImpl::resolve_type_reference(
+                    isl_version,
+                    isl_type_reference,
+                    type_store,
+                    pending_types,
+                )?;
+                Ok(Constraint::FieldNames(FieldNamesConstraint::new(
+                    type_reference,
+                    *distinct,
                 )))
             }
             IslConstraintImpl::Fields(fields, content_closed) => {
@@ -476,6 +505,9 @@ impl Constraint {
                 container_length.validate(value, type_store, ion_path)
             }
             Constraint::Element(element) => element.validate(value, type_store, ion_path),
+            Constraint::FieldNames(field_names) => {
+                field_names.validate(value, type_store, ion_path)
+            }
             Constraint::Fields(fields) => fields.validate(value, type_store, ion_path),
             Constraint::Not(not) => not.validate(value, type_store, ion_path),
             Constraint::OneOf(one_of) => one_of.validate(value, type_store, ion_path),
@@ -1032,6 +1064,117 @@ impl ConstraintValidator for FieldsConstraint {
             ));
         }
         Ok(())
+    }
+}
+
+/// Implements an `field_names` constraint of Ion Schema
+/// [field_names]: https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#field_names
+#[derive(Debug, Clone, PartialEq)]
+pub struct FieldNamesConstraint {
+    type_reference: TypeReference,
+    requires_distinct: bool,
+}
+
+impl FieldNamesConstraint {
+    pub fn new(type_reference: TypeReference, requires_distinct: bool) -> Self {
+        Self {
+            type_reference,
+            requires_distinct,
+        }
+    }
+}
+
+impl ConstraintValidator for FieldNamesConstraint {
+    fn validate(
+        &self,
+        value: &IonSchemaElement,
+        type_store: &TypeStore,
+        ion_path: &mut IonPath,
+    ) -> ValidationResult {
+        let mut violations: Vec<Violation> = vec![];
+
+        // create a set for checking duplicate elements
+        let mut element_set = vec![];
+
+        // get elements for given container in the form (ion_path_element, element_value)
+        let elements: Vec<(IonPathElement, &Element)> = match value {
+            IonSchemaElement::SingleElement(element) => {
+                // Check for null struct
+                if element.is_null() {
+                    return Err(Violation::new(
+                        "field_names",
+                        ViolationCode::TypeMismatched,
+                        "expected a struct but found null",
+                        ion_path,
+                    ));
+                }
+
+                return match element.as_struct() {
+                    Some(struct_value) => {
+                        let field_names: Vec<(IonPathElement, &Symbol)> = struct_value
+                            .iter()
+                            .map(|(field_name, val)| {
+                                (
+                                    IonPathElement::Field(field_name.text().unwrap().to_owned()),
+                                    field_name,
+                                )
+                            })
+                            .collect();
+                        // validate field_names constraint
+                        for (ion_path_element, field_name) in field_names {
+                            ion_path.push(ion_path_element);
+                            let schema_element: IonSchemaElement =
+                                (&Element::symbol(field_name)).into();
+
+                            if let Err(violation) =
+                                self.type_reference
+                                    .validate(&schema_element, type_store, ion_path)
+                            {
+                                violations.push(violation);
+                            }
+                            if self.requires_distinct && element_set.contains(&field_name) {
+                                violations.push(Violation::new(
+                                    "field_names",
+                                    ViolationCode::ElementNotDistinct,
+                                    format!("expected distinct field names but found duplicate field name {field_name}", ),
+                                    ion_path,
+                                ))
+                            }
+                            element_set.push(field_name);
+                            ion_path.pop();
+                        }
+                        if !violations.is_empty() {
+                            return Err(Violation::with_violations(
+                                "field_names",
+                                ViolationCode::ElementMismatched,
+                                "one or more field names don't satisfy field_names constraint",
+                                ion_path,
+                                violations,
+                            ));
+                        }
+                        Ok(())
+                    }
+                    None => {
+                        // return Violation if value is not a struct
+                        Err(Violation::new(
+                            "field_names",
+                            ViolationCode::TypeMismatched,
+                            format!("expected a struct but found {}", element.ion_type()),
+                            ion_path,
+                        ))
+                    }
+                };
+            }
+            IonSchemaElement::Document(_) => {
+                // return Violation if value is not a struct
+                return Err(Violation::new(
+                    "field_names",
+                    ViolationCode::TypeMismatched,
+                    "expected a struct but found document",
+                    ion_path,
+                ));
+            }
+        };
     }
 }
 
--- a/ion-schema/src/constraint.rs
+++ b/ion-schema/src/constraint.rs
@@ -358,7 +358,11 @@ impl Constraint {
                     *require_distinct_elements,
                 )))
             }
-            IslConstraintImpl::Fields(fields) => {
+            IslConstraintImpl::Fields(fields, content_closed) => {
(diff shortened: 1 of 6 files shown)


ion-schema/src/constraint.rs
```diff
@@ -1031,6 +1067,117 @@ impl ConstraintValidator for FieldsConstraint {
...
+                        "expected a struct but found null",
+                        ion_path,
+                    ));
+                }
+
+                return match element.as_struct() {
+                    Some(struct_value) => {
+                        let field_names: Vec<(IonPathElement, &Symbol)> = struct_value
+                            .iter()
+                            .map(|(field_name, val)| {
```
popematt, line 1116:
```suggestion
                            .map(|(field_name, _)| {
```

ion-schema/src/constraint.rs
```diff
@@ -1031,6 +1067,117 @@ impl ConstraintValidator for FieldsConstraint {
...
+                        "field_names",
+                        ViolationCode::TypeMismatched,
+                        "expected a struct but found null",
+                        ion_path,
+                    ));
+                }
+
+                return match element.as_struct() {
+                    Some(struct_value) => {
+                        let field_names: Vec<(IonPathElement, &Symbol)> = struct_value
```
popematt, line 1114: Given that the IonPathElement is the text of the &Symbol, is it necessary for this to be a `Vec<(IonPathElement, &Symbol)>`? My _opinion_ is that it would be easier to read/follow the code if we just stored `Vec<String>` and then constructed the `IonPathElement` in the loop that actually checks the field names.

ion-schema/src/constraint.rs
```diff
@@ -1031,6 +1067,117 @@ impl ConstraintValidator for FieldsConstraint {
...
+                            .iter()
+                            .map(|(field_name, val)| {
+                                (
+                                    IonPathElement::Field(field_name.text().unwrap().to_owned()),
+                                    field_name,
+                                )
+                            })
+                            .collect();
+                        // validate field_names constraint
```
popematt, line 1123: minor—not a useful comment; please remove.

ion-schema/src/constraint.rs
```diff
@@ -1031,6 +1067,117 @@ impl ConstraintValidator for FieldsConstraint {
...
+    fn validate(
+        &self,
+        value: &IonSchemaElement,
+        type_store: &TypeStore,
+        ion_path: &mut IonPath,
+    ) -> ValidationResult {
+        let mut violations: Vec<Violation> = vec![];
+
+        // create a set for checking duplicate elements
+        let mut element_set = vec![];
```
popematt, line 1097: Both `Vec` and `HashSet` have amortized `O(1)` time for inserting an element (assuming at the end, in the case of `Vec`). However, `Vec` has `O(n)` time complexity for checking if a single value is present in an unsorted list, so checking for duplicates is `O(n^2)`, but the `HashSet`, has `O(1)` time complexity to check for membership in the set, so checking for duplicates is overall (amortized) `O(n)`.

ion-schema/src/constraint.rs
```diff
@@ -1031,6 +1067,117 @@ impl ConstraintValidator for FieldsConstraint {
...
+                            }
+                            if self.requires_distinct && element_set.contains(&field_name) {
+                                violations.push(Violation::new(
+                                    "field_names",
+                                    ViolationCode::ElementNotDistinct,
+                                    format!("expected distinct field names but found duplicate field name {field_name}", ),
+                                    ion_path,
+                                ))
+                            }
+                            element_set.push(field_name);
```
popematt, line 1143: We don't need to add things to `element_set` unless we're checking for duplicates, and we don't need to add a field name to `element_set` if we've already determined that it's a duplicate.

```suggestion
                            if self.requires_distinct {
                                if element_set.contains(&field_name) {
                                    violations.push(Violation::new(
                                        "field_names",
                                        ViolationCode::ElementNotDistinct,
                                        format!("expected distinct field names but found duplicate field name {field_name}", ),
                                        ion_path,
                                    ))
                                } else {
                                    element_set.push(field_name);
                                }
                            }
```
This is just one possible way of dealing with the things I mentioned above. It _might_ be more readable to invert the inner `if` to `if not in set { add to set } else { add violation }`, or even factor it a different way. Your call.

ion-schema/src/constraint.rs
```diff
@@ -1031,6 +1067,117 @@ impl ConstraintValidator for FieldsConstraint {
...
+                // Check for null struct
+                if element.is_null() {
+                    return Err(Violation::new(
+                        "field_names",
+                        ViolationCode::TypeMismatched,
+                        "expected a struct but found null",
+                        ion_path,
+                    ));
+                }
+
```
popematt, line 1111: We can reduce the levels of nesting (a readability improvement, IMO) by using early returns in an `if let ... else` block. E.g.
```rust
let element = if let SingleElement(e) = value { e } else {
    return Err(Violation::new(
        "field_names",
        ViolationCode::TypeMismatched,
        format!("expected a struct but found document"),
        ion_path,
    ))
}

let the_struct = if let Some(s) = element.as_struct() { s } else {
    return Err(Violation::new(
        "field_names",
        ViolationCode::TypeMismatched,
        format!("expected a struct but found {}", element.ion_type()),
        ion_path,
    ))
}
```

ion-schema/src/constraint.rs
```diff
@@ -1031,6 +1067,117 @@ impl ConstraintValidator for FieldsConstraint {
...
+                            .iter()
+                            .map(|(field_name, val)| {
+                                (
+                                    IonPathElement::Field(field_name.text().unwrap().to_owned()),
+                                    field_name,
+                                )
+                            })
+                            .collect();
+                        // validate field_names constraint
+                        for (ion_path_element, field_name) in field_names {
```
popematt, line 1124: Alternately, we could skip the `field_names` variable altogether and use:
```rust
for (field_name, _) in struct_value {
    // ...
}
```

ion-schema/src/isl/isl_constraint.rs
```diff
@@ -490,7 +498,14 @@ pub(crate) enum IslConstraintImpl {
     // For ISL 1.0 which doesn't support `distinct` elements this will be (type_reference, false).
     Element(IslTypeRefImpl, bool),
     Exponent(Range),
-    Fields(HashMap<String, IslTypeRefImpl>),
+    // Represents Fields(fields, content_closed)
+    // For ISL 2.0 true/false is specified based on whether `closed::` annotation is present or not
+    // For ISL 1.0 this will always be (fields, false) as it doesn't support `closed::` annotation on fields constraint
+    Fields(HashMap<String, IslTypeRefImpl>, bool),
```
popematt, line 504: Does not need to be addressed in this PR. However, do you think we should change this to `Fields(Fields)` and add a `Fields { fields: HashMap<...>, is_closed: bool }` struct? (And similarly for the other constraints?)

@zslayton, do you have an opinion too?

ion-schema/src/constraint.rs
desaikd: I am using `Vec` as `Hash` implementation is not available for `Element`.

popematt: But `Hash` is available for `String`, and we can convert the field names to `String`.

desaikd: Thats true :)

ion-schema/src/isl/isl_constraint.rs
desaikd: We can also do something like following if the intention is to have a name suggesting what the enum vvariant arguments are:
```rust
enum IslConstraintImpl {
   ...
   Fields({ field: HashMap<...>, is_closed: bool })
   ...
}
```

popematt: If we're doing that, we may as well get rid of the tuple.
```rust
enum IslConstraintImpl {
   ...
   Fields { field: HashMap<...>, is_closed: bool }
   ...
}
```

What I'm really getting at, I suppose, is whether the enum variants should hold an opaque type.

ion-schema/src/constraint.rs
```diff
@@ -252,6 +253,22 @@ impl Constraint {
...
+        ))
+    }
+
+    /// Creates a [Constraint::FieldNames] referring to the type represented by the provided [TypeId] where each field name.
+    pub fn field_names(type_id: TypeId) -> Constraint {
+        Constraint::FieldNames(FieldNamesConstraint::new(
+            TypeReference::new(type_id, NullabilityModifier::Nothing),
+            false,
+        ))
+    }
```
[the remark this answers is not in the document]
desaikd: No particular reason since these methods are all part of a module which is not public we can choose either way so probably a single method will be better.

ion-schema/src/isl/isl_constraint.rs
desaikd: Sorry. I didn't mean to include the tuple in my example. I understand what you are suggesting here but I don't see it to be necessary since this enum is crate level enum now. I will create an issue to keep this discussion open though.

ion-schema/src/isl/isl_constraint.rs
```diff
@@ -888,12 +961,22 @@ impl IslConstraintImpl {
...
+                    .as_struct()
+                    .unwrap()
+                    .get_all(f.text().unwrap())
+                    .count()
+                    > 1
+                {
+                    return invalid_schema_error(
+                        "fields must be a struct with no repeated field names",
+                    );
+                }
```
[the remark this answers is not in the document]
desaikd: Used map length to detect duplicates.

ion-schema/src/constraint.rs
desaikd: Using match arms with condition to improve readability.

popematt: The match arms with condition eliminates one layer of the nesting/indentation, but it's not quite what I was suggesting.

I had forgotten that there is a `expect_element_of_type` function, but it is more along the lines of what I would like, and in fact, it's even better than what I originally suggested. You should do here what you did with the fields constraint validator. I.e.:
```rust
let ion_struct = value
    .expect_element_of_type(&[IonType::Struct], "field_names")?
    .as_struct()
    .unwrap();
```

ion-schema/src/isl/isl_constraint.rs
popematt: Oh, I thought that since this was in `isl_constraint.rs` that this was the public version of the model. Is it not?

If it's not a public enum, then I have no objection to keeping it the way it is.

ion-schema/src/isl/isl_constraint.rs
```diff
@@ -645,6 +660,42 @@ impl IslConstraintImpl {
...
+                            value.to_owned(),
+                        ))
+                    }
+                    IslVersion::V2_0 => {
+                        // return error if there are any annotations other than `distinct`
+                        if value.annotations().any(|a| a.text() != Some("distinct")) {
+                            return Err(invalid_schema_error_raw(
+                                "field_names constraint can only contain `distinct` annotation",
+                            ));
+                        }
```
popematt, line 680: Suggestion—make this code like the code for `fields`. It ensures that there's at most 1 occurrence of "distinct".
```suggestion
                        if value.annotations().count() > 1
                            || value.annotations().any(|a| a.text() != Some("distinct"))
                        {
                            return invalid_schema_error(
                                "field_names constraint may only be annotated with 'distinct'",
                            );
                        }
```

ion-schema/src/isl/isl_constraint.rs
```diff
@@ -645,6 +660,42 @@ impl IslConstraintImpl {
...
+
+                        // verify whether `distinct`annotation is present or not
+                        let require_distinct = value.has_annotation("distinct");
+
```
popematt, line 684:
```suggestion
```

ion-schema/src/isl/isl_constraint.rs
```diff
@@ -645,6 +660,42 @@ impl IslConstraintImpl {
...
+                        // return error if the type reference contains `occurs` constraint
+                        if type_reference.get_occurs_range().is_some() {
+                            return Err(invalid_schema_error_raw(
+                                "field_names constraint can not contain type references that contain `occurs` constraint",
+                            ));
+                        }
+
+                        Ok(IslConstraintImpl::FieldNames(
+                            type_reference,
+                            require_distinct,
```
popematt, line 694:
```suggestion
                            value.has_annotation("distinct"),
```

ion-schema/src/isl/isl_constraint.rs
```diff
@@ -658,7 +709,22 @@ impl IslConstraintImpl {
...
-                Ok(IslConstraintImpl::Fields(fields))
+                match isl_version {
+                    IslVersion::V1_0 => Ok(IslConstraintImpl::Fields(fields, false)),
+                    IslVersion::V2_0 => {
+                        if value.annotations().count() > 1
+                            || value.annotations().any(|a| a.text() != Some("closed"))
+                        {
+                            return Err(invalid_schema_error_raw(
+                                "fields constraint may only be annotated with 'closed'",
+                            ));
```
popematt, line 720: Slightly more concise:
```suggestion
                            return invalid_schema_error("fields constraint may only be annotated with 'closed'");
```

ion-schema/src/isl/isl_constraint.rs
```diff
@@ -645,6 +660,42 @@ impl IslConstraintImpl {
...
+                        }
+
+                        // verify whether `distinct`annotation is present or not
+                        let require_distinct = value.has_annotation("distinct");
+
+                        // return error if the type reference contains `occurs` constraint
+                        if type_reference.get_occurs_range().is_some() {
+                            return Err(invalid_schema_error_raw(
+                                "field_names constraint can not contain type references that contain `occurs` constraint",
+                            ));
```
popematt, line 689:
```suggestion
                            return invalid_schema_error(
                                "field_names constraint can not contain type references that contain `occurs` constraint",
                            );
```

ion-schema/src/isl/isl_constraint.rs
desaikd: Yes, its in `isl_constraint` but now I have added another struct `IslConstraint` which is the on visible publicly and this enum is just used by `IslCosntraint`.

ion-schema/src/isl/isl_constraint.rs
```diff
@@ -544,16 +544,16 @@ impl IslConstraintImpl {
             }
             "annotations" => {
                 if value.is_null() {
-                    return Err(invalid_schema_error_raw(
+                    return invalid_schema_error(
```
popematt, line 547: FYI—It's fine to save related clean-up changes like this for a follow up PR.

ion-schema/src/constraint.rs
```diff
@@ -1079,66 +1079,44 @@ impl ConstraintValidator for FieldNamesConstraint {
...
-            IonSchemaElement::Document(_) => Err(Violation::new(
+            ion_path.pop();
+        }
+
+        if !violations.is_empty() {
+            return Err(Violation::with_violations(
                 "field_names",
-                ViolationCode::TypeMismatched,
-                "expected a struct but found document",
+                ViolationCode::ElementMismatched,
```
popematt, line 1113: Doesn't need to be this PR, but we should evaluate whether `ElementMismatched` and `ElementNotDistinct` are suitable to re-use for the `field_names` constraint.

popematt (approved):

PR #174, opened 2023-05-03
adds implementation of `ieee754_float` constraint

desaikd:
### Description of changes:
This PR works on adding struct fields related constraint field_names and closed:: modifier for fields constraint.

### List of changes:

-  adds changes for `ieee754_float` constraint implementation
-  adds `Ieee754Float` enum variants for `IslConstraintImpl` and `Constraint`
-  adds Ieee754FloatConstraint implementations
-  adds validation logic for `ieee754_float`
- adds dependency to [`half`](https://crates.io/crates/half/) crate for binary16 conversion 
-  adds unit tests for `ieee754_float`

----

### DIFF
--- a/ion-schema/Cargo.toml
+++ b/ion-schema/Cargo.toml
@@ -28,6 +28,7 @@ num-bigint = "0.3"
 num-traits = "0.2"
 chrono = "0.4"
 regex = "1.5.6"
+half = "2.2.1"
 
 [dev-dependencies]
 rstest = "0.9"
--- a/ion-schema/src/constraint.rs
+++ b/ion-schema/src/constraint.rs
@@ -2345,38 +2345,29 @@ impl ConstraintValidator for Ieee754FloatConstraint {
             return Ok(());
         }
 
-        match self.interchange_format {
+        let is_valid = match self.interchange_format {
             Ieee754InterchangeFormat::Binary16 => {
-                let f16_value = half::f16::from_f64(float_value);
-                if f16_value.to_f64() != float_value {
-                    return Err(Violation::new(
-                        "ieee754_float",
-                        ViolationCode::InvalidIeee754Float,
-                        format!("value cannot be losslessly represented by the IEEE-754 {} interchange format.", self.interchange_format),
-                        ion_path,
-                    ));
-                }
-                Ok(())
+                half::f16::from_f64(float_value).to_f64() == float_value
             }
-            Ieee754InterchangeFormat::Binary32 => {
-                let f64_value = float_value.to_f32().and_then(|f32_value| f32_value.to_f64()).ok_or(Violation::new(
-                    "ieee754_float",
-                    ViolationCode::InvalidIeee754Float,
-                    format!("value cannot be losslessly represented by the IEEE-754 {} interchange format.", self.interchange_format),
-                    ion_path,
-                ))?;
+            Ieee754InterchangeFormat::Binary32 => float_value
+                .to_f32()
+                .and_then(|f32_value| f32_value.to_f64().map(|f64_value| f64_value == float_value))
+                .unwrap_or(false),
+            Ieee754InterchangeFormat::Binary64 => true,
+        };
 
-                if f64_value != float_value {
-                    return Err(Violation::new(
-                        "ieee754_float",
-                        ViolationCode::InvalidIeee754Float,
-                        format!("value cannot be losslessly represented by the IEEE-754 {} interchange format.", self.interchange_format),
-                        ion_path,
-                    ));
-                }
-                Ok(())
-            }
-            Ieee754InterchangeFormat::Binary64 => Ok(()),
+        if is_valid {
+            Ok(())
+        } else {
+            Err(Violation::new(
+                "ieee754_float",
+                ViolationCode::InvalidIeee754Float,
+                format!(
+                    "value cannot be losslessly represented by the IEEE-754 {} interchange format.",
+                    self.interchange_format
+                ),
+                ion_path,
+            ))
         }
     }
 }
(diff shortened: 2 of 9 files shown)


ion-schema/Cargo.toml
```diff
@@ -28,6 +28,7 @@ num-bigint = "0.3"
 num-traits = "0.2"
 chrono = "0.4"
 regex = "1.5.6"
+half = "2.2.1"
```
popematt, line 31: Nice! I'm glad we didn't have to have some hokey thing like in `ion-schema-kotlin`.

ion-schema/src/constraint.rs
```diff
@@ -2295,3 +2310,73 @@ impl ConstraintValidator for TimestampOffsetConstraint {
...
+                        "ieee754_float",
+                        ViolationCode::InvalidIeee754Float,
+                        format!("value cannot be losslessly represented by the IEEE-754 {} interchange format.", self.interchange_format),
+                        ion_path,
+                    ));
+                }
+                Ok(())
+            }
+            Ieee754InterchangeFormat::Binary64 => Ok(()),
+        }
```
popematt, line 2380: You can simplify and get rid of repeated code by moving the violations out of the `match` expression. (I haven't checked this example to make sure it's syntactically correct, but this is the general idea.)

```suggestion
        let is_valid = match self.interchange_format {
            Ieee754InterchangeFormat::Binary16 => {
                half::f16::from_f64(float_value).to_f64() == float_value
            }
            Ieee754InterchangeFormat::Binary32 => float_value.to_f32()
                .and_then(|f32_value| f32_value.to_f64() == float_value)
                .ok_or(false),
            Ieee754InterchangeFormat::Binary64 => true,
        }
        if is_valid {
            Ok(())
        } else {
            Err(Violation::new(
                "ieee754_float",
                ViolationCode::InvalidIeee754Float,
                format!("value cannot be losslessly represented by the IEEE-754 {} interchange format.", self.interchange_format),
                ion_path,
            ));
        }
```

ion-schema/src/isl/isl_constraint.rs
```diff
@@ -730,6 +741,29 @@ impl IslConstraintImpl {
...
             }
+            "ieee754_float" => {
+                if value.is_null() {
+                    return invalid_schema_error(
+                        "expected a symbol for an `ieee754_float` constraint, found null",
+                    );
+                }
+                if !value.annotations().is_empty() {
+                    return invalid_schema_error(
+                        "`ieee_754_float` constraint must be one of binary16, binary32, binary64",
```
popematt, line 752: The message is a little vague for the condition that's being checked; maybe something like this?
```suggestion
                        "`ieee_754_float` argument must not have annotations ",
```

ion-schema/src/isl/isl_constraint.rs
```diff
@@ -730,6 +741,29 @@ impl IslConstraintImpl {
                     }
                 }
             }
+            "ieee754_float" => {
+                if value.is_null() {
+                    return invalid_schema_error(
+                        "expected a symbol for an `ieee754_float` constraint, found null",
+                    );
+                }
```
popematt, line 749: This is technically unnecessary because of the `as_symbol()` on L757, right? If I'm reading it correctly, it looks like the reason it's here is to provide a null-specific error message, but we could change it to eliminate this check.

ion-schema/src/isl/isl_constraint.rs
```diff
@@ -730,6 +741,29 @@ impl IslConstraintImpl {
...
+                let string_value =
+                    value
+                        .as_symbol()
+                        .map(|s| s.text().unwrap())
+                        .ok_or_else(|| {
+                            invalid_schema_error_raw(format!(
+                                "expected ieee754_float to contain a symbol but found: {}",
+                                value.ion_type()
```
popematt, line 762: suggestion—more more actionable error message, and also applicable for `null.symbol`.
```suggestion
                                "expected ieee754_float to be one of 'binary16', 'binary32', or 'binary64', but it was: {value}",
```

ion-schema/src/isl/util.rs
```diff
@@ -272,3 +272,42 @@ impl Display for TimestampOffset {
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "{}",
+            match self {
+                Ieee754InterchangeFormat::Binary16 => "binary16",
+                Ieee754InterchangeFormat::Binary32 => "binary32",
+                Ieee754InterchangeFormat::Binary64 => "binary64",
+            }
+        )
```
popematt, line 311: This is my personal preference because I don't like nesting things more than I have to. Take it or leave it.
```suggestion
        let text = match self {
            Ieee754InterchangeFormat::Binary16 => "binary16",
            Ieee754InterchangeFormat::Binary32 => "binary32",
            Ieee754InterchangeFormat::Binary64 => "binary64",
        };
        write!(f, "{text}")
```

ion-schema/src/isl/util.rs
```diff
@@ -272,3 +272,42 @@ impl Display for TimestampOffset {
...
+            "binary16" => Binary16,
+            "binary32" => Binary32,
+            "binary64" => Binary64,
+            _ => {
+                return invalid_schema_error(format!(
+                    "unrecognized `ieee754_float` value {}",
+                    &string_value
+                ))
+            }
+        })
```
popematt, line 297: Returning an Err from inside the `Ok(...)` seems unnecessarily complicated to me. I think it's cleaner and clearer to write this:
```suggestion
        match string_value {
            "binary16" => Ok(Binary16),
            "binary32" => Ok(Binary32),
            "binary64" => Ok(Binary64),
            _ => invalid_schema_error(format!("unrecognized `ieee754_float` value {string_value}"))
        }
```

popematt (changes requested):

ion-schema/src/isl/isl_constraint.rs
desaikd: Do you mind if I do this in a separate PR, there will be many places where I need to change this.

popematt: No need to increase the scope of this PR. Just fix this one instance that is part of the PR and create an issue to clean up other instances.

desaikd: Sure. Will do. Thanks!

popematt (approved):

PR #175, opened 2023-05-10
adds implementation of annotations constraint ISL 2.0

desaikd:
### Description of changes:
This PR adds implementation of annotations constraint ISL 2.0.

### Grammar:
```
<ANNOTATIONS> ::= annotations: <ANNOTATIONS_MODIFIER>... [ <SYMBOL>... ]
                | annotations: <TYPE_ARGUMENT>

<ANNOTATIONS_MODIFIER> ::= required::
                         | closed::

```

### Ion schema specification:
https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#annotations

### List of changes:
* adds support for standard and simple syntax for annotations constraint as per ISL 2.0
* adds enum `IslAnnotationsConstraint` which holds two variants for
standard and simple syntax of annotations constraint
* adds `Anotations2_0Constraint` to handle ISL 2.0 related annotations
validation based on a type reference
* adds unit tests for `annotations` constraint ISL 2.0
----

### DIFF
--- a/ion-schema-tests-runner/tests/ion-schema-tests-2-0.rs
+++ b/ion-schema-tests-runner/tests/ion-schema-tests-2-0.rs
@@ -2,14 +2,12 @@ use ion_schema_tests_runner::ion_schema_tests;
 
 ion_schema_tests!(
     root = "ion-schema-tests/ion_schema_2_0/",
-    // Support for ISL 2.0 is not implemented yet, so all tests are ignored.
+    // Support for ISL 2.0 is not implemented yet, so some tests are ignored.
     ignored(
         "open_content",
         "imports",
         "schema::*",
         "constraints::all_of",
-        "constraints::annotations_simplified",
-        "constraints::annotations_standard",
         "constraints::any_of",
         "constraints::byte_length",
         "constraints::codepoint_length",
--- a/ion-schema-tests-runner/tests/ion-schema-tests-2-0.rs
+++ b/ion-schema-tests-runner/tests/ion-schema-tests-2-0.rs
@@ -2,7 +2,7 @@ use ion_schema_tests_runner::ion_schema_tests;
 
 ion_schema_tests!(
     root = "ion-schema-tests/ion_schema_2_0/",
-    // Support for ISL 2.0 is not implemented yet, so some tests are ignored.
+    // Support for ISL 2.0 is not completely implemented yet, so some tests are ignored.
     ignored(
         "open_content",
         "imports",
--- a/ion-schema/src/constraint.rs
+++ b/ion-schema/src/constraint.rs
@@ -1567,35 +1567,26 @@ impl ConstraintValidator for AnnotationsConstraint2_0 {
         type_store: &TypeStore,
         ion_path: &mut IonPath,
     ) -> ValidationResult {
-        let mut violations: Vec<Violation> = vec![];
-
         match value {
             IonSchemaElement::SingleElement(element) => {
-                let mut sequence_builder = Element::sequence_builder();
-                for annotation in element.annotations().iter() {
-                    sequence_builder =
-                        sequence_builder.push(Element::symbol(annotation.to_owned()));
-                }
+                let schema_element: IonSchemaElement = (&element
+                    .annotations()
+                    .iter()
+                    .map(Element::symbol)
+                    .collect::<Vec<_>>())
+                    .into();
 
-                let schema_element: IonSchemaElement =
-                    (&Element::from(sequence_builder.build_list())).into();
-                if let Err(violation) =
-                    self.type_ref
-                        .validate(&schema_element, type_store, ion_path)
-                {
-                    violations.push(violation);
-                }
-
-                if !violations.is_empty() {
-                    return Err(Violation::with_violations(
-                        "annotations",
-                        ViolationCode::AnnotationMismatched,
-                        "one or more annotations don't satisfy annotations constraint",
-                        ion_path,
-                        violations,
-                    ));
-                }
-                Ok(())
+                self.type_ref
+                    .validate(&schema_element, type_store, ion_path)
+                    .map_err(|v| {
+                        Violation::with_violations(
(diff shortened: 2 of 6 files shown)


ion-schema/src/constraint.rs
```diff
@@ -1504,6 +1545,71 @@ impl ConstraintValidator for ElementConstraint {
...
+        match value {
+            IonSchemaElement::SingleElement(element) => {
+                let mut sequence_builder = Element::sequence_builder();
+                for annotation in element.annotations().iter() {
+                    sequence_builder =
+                        sequence_builder.push(Element::symbol(annotation.to_owned()));
+                }
+
+                let schema_element: IonSchemaElement =
+                    (&Element::from(sequence_builder.build_list())).into();
```
popematt, line 1581: I think it might work to do this.
```suggestion
                let schema_element: IonSchemaElement = 
                    element.annotations().map(Ion::symbol).collect::<Vec<_>>().into();
```

ion-schema/src/constraint.rs
```diff
@@ -1504,6 +1545,71 @@ impl ConstraintValidator for ElementConstraint {
...
+
+                if !violations.is_empty() {
+                    return Err(Violation::with_violations(
+                        "annotations",
+                        ViolationCode::AnnotationMismatched,
+                        "one or more annotations don't satisfy annotations constraint",
+                        ion_path,
+                        violations,
+                    ));
+                }
```
popematt, line 1597: This seems redundant to check for an `Err(violation)` and then check if `violations` is empty. Can it be simplified?

ion-schema/src/isl/isl_constraint.rs
```diff
@@ -433,12 +439,37 @@ pub mod v_2_0 {
         )
     }
 
-    /// Creates an `annotations` constraint using [str]s and [Element]s specified inside it
-    pub fn annotations<'a, A: IntoIterator<Item = &'a str>, B: IntoIterator<Item = Element>>(
-        annotations_modifiers: A,
-        annotations: B,
+    /// Creates an `annotations` constraint using a list of valid annotations and specify whether annotations are required or closed or both.
+    pub fn simple_annotations<A: IntoIterator<Item = Element>>(
```
popematt, line 443: minor naming suggestion—start with `annotations`.
```suggestion
    pub fn annotations_simplified<A: IntoIterator<Item = Element>>(
```

ion-schema/src/isl/isl_constraint.rs
```diff
@@ -433,12 +439,37 @@ pub mod v_2_0 {
...
+        IslConstraint::new(
+            IslVersion::V2_0,
+            IslConstraintImpl::Annotations(IslAnnotationsConstraint::SimpleAnnotations(
+                IslSimpleAnnotationsConstraint::new(is_closed, false, is_required, annotations),
+            )),
+        )
+    }
+
+    /// Creates an `annotations` constraint using an [IslTypeRef].
+    pub fn standard_annotations(isl_type: IslTypeRef) -> IslConstraint {
```
popematt, line 466: minor naming suggestion—if it's "standard", then we probably don't need to qualify the name.
```suggestion
    pub fn annotations(isl_type: IslTypeRef) -> IslConstraint {
```

ion-schema/src/isl/isl_constraint.rs
```diff
@@ -433,12 +439,37 @@ pub mod v_2_0 {
...
     }
 
-    /// Creates an `annotations` constraint using [str]s and [Element]s specified inside it
-    pub fn annotations<'a, A: IntoIterator<Item = &'a str>, B: IntoIterator<Item = Element>>(
-        annotations_modifiers: A,
-        annotations: B,
+    /// Creates an `annotations` constraint using a list of valid annotations and specify whether annotations are required or closed or both.
+    pub fn simple_annotations<A: IntoIterator<Item = Element>>(
+        is_required: bool,
+        is_closed: bool,
```
popematt, line 445: question (non-blocking)—Is it possible to model this a different way so that it won't compile unless one of these is true?

ion-schema-tests-runner/tests/ion-schema-tests-2-0.rs
```diff
@@ -2,14 +2,12 @@ use ion_schema_tests_runner::ion_schema_tests;
 
 ion_schema_tests!(
     root = "ion-schema-tests/ion_schema_2_0/",
-    // Support for ISL 2.0 is not implemented yet, so all tests are ignored.
+    // Support for ISL 2.0 is not implemented yet, so some tests are ignored.
```
popematt, line 5: If we're being pedantic... 😉 
```suggestion
    // Support for ISL 2.0 is not completely implemented yet, so some tests are ignored.
```

ion-schema/src/isl/isl_constraint.rs
desaikd: How about adding a note in the doc comment that by default required would be set to `true` if both of them are passed as `false`? or set function argument to be an enum representing `Required`, `Closed` and `RequiredAndClosed`?

popematt: > How about adding a note in the doc comment that by default required would be set to true if both of them are passed as false? 

This is not a great idea. It's better to return an `Err` (or even panic) than it is to change an explicitly provided argument.

> or set function argument to be an enum representing Required, Closed and RequiredAndClosed?

Something like this could work. Another option might be to have overloaded functions for `annotations_closed`, `annotations_closed_required`, and `annotations_required`.

I think for consumers of the library, an enum is probably the best choice. For internal use (i.e. for reading Ion Schema), it might be more useful to keep this signature but return `Err` if there's a problem.

desaikd: I just checked that for `valid_values`  we return an error when there are annotations other than `range`. I think probably something similar would make sense for here.

popematt (approved):

PR #177, opened 2023-05-16
adds changes for timestamp ranges ISL 2.0

desaikd:
### Description of changes:
This PR adds changes for timestamp ranges as per ISL 2.0.

### List of changes:
* adds changes to allow unknown offsets on timestamp ranges for ISL 2.0
* adds changes to pass isl_version for `ValidValues` constraint
* adds changes to check for `exclusive` annotation on min/max range boundary value
* adds unit tests

----

### DIFF
--- a/ion-schema/src/isl/isl_range.rs
+++ b/ion-schema/src/isl/isl_range.rs
@@ -1,5 +1,6 @@
 use crate::isl::isl_range::RangeBoundaryValue::*;
 use crate::isl::util::TimestampPrecision;
+use crate::isl::IslVersion;
 use crate::result::{
     invalid_schema_error, invalid_schema_error_raw, IonSchemaError, IonSchemaResult,
 };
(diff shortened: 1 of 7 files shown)


ion-schema/src/isl/mod.rs
```diff
@@ -577,6 +577,16 @@ mod isl_tests {
...
+            IslVersion::V1_0,
+        )
+    }
+
+    // helper function to create a range
+    fn load_range_v2_0(text: &str) -> IonSchemaResult<Range> {
+        Range::from_ion_element(
+            &Element::read_one(text.as_bytes()).expect("parsing failed unexpectedly"),
+            RangeType::Any,
+            IslVersion::V2_0,
```
popematt, line 589: The `load_range_...` functions appear to be private functions, so you could combine these functions and just make the ISL version an argument to the function.

ion-schema/src/isl/isl_range.rs
```diff
@@ -848,12 +838,23 @@ impl TypedRangeBoundaryValue {
...
+                RangeType::NumberOrTimestamp | RangeType::Any => {
+                    // For ISL 1.0, verify that range boundary here doesn't have an unknown offset
+                    // For timestamp ranges neither boundaries should have an unknown offset
+                    if isl_version == IslVersion::V1_0
+                        && boundary.as_timestamp().unwrap().offset().is_none()
+                    {
+                        return invalid_schema_error(
+                            "Timestamp range boundary can not have an unknown offset",
+                        );
+                    }
```
popematt, line 850: I wish we didn't have this. 😢

popematt (approved):
One minor suggestion.

ion-schema/src/isl/mod.rs
desaikd: Makes sense. Will change that.

Issue #157, opened 2023-04-06
add implementation of `VariablyOccuringTypeReference`


desaikd:
We should be modeling `VariablyOccuringTypeReference` as a distinct type from `TypeReference` and stop treating `occurs` as a constraint.

_Originally posted by @popematt in https://github.com/amazon-ion/ion-schema-rust/pull/156#discussion_r1160291498_

desaikd:
Fixed with #179

PR #178, opened 2023-05-22
adds changes for logic and length constraints ISL 2.0

desaikd:
### Description of changes:
This PR works on adding changes for logic and length constraints as per ISL 2.0.

### List of changes:
* adds check for occurs to only be used for fields and `ordered_elements` constraint
* adds check for empty non negative integer range
* adds  ion-schema-tests for these constraints

----


ion-schema/src/isl/isl_type_reference.rs
```diff
@@ -131,6 +131,7 @@ impl IslTypeRefImpl {
         isl_version: IslVersion,
         value: &Element,
         inline_imported_types: &mut Vec<IslImportType>,
+        allow_variably_occurring_type: bool,
```
popematt, line 134: This looks like it is creating more tech debt that will need to be dealt with when we get around to #157. What are your thoughts on fixing #157 now and then resubmitting this PR vs. keeping it as it?

ion-schema/src/isl/isl_range.rs
```diff
@@ -525,8 +525,18 @@ impl<T: std::cmp::PartialOrd> RangeImpl<T> {
         end: TypedRangeBoundaryValue,
     ) -> IonSchemaResult<RangeImpl<usize>> {
         match (start, end) {
-            (TypedRangeBoundaryValue::NonNegativeInteger(v1), TypedRangeBoundaryValue::NonNegativeInteger(v2)) => {
-                RangeImpl::range(v1, v2)
+            (TypedRangeBoundaryValue::NonNegativeInteger(Value(v1, v1_type)), TypedRangeBoundaryValue::NonNegativeInteger(Value(v2, v2_type))) => {
+                // verify this is not an empty range for which there is no valid non negative integer values
+                if v2 > v1 && v2 - v1 == 1
+                    && v1_type == RangeBoundaryType::Exclusive
+                    && v2_type == RangeBoundaryType::Exclusive
```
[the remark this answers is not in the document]
desaikd: I was thinking to do the same but since RangeImpl is generic, I would have to unwrap `RangeImpl<T>` to get a particular range type and then perform `is_empty()` check based on `T`. Which is why adding it here made sense. But that being said I could add something similar in `TypedRangeBoundary`.

ion-schema/src/isl/isl_type_reference.rs
desaikd: I will create another PR for `VariablyOccuringTypeReference`, until then keeping this PR open.

popematt (approved):

PR #179, opened 2023-05-23
adds implementation of `IslVariablyOccurringTypeRef`

desaikd:
### Description of changes:
This PR works on adding a variably occurring type reference implementation which will be used by the schema model.

### Grammar:
```ANTLR
<VARIABLY_OCCURRING_TYPE_ARGUMENT> ::= { <OCCURS>, <CONSTRAINT>... }
                                     | <TYPE_ARGUMENT>

<OCCURS> ::= occurs: <INT>
           | occurs: <RANGE_INT>
           | occurs: optional
           | occurs: required
```

### Ion Schema Specification:
https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#variably-occurring-type-arguments

### List of changes:
* adds implementation of `IslVariablyOccurringTypeRef` in isl module.
* adds changes for converting `IslVariablyOccurringTypeRef` to `VariablyOccurringTypeRef` which contains resolved `TypeId`.
* Construction of `ordered_elements` and `fields` constraints takes in `IslVariablyOccurringTypeRef` instead of previous `IslTypeRef`.
* modifies unit tests for variably occurring type references.

----

### DIFF
--- a/ion-schema/src/isl/isl_type_reference.rs
+++ b/ion-schema/src/isl/isl_type_reference.rs
@@ -6,7 +6,7 @@ use crate::result::{
     invalid_schema_error, invalid_schema_error_raw, unresolvable_schema_error, IonSchemaResult,
 };
 use crate::system::{PendingTypes, TypeId, TypeStore};
-use crate::type_reference::TypeReference;
+use crate::type_reference::{TypeReference, VariablyOccurringTypeRef};
 use crate::types::TypeDefinitionImpl;
 use crate::IslConstraintImpl;
 use ion_rs::element::Element;
@@ -15,8 +15,11 @@ use ion_rs::IonType;
 /// Provides public facing APIs for constructing ISL type references programmatically for ISL 1.0
 pub mod v_1_0 {
     use crate::isl::isl_constraint::{IslConstraint, IslConstraintImpl};
+    use crate::isl::isl_range::Range;
     use crate::isl::isl_type::IslTypeImpl;
-    use crate::isl::isl_type_reference::{IslTypeRef, IslTypeRefImpl, NullabilityModifier};
+    use crate::isl::isl_type_reference::{
+        IslTypeRef, IslTypeRefImpl, IslVariablyOccurringTypeRef, NullabilityModifier,
+    };
     use ion_rs::IonType;
 
     /// Creates a named [IslTypeRef] using the name of the type referenced inside it
@@ -47,12 +50,23 @@ pub mod v_1_0 {
             NullabilityModifier::Nothing,
         ))
     }
+
+    /// Creates an [IslVariablyOccurringTypeRef] using the [IslConstraint]s and [Range] referenced inside it
+    pub fn variably_occurring_type_ref(
+        type_ref: IslTypeRef,
+        occurs: Option<Range>,
+    ) -> IslVariablyOccurringTypeRef {
+        IslVariablyOccurringTypeRef::new(type_ref, occurs)
+    }
 }
 
 /// Provides public facing APIs for constructing ISL type references programmatically for ISL 2.0
 pub mod v_2_0 {
     use crate::isl::isl_constraint::IslConstraint;
-    use crate::isl::isl_type_reference::{v_1_0, IslTypeRef, IslTypeRefImpl, NullabilityModifier};
+    use crate::isl::isl_range::Range;
+    use crate::isl::isl_type_reference::{
+        v_1_0, IslTypeRef, IslTypeRefImpl, IslVariablyOccurringTypeRef, NullabilityModifier,
+    };
 
     /// Creates a named [IslTypeRef] using the name of the type referenced inside it
     pub fn named_type_ref<A: Into<String>>(name: A) -> IslTypeRef {
@@ -70,6 +84,14 @@ pub mod v_2_0 {
     /// Creates an anonymous [IslTypeRef] using the [IslConstraint]s referenced inside it
     pub fn anonymous_type_ref<A: Into<Vec<IslConstraint>>>(constraints: A) -> IslTypeRef {
         v_1_0::anonymous_type_ref(constraints)
+    }
+
+    /// Creates an anonymous [IslTypeRef] using the [IslConstraint]s and [Range] referenced inside it
+    pub fn variably_occurring_type_ref(
+        type_ref: IslTypeRef,
+        occurs: Option<Range>,
+    ) -> IslVariablyOccurringTypeRef {
+        v_1_0::variably_occurring_type_ref(type_ref, occurs)
     }
 }
 
@@ -126,12 +148,11 @@ impl IslTypeRefImpl {
         }
     }
 
-    /// Tries to create an [IslTypeRef] from the given Element
-    pub fn from_ion_element(
+    fn from_ion_element_with_occurs_flag(
         isl_version: IslVersion,
         value: &Element,
         inline_imported_types: &mut Vec<IslImportType>,
-        allow_variably_occurring_type: bool,
+        allow_occurs_field: bool,
     ) -> IonSchemaResult<Self> {
         use crate::isl::isl_type_reference::NullabilityModifier::*;
         let nullability = if value.annotations().contains("nullable") {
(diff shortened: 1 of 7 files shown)


ion-schema/src/isl/isl_type_reference.rs
```diff
@@ -311,3 +326,90 @@ impl IslTypeRefImpl {
...
+///            | occurs: <RANGE_INT>
+///            | occurs: optional
+///            | occurs: required
+/// ```
+///
+/// [variably occurring type reference]: https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#variably-occurring-type-arguments
+#[derive(Debug, Clone, PartialEq)]
+pub struct IslVariablyOccurringTypeRef {
+    type_ref: IslTypeRefImpl,
+    occurs: Option<Range>,
```
popematt, line 346: A variably occurring type reference always has an occurs range, though sometimes it is implicit. What do you think about modeling this not optional? 

To keep it nice to construct, we could have constructors such as:
```rust 
    pub fn optional(type_ref: IslTypeRef) -> Self {
        Self {
            type_ref: type_ref.type_reference,
            // TODO: add occurs optional here
        }
    }

    pub fn required(type_ref: IslTypeRef) -> Self {
        Self {
            type_ref: type_ref.type_reference,
            // TODO: add occurs required here
        }
    }
```

ion-schema/src/isl/isl_type_reference.rs
```diff
@@ -311,3 +326,90 @@ impl IslTypeRefImpl {
...
+                if constraint_name == "fields" {
+                    Range::optional()
+                } else {
+                    Range::required()
+                }
+            }
+            Some(range) => range.to_owned(),
+        };
+
+        Ok(VariablyOccurringTypeRef::new(type_ref, occurs))
```
popematt, line 413: This seems out of place. It's not a part of resolving the type reference. Rather it's a context-dependent part of reading the ISL in the first place.

desaikd: modified this to be part of the construction process rather then type reference resolution.

ion-schema/src/isl/isl_type_reference.rs
```diff
@@ -311,3 +326,104 @@ impl IslTypeRefImpl {
...
+        inline_imported_types: &mut Vec<IslImportType>,
+    ) -> IonSchemaResult<Self> {
+        let type_ref = IslTypeRefImpl::from_ion_element_with_occurs_flag(
+            isl_version,
+            value,
+            inline_imported_types,
+            true,
+        )?;
+
+        let occurs: Range = if let IslTypeRefImpl::Anonymous(isl_type, _) = &type_ref {
```
popematt, line 389: If you change this line to `let occurs: Option<Range> = ...`, then you can eliminate most/all of L397-408, and replace the return value with
```rust
match occurs {
    Some(range) => Ok(IslVariablyOccurringTypeRef::new(type_ref, occurs)),
    None if constraint_name == "fields" => Ok(IslVariablyOccurringTypeRef::optional(type_ref)),
    None => Ok(IslVariablyOccurringTypeRef::required(type_ref))
}
```

That is one possible way to fix the clippy warning and to get rid of the duplicated if/else expression.

desaikd: I have used a similar solution using `and_then` and `unwrap_or`. Please let me know if that sounds good?

ion-schema/src/isl/isl_type_reference.rs
```diff
@@ -311,3 +326,104 @@ impl IslTypeRefImpl {
...
+            isl_type
+                .constraints()
+                .iter()
+                .find_map(|c| match c {
+                    IslConstraintImpl::Occurs(occurs) => Some(occurs.to_owned()),
```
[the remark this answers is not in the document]
desaikd: Removed `Occurs` from `IslConstraintImpl` and `Constraint`

ion-schema/src/isl/isl_type_reference.rs
```diff
@@ -208,20 +207,21 @@ impl IslTypeRefImpl {
...
                         return invalid_schema_error(
                             "`$null_or` annotation is not supported for a type reference with an explicit `occurs` field",
                         )
                     }
 
-                    if !allow_occurs_field && isl_version == IslVersion::V2_0 && type_def.constraints()
-                        .iter()
-                        .any(|c| matches!(c, IslConstraintImpl::Occurs(_))) {
-                        return invalid_schema_error(
+                    if let Some(occurs_constraint) = value_struct.get("occurs") {
```
popematt, line 216: Little nitpick about `occurs` not being a constraint—maybe use `occurs_field` instead of `occurs_constraint`.

ion-schema/src/isl/isl_type_reference.rs
```diff
@@ -208,20 +207,21 @@ impl IslTypeRefImpl {
...
+                        if !allow_occurs_field && isl_version == IslVersion::V2_0 {
+                            return invalid_schema_error(
                             "A type reference with an explicit `occurs` field can only be used for `fields` and `ordered_elements` constraint",
-                        )
+                            )
+                        } else {
+                            // for ISL 1.0 ``occurs` field is a no op when used with constraints other than `fields` and `ordered_elements`
+                            IslTypeRefImpl::occurs_from_ion_element(occurs_constraint, isl_version)?;
```
popematt, line 223: Is this here just to check for errors of some sort?

ion-schema/src/isl/isl_type_reference.rs
```diff
@@ -247,6 +247,43 @@ impl IslTypeRefImpl {
         }
     }
 
+    fn occurs_from_ion_element(value: &Element, isl_version: IslVersion) -> IonSchemaResult<Range> {
```
popematt, line 250: I'd suggest that this function is better suited as a part of `IslVariablyOccurringTypeRef`.

ion-schema/src/isl/isl_type_reference.rs
popematt: Yes. Looks good.

ion-schema/src/isl/isl_type_reference.rs
desaikd: Correct. Just to verify occurs is not zero/null or empty range.

popematt: Is there any way it can be done in `IslVariablyOccurringTypeRef` instead?

desaikd: `IslVariablyOccurringTypeRef` will be created for constraints `fields` and `ordered_elements`. For any other constraints `IslTypeRef` will be created. For ISL 2.0 this will return an error but for ISL 1.0 we don't have the requirement to return an error but the `occurs` field related rules still counts, which is why I thought this will be the best place for it.
Why do you think this should be part of `IslVariablyOccurringTypeRef`?

popematt: > Why do you think this should be part of `IslVariablyOccurringTypeRef`?

`IslVariablyOccurringTypeRef` is the type that actually cares/uses the `occurs` value, hence it seems appropriate there.

> For any other constraints IslTypeRef will be created. For ISL 2.0 this will return an error but for ISL 1.0 we don't have the requirement to return an error but the occurs field related rules still counts

So, in ISL 1.0, you're saying that occurs _is_ allowed in locations where it's basically just open content, but it still has to be a proper range in those locations? That's really weird, but if that's the case I guess it makes sense to check it here. I wonder if it would be better to just ignore `occurs` in wrong locations.

desaikd: > So, in ISL 1.0, you're saying that occurs is allowed in locations where it's basically just open content, but it still has to be a proper range in those locations? 
`occurs` can not be treated as open content as its part of the schema specification. But it can be used  in wrong locations with ISL 1.0 since it doesn't restrict `occurs` to be used with `fields` and `ordered_elements` only. 
This is definitely weird but its has to do with how ISL 1.0 accepts `occurs` in wrong locations but still doesn't consider it open content but rather serializes it as `occurs` and afterwards ignores it because its used in wrong location. 
Please let me know if I am misunderstanding this so that I can refactor accordingly.

popematt: > Please let me know if I am misunderstanding this so that I can refactor accordingly.

No, you're not misunderstanding. I was just wistfully speculating that it would be nicer to implement if we could treat it as open content in places where it does nothing.

A little more info in the comment explaining why it is necessary to check `occurs` here might be helpful because other people might have the same question that I did.

popematt (approved):

ion-schema/src/isl/isl_type_reference.rs
desaikd: Got it. sure. I will add that.

PR #180, opened 2023-06-05
adds fixes for ISL 2.0 open content

desaikd:
### Description of changes:
This PR works on adding fixes for ISL 2.0 open content.

### List of changes:
* adds version information in the `IslSchema` model
* modifies generator for invalid schema tests to return `Ok` on violation
* modifies user reserved fields to not include reserved keywords and ISL 2.0 keywords
* modifies `isl_schema_from_elements` to consider the position of version marker for ISL 2.0

### Test:
adds tests from `ion-schema-tests` for ISL 2.0 open content

----

### DIFF
--- a/ion-schema/src/lib.rs
+++ b/ion-schema/src/lib.rs
@@ -45,7 +45,7 @@ pub mod external {
 }
 
 static ISL_VERSION_MARKER_REGEX: OnceLock<Regex> = OnceLock::new();
-static RESERVED_KEYWORD_ISL_VERSION_MARKER: OnceLock<Regex> = OnceLock::new();
+static RESERVED_WORD_REGEX: OnceLock<Regex> = OnceLock::new();
 
 /// Checks if a value is an ISL version marker.
 fn is_isl_version_marker(text: &str) -> bool {
@@ -55,8 +55,8 @@ fn is_isl_version_marker(text: &str) -> bool {
 }
 
 /// Checks is a value is reserved keyword ISL version maker.
-fn is_reserved_keyword_isl_version_marker(text: &str) -> bool {
-    RESERVED_KEYWORD_ISL_VERSION_MARKER
+fn is_reserved_word(text: &str) -> bool {
+    RESERVED_WORD_REGEX
         .get_or_init(|| Regex::new(r"^(\$ion_schema(_.*)?|[a-z][a-z0-9]*(_[a-z0-9]+)*)$").unwrap())
         .is_match(text)
 }
@@ -273,10 +273,10 @@ impl UserReservedFields {
             return invalid_schema_error("User reserved fields mut be unannotated");
         }
 
-        if user_reserved_fields.iter().any(|f| {
-            is_reserved_keyword_isl_version_marker(f)
-                || ISL_2_0_KEYWORDS.binary_search(&f.as_str()).is_ok()
-        }) {
+        if user_reserved_fields
+            .iter()
+            .any(|f| is_reserved_word(f) || ISL_2_0_KEYWORDS.binary_search(&f.as_str()).is_ok())
+        {
             return invalid_schema_error(
                 "ISl 2.0 keywords may not be declared as user reserved fields",
             );
--- a/ion-schema/src/lib.rs
+++ b/ion-schema/src/lib.rs
@@ -11,6 +11,7 @@ use ion_rs::element::{Element, Struct};
 use ion_rs::Symbol;
 use regex::Regex;
 use std::fmt::{Display, Formatter};
+use std::sync::OnceLock;
 /// A [`try`]-like macro to workaround the [`Option`]/[`Result`] nested APIs.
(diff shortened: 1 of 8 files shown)


ion-schema/src/schema.rs
```diff
@@ -324,13 +323,12 @@ mod schema_tests {
         let mut resolver = Resolver::new(vec![]);
 
         // create a isl from owned_elements and verifies if the result is `ok`
-        let isl =
-            resolver.isl_schema_from_elements(IslVersion::V1_0, owned_elements, "my_schema.isl");
-        assert!(isl.is_ok());
+        let isl_result = resolver.isl_schema_from_elements(owned_elements, "my_schema.isl");
+        assert!(isl_result.is_ok());
 
+        let isl = isl_result.unwrap();
```
popematt, line 329: minor suggestion—it seems strange to have an `assert!` followed by an `unwrap`. Why not just unwrap right away? (Or even better, use `expect(...)` with a message to make the test code more self-describing and the failure message more helpful for future developers.)
```suggestion
        let isl = isl_result.expect("<Thing we're checking> should be <condition we're checking for>");
```

ion-schema/src/system.rs
```diff
@@ -754,37 +753,79 @@ impl Resolver {
         let mut isl_inline_imports: Vec<IslImportType> = vec![];
         let mut open_content = vec![];
         let mut isl_user_reserved_fields = UserReservedFields::default();
+        let mut isl_version = IslVersion::V1_0;
+        let isl_version_marker: Regex = Regex::new(r"^\$ion_schema_\d.*$").unwrap();
+        let reserved_keyword_version_marker =
+            Regex::new(r"^(\$ion_schema(_.*)?|[a-z][a-z0-9]*(_[a-z0-9]+)*)$").unwrap();
```
popematt, line 759: Can we move these to a static [`OnceLock`](https://doc.rust-lang.org/std/sync/struct.OnceLock.html)? E.g.
```rust
static ISL_VERSION_MARKER_REGEX: OnceLock<Regex> = OnceLock::new();

/// Checks if a value is an ISL version marker.
fn is_isl_version_marker(text: &str) -> bool {
    ISL_VERSION_MARKER_REGEX
        .get_or_init(|| Regex::new(r"^\$ion_schema_\d.*$").unwrap())
        .is_match(text)
}
```

Alternately, we can create a new issue to update the Rust version and deal with all of the regexes.

popematt (changes requested):
Overall it looks good.

Can you update the `ion-schema-tests` submodule as part of this PR? There are some changes for bad/broken open content tests that were added since the the current submodule commit (in particular, amazon-ion/ion-schema-tests#46 and amazon-ion/ion-schema-tests#47).

ion-schema/src/system.rs
desaikd: I have used `OnceLock<Regex>` here.

ion-schema/src/lib.rs
```diff
@@ -175,37 +225,63 @@ pub struct UserReservedFields {
...
             .collect::<Option<Vec<String>>>()
-            .unwrap_or(vec![])
+            .unwrap_or(vec![]);
+
+        if user_reserved_fields.len() != user_reserved_elements.len() {
+            return invalid_schema_error("User reserved fields mut be unannotated");
+        }
+
+        if user_reserved_fields.iter().any(|f| {
+            is_reserved_keyword_isl_version_marker(f) || ISL_2_0_KEYWORDS.contains(&f.as_str())
```
zslayton, line 277: Small optimization opportunity since `ISL_2_0_KEYWORDS` is in alphabetical order: use [`binary_search`](https://doc.rust-lang.org/std/primitive.slice.html#method.binary_search) instead of `contains`.

ion-schema/src/system.rs
```diff
@@ -754,37 +753,78 @@ impl Resolver {
         let mut isl_inline_imports: Vec<IslImportType> = vec![];
         let mut open_content = vec![];
         let mut isl_user_reserved_fields = UserReservedFields::default();
+        let mut isl_version = IslVersion::V1_0;
+        let reserved_keyword_version_marker =
+            Regex::new(r"^(\$ion_schema(_.*)?|[a-z][a-z0-9]*(_[a-z0-9]+)*)$").unwrap();
```
zslayton, line 758: Does this need a `OnceLock`?

popematt: I recommended it. `OnceLock` is the thread-safe version of `OnceCell` that was just release in Rust 1.70, and it's a good way to move the compilation of the regex out of the regular codepath and into a separate initialization code path.

desaikd: Yes, I meant to add it but had missed this place.

popematt (commented):
Overall I think it looks pretty good, but the `ion-schema-tests` submodule still ought to be updated since it has relevant test cases.

popematt (approved):

PR #182, opened 2023-06-07
adds `new_schema()` for `SchemaSystem` and minor ISL model fixes

desaikd:
### Description of changes:
This PR works on adding `new_schema` for `SchemaSystem` and adds minor ISL model related fixes.

### List of changes:
* adds new_schema and new_isl_schema in `SchemaSystem`
* adds fixes for user reserved fields
  * `user_reserved_fields` should not consider validation for `imports`.
  * Incorrect error "User reserved field must be non-null and unannotated struct" was thrown when `user_reserved_fields` didn't even exist. 
* adds version information for Annotation
  * this will be useful to determine whether annotation is allowed or not
for each annotation in the annotations list
* adds a fix for inline import
  * inline import should not accept `as` for ISL 2.0
  * inline import should consider `as` open content for ISL 1.0

### Test: 
adds tests for `new_schema()` and `new_isl_schema()` .

----

### DIFF
--- a/ion-schema/src/system.rs
+++ b/ion-schema/src/system.rs
@@ -1144,11 +1144,32 @@ impl SchemaSystem {
             .load_schema(id, &mut TypeStore::default(), None)
     }
 
+    /// Constructs a new schema using provided ISL content.
+    pub fn new_schema(&mut self, schema_content: &[u8], id: &str) -> IonSchemaResult<Arc<Schema>> {
+        let elements = Element::read_all(schema_content)?;
+        let isl = self
+            .resolver
+            .isl_schema_from_elements(elements.into_iter(), id)?;
+        self.resolver
+            .schema_from_isl_schema(isl.version(), isl, &mut TypeStore::default(), None)
+    }
+
     /// Requests each of the provided [`DocumentAuthority`]s, in order, to get ISL model for the
     /// requested schema id until one successfully resolves it.
     /// If an authority throws an exception, resolution silently proceeds to the next authority.
     pub fn load_isl_schema<A: AsRef<str>>(&mut self, id: A) -> IonSchemaResult<IslSchema> {
         self.resolver.load_isl_schema(id, None)
+    }
+
+    /// Constructs a new ISL model using provided ISL content.
+    pub fn new_isl_schema(
+        &mut self,
+        schema_content: &[u8],
+        id: &str,
+    ) -> IonSchemaResult<IslSchema> {
+        let elements = Element::read_all(schema_content)?;
+        self.resolver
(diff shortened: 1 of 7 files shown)


ion-schema/src/system.rs
```diff
@@ -796,21 +796,22 @@ impl Resolver {
...
                     {
+                        if !user_reserved_fields_element.annotations().is_empty() {
+                            return invalid_schema_error(
+                                "User reserved field must be an unannotated struct",
+                            )?;
+                        }
+                        let user_reserved_fields_struct = user_reserved_fields_element
+                            .as_struct()
+                            .ok_or(invalid_schema_error_raw(
+                                "User reserved field must be non-null struct",
```
zslayton, line 810:
```suggestion
                                "User reserved field must be a non-null struct",
```

ion-schema/src/system.rs
```diff
@@ -1144,13 +1145,34 @@ impl SchemaSystem {
...
+    /// Constructs a new ISL model using provided ISL content.
+    pub fn new_isl_schema(
+        &mut self,
+        schema_content: &[u8],
+        id: &str,
+    ) -> IonSchemaResult<IslSchema> {
+        let elements = Element::read_all(schema_content)?;
+        self.resolver
+            .isl_schema_from_elements(elements.into_iter(), id)
+    }
```
zslayton, line 1174: As discussed offline: since this returns an unresolved model of the desired schema, it might be possible to have it live somewhere else to avoid the naming confusion with the `new_schema` method above. If it's feasible, that move doesn't have to happen in this PR.

ion-schema/src/system.rs
```diff
@@ -2167,6 +2191,90 @@ mod schema_system_tests {
...
+                schema_header::{}
+                
+                type::{
+                  name: my_type,
+                  type: string,
+                }
+                
+                schema_footer::{}
+            "#
+            .as_bytes(),
```
zslayton, line 2209: You can use the [`b` string literal prefix](https://doc.rust-lang.org/reference/tokens.html#byte-string-literals) to get a `&[u8]` from a `&str`:
```rust
fn takes_bytes(_data: &[u8]) {}

fn main() {
    takes_bytes(b"hello");
}
```
[Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6af0a4ba5de9205cc7ba1a94a623bbfd)

In your example, it'd be `br#"...",`.

zslayton (approved):

ion-schema/src/system.rs
desaikd: I will create a separate PR for this as currently this involves getting `isl_schema_from_elements` that lives in the resolver.

PR #183, opened 2023-06-08
adds implementation for `WriteToIsl`

desaikd:
### Description of changes:
This PR works on adding implementation of `WriteToIsl` which serializes a schema model into a schema file using Ion writer.

### Example:
Here is a quick example of how to serialize a schema model using Ion writer.
```rust
// construct an ISL type programmatically
let isl_type = named_type(
    // represents the `name` of the defined type
    "my_number_type".to_owned(),
    vec![
        // represents `one_of` constraint
        one_of(vec![
            named_type_ref("int"),
            named_type_ref("float"),
            named_type_ref("decimal"),
        ]),
    ],
);

// create an ISL schema using above `isl_type`
let isl_schema = IslSchema::schema_v_1_0(
    "my_schema.isl", // represents schem id which is usually same as the schema file name
    vec![], // represents imports
    vec![isl_type.to_owned()],
    vec![], // represents inline imports
    vec![], // represents open content
);

// use an Ion Writer to write the ISL model
let mut buffer = Vec::new();
let mut writer = TextWriterBuilder::pretty().build(&mut buffer)?;
isl_schema.write_to(&mut writer)?;

// write into schema file
let mut file = File::create("my_schema.isl")?;
file.write_all(writer.output().as_bytes())?;
```
Output:
```ion
$ion_schema_1_0
schema_header::{}
type::{
    name: "my_number_type",
    one_of: [
        int,
        float,
        decimal
    ]
}
schema_footer::{}
```
### List of changes:
* adds `WriteToIsl` to serialize isl model into schema file
  * Defines `WriteToIsl` trait which will be sued by all ISL model structs to write schema content to an Ion writer
  * Implements `WriteToIsl` for `IslSchema`
* adds `WriteToIsl` implementation for 
  * `UserReservedFields`
  * ISL imports
  * ISL type references
  * ISL constraints
  * ISL ranges
  * ISL model utils
### Test:
Adds unit tests for `WriteToIsl` implementations. 

----


ion-schema/src/isl/mod.rs
```diff
@@ -945,4 +1007,50 @@ mod isl_tests {
         write!(&mut buf, "{}", range.into()).unwrap();
         assert_eq!(expected, String::from_utf8(buf).unwrap());
     }
+
+    const SKIP_LIST: [&str; 5] = [
+        "ion-schema-schemas/json/json.isl", // the file contains `nan` which fails on equivalence for two schemas
+        "ion-schema-tests/ion_schema_1_0/nullable.isl", // Needs `nullable` annotation related fixes
+        "ion-schema-tests/ion_schema_1_0/schema/import/import_inline.isl", // related to order of types in the schema file
+        "ion-schema-tests/ion_schema_2_0/imports/tree/inline_import_a.isl", // related to order of types in the schema file
+        "ion-schema-tests/ion_schema_2_0/imports/tree/inline_import_c.isl", // related to order of types in the schema file
```
popematt, line 1016: Do we have any GitHub issue to track the `nan` issue or the "order of types in a schema file" issue?

ion-schema/src/lib.rs
```diff
@@ -359,3 +360,30 @@ impl UserReservedFields {
         Ok(())
     }
 }
+
+impl WriteToIsl for UserReservedFields {
+    fn write_to<W: IonWriter>(&self, writer: &mut W) -> IonSchemaResult<()> {
```
popematt, line 365: _Very_ minor suggestion—there's a lot of stepping in and out in this function. It might be helpful to add comments or spacing to make it clear what bits belong together.

ion-schema/src/isl/mod.rs
```diff
@@ -945,4 +1007,50 @@ mod isl_tests {
...
+    fn is_skip_list_path(file_name: &str) -> bool {
+        SKIP_LIST
+            .iter()
+            .map(|p| p.replace('/', &MAIN_SEPARATOR.to_string()))
+            .any(|p| p == file_name)
+    }
+
+    #[test_resources("ion-schema-tests/**/*.isl")]
+    #[test_resources("ion-schema-schemas/**/*.isl")]
+    fn test_write_to_isl(file_name: &str) {
```
popematt, line 1028: I think I know what it's doing, but can you add a comment describing how this is testing the writer trait impl?

ion-schema/src/isl/isl_type_reference.rs
```diff
@@ -460,3 +502,31 @@ impl IslVariablyOccurringTypeRef {
...
+            IslTypeRefImpl::Named(name, nullability_modifier) => {
+                nullability_modifier.write_to(writer)?;
+                writer.write_symbol(name)?;
+            }
+            IslTypeRefImpl::TypeImport(type_import, nullability_modifier) => {
+                nullability_modifier.write_to(writer)?;
+                writer.step_in(IonType::Struct)?;
+                type_import.write_to(writer)?;
+                writer.step_out()?;
+            }
```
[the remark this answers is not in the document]
desaikd: Right, for `TypeImport` I will modify it to add the `occurs` field here.  For `Named` if it has an `occurs` other than the default value then it becomes an `Anonymous`(inline) type reference. 
e.g.
```
// Named
fields : {
   name: string
}

// Anonymous
fields: {
    name: { type: string, occurs: required }
}
```

ion-schema/src/isl/mod.rs
desaikd: Created an issue for this, I will add it in the comments for these `SKIP_LIST` items.

ion-schema/src/isl/isl_range.rs
```diff
@@ -944,6 +976,23 @@ impl<T: Display> Display for RangeBoundaryValue<T> {
...
+impl<T: Display> WriteToIsl for RangeBoundaryValue<T> {
+    fn write_to<W: IonWriter>(&self, writer: &mut W) -> IonSchemaResult<()> {
+        match &self {
+            Max => writer.write_symbol("max")?,
+            Min => writer.write_symbol("min")?,
+            Value(value, range_boundary_type) => {
+                if range_boundary_type == &RangeBoundaryType::Exclusive {
+                    writer.set_annotations(["exclusive"]);
+                }
+                let element = Element::read_one(format!("{value}").as_bytes())?;
```
zslayton, line 988: I don't have a concrete suggestion at the moment, but this bit feels a bit fragile to me. We're using the constraint `T: Display` but actually hoping that `T` is something that's serialized as text Ion, which is a narrower set of types.

ion-schema/src/isl/isl_type.rs
```diff
@@ -217,6 +219,20 @@ impl IslTypeImpl {
     }
 }
 
+impl WriteToIsl for IslTypeImpl {
+    fn write_to<W: IonWriter>(&self, writer: &mut W) -> IonSchemaResult<()> {
+        writer.set_annotations(["type"]);
+        writer.step_in(IonType::Struct)?;
+        writer.set_field_name("name");
+        writer.write_symbol(self.name.clone().unwrap())?;
```
zslayton, line 227: In this circumstance, is it guaranteed that `self.name` is always `Some(_)`?

This version will return an error instead of panicking if that were ever to occur:
```rust
self.name.map(|name| writer.write_symbol(name))?;
```

ion-schema/src/isl/mod.rs
```diff
@@ -273,14 +331,18 @@ mod isl_tests {
...
     use crate::result::IonSchemaResult;
+    use crate::system::SchemaSystem;
     use ion_rs::element::Element;
     use ion_rs::types::decimal::*;
     use ion_rs::types::integer::Int as IntegerValue;
     use ion_rs::types::timestamp::Timestamp;
-    use ion_rs::IonType;
     use ion_rs::Symbol;
+    use ion_rs::{IonType, TextWriterBuilder};
+    use nom::AsBytes;
```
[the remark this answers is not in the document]
zslayton: `nom` is a pretty heavy dependency to bring in for [the `AsBytes` trait](https://docs.rs/nom/latest/nom/trait.AsBytes.html), which seems to be a helper for fixed-sized arrays. Consider either:
* Removing the dependency and using [`Vec::as_slice`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.as_slice) to access the bytes of the `Vec<u8>`.
* Downgrading the dependency to a dev dependency.

zslayton (approved):

ion-schema/src/isl/mod.rs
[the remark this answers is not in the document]
desaikd: > `nom` is a pretty heavy dependency to bring in for [the `AsBytes` trait](https://docs.rs/nom/latest/nom/trait.AsBytes.html), which seems to be a helper for fixed-sized arrays. Consider either:
> 
>     * Removing the dependency and using [`Vec::as_slice`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.as_slice) to access the bytes of the `Vec<u8>`.
> 
>     * Downgrading the dependency to a dev dependency.

According to your suggestion and since I am using this only for tests, I will remove the dependency and just use `as_slice`.

popematt (approved):

PR #188, opened 2023-06-28
Updates `ion-rs` dependency to v0.18.1

desaikd:
### Description of changes:
This PR updates `ion-rs` dependency to v0.18.1.

### List of changes:
- changes the imports statements for some of the modules.
- Uses `IonData` which provides `ion_eq` or Ion data model equivalence based on [Ion specification](https://amazon-ion.github.io/ion-docs/docs/spec.html).

----

### DIFF
--- a/ion-schema/src/constraint.rs
+++ b/ion-schema/src/constraint.rs
@@ -18,7 +18,7 @@ use crate::types::TypeValidator;
 use crate::violation::{Violation, ViolationCode};
 use crate::IonSchemaElement;
 use ion_rs::element::Element;
-use ion_rs::ion_eq::IonEq;
+use ion_rs::IonData;
 use ion_rs::{Int, IonType};
 use num_traits::ToPrimitive;
 use regex::{Regex, RegexBuilder};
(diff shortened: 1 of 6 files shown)


ion-schema/src/constraint.rs
```diff
@@ -2042,9 +2042,11 @@ impl ConstraintValidator for ValidValuesConstraint {
                         },
                         ValidValue::Element(element) => {
                             // get value without annotations
-                            let value = value.value();
+                            let value: IonData<_> = value.value().into();
+                            let actual_value: IonData<_> = element.value().into();
 
-                            if element.value().ion_eq(value) {
+                            // this comparison uses the Ion equivalence based on Ion specification
+                            if actual_value == value {
```
popematt, line 2049: FYI—It should also be possible to do this:

```rust
if IonData::eq(value.value(), element.value()) {
```

popematt (approved):

ion-schema/src/constraint.rs
desaikd: It seems to require types for both the arguments of `IonData::eq()`.

PR #190, opened 2023-06-28
Refactor Ranges so that they are easier to use

popematt:
### Issue #, if available:

N/A

### Description of changes:

Current implementation of Range has unnecessary structs/enums that make it confusing to use and maintain. `Range` was an enum of the possible types of ranges, and was used in places where there should only be one type of range. The Rust-y thing to do would be to rely on generics/monomorphization of `RangeImpl`. (This is also problem in the Kotlin implementation that was ported over when `ion-schema-rust` was in its early days.)

In addition, `TypedRangeBoundaryValue` and `RangeBoundaryValue` had a lot of redundancy, and `RangeBoundaryType` models inclusivity/exclusivity, but it's not truly orthogonal with the boundedness/unboundedness (which is modeled in `RangeBoundaryValue` and `TypedRangeBoundaryValue`).

Major changes in this PR include:

* There's no longer an untyped range
* Collapsed the exclusivity and the boundedness into a single enum (`Limit`)
* Updated all range-based constraints to have the appropriate type of range (e.g. `Utf8ByteLength(Range)` => `Utf8ByteLength(UsizeRange)`)
* Change `ValidValue` to be an enum of `Element`, `NumberRange`, and `TimestampRange` – this is required because there's no longer a single range type that covers all ranges.
* Change `ValidValue::Element` to hold a `Value`. Now it's impossible to construct a valid values constraint that has annotations on a `ValidValue::Element`. (May as well use the type system to our advantage.)

There's also PR tour notes in the PR diff that call out smaller changes.

Some things that I don't have a strong opinion about and/or want feedback:
* Naming of `Limit` – it could also be `Boundary` and the variants could be `Unbounded`, `Inclusive`, and `Exclusive`.
* Should the `precision` constraint use `NonZeroU64` instead of `u64`? I wasn't sure if the benefit or the non-zero type would outweigh the loss of the convenience of using a primitive (`u64`) for this.

----


ion-schema/src/constraint.rs
```diff
@@ -130,7 +133,7 @@ impl Constraint {
             .map(|id| {
                 VariablyOccurringTypeRef::new(
                     TypeReference::new(*id, NullabilityModifier::Nothing),
-                    Range::required(),
+                    UsizeRange::new_single_value(1),
```
popematt, line 136: 🗺️ The concepts of "required" and "optional" don't inherently belong to ranges, so we're using `new_single_value(1)` here. There's also a `UsizeRange::zero_or_one()` that replaces `Range::optional()`.

ion-schema/src/isl/isl_type_reference.rs
```diff
@@ -433,55 +435,35 @@ impl IslVariablyOccurringTypeRef {
...
             })
             .unwrap_or(if constraint_name == "fields" {
-                Ok(Range::optional())
+                Ok(UsizeRange::zero_or_one())
             } else {
-                Ok(Range::required())
+                Ok(UsizeRange::new_single_value(1))
             })?;
 
+        isl_require!(occurs.upper() != &Limit::Closed(0usize) => "Occurs cannot be 0")?;
```
popematt, line 450: 🗺️ `isl_require!` is a macro I created along the lines of Kotlin's `require()` function to make it more concise to check ISL syntax. It takes a condition that must be true, and format args for an error message if it's false. You'll see the definition of the macro further down in `result.rs`.

ion-schema/src/isl/mod.rs
```diff
@@ -730,338 +716,6 @@ mod isl_tests {
         values.iter().cloned().map(|v| v.into()).collect()
     }
 
-    #[rstest(
```
popematt, line 733: 🗺️  These tests are for ranges, so when I rewrote the Range implementation, I moved/ported them from here to the new `ranges` module.

ion-schema/src/isl/mod.rs
```diff
@@ -1075,14 +729,14 @@ mod isl_tests {
...
-            .map(|p| p.replace('/', &MAIN_SEPARATOR.to_string()))
+            .map(|p| p.replace('/', std::path::MAIN_SEPARATOR_STR))
             .any(|p| p == file_name)
     }
 
     #[test_resources("ion-schema-tests/**/*.isl")]
     #[test_resources("ion-schema-schemas/**/*.isl")]
     fn test_write_to_isl(file_name: &str) {
-        if is_skip_list_path(&file_name) {
+        if is_skip_list_path(file_name) {
```
popematt, line 739: 🗺️ This is a Clippy fix.

ion-schema/src/isl/mod.rs
```diff
@@ -1075,14 +729,14 @@ mod isl_tests {
     fn is_skip_list_path(file_name: &str) -> bool {
         SKIP_LIST
             .iter()
-            .map(|p| p.replace('/', &MAIN_SEPARATOR.to_string()))
+            .map(|p| p.replace('/', std::path::MAIN_SEPARATOR_STR))
```
popematt, line 732: 🗺️ I'm pretty sure this was a Clippy fix.

ion-schema/src/isl/ranges.rs
```diff
@@ -0,0 +1,471 @@
...
+impl RangeOverType for Timestamp {}
+impl RangeOverType for TimestampPrecision {}
+impl RangeOverType for usize {
+    fn get_adjustment_for_open_lower_bound(self) -> Limit<Self> {
+        Limit::Closed(self + 1)
+    }
+    fn get_adjustment_for_open_upper_bound(self) -> Limit<Self> {
+        Limit::Closed(self - 1)
+    }
+}
```
popematt, line 76: 🗺️ The adjustment from `Open(n)` to `Closed(n ± 1)` is necessary for integer types in order to be able to make sure that the range is non-empty.

ion-schema/src/isl/ranges.rs
```diff
@@ -0,0 +1,471 @@
...
+            Limit::Unbounded => usize::MAX,
+            Limit::Closed(x) => x,
+            Limit::Open(x) => x - 1,
+        };
+        (lower, upper)
+    }
+}
+
+// TODO: consider changing to Range<NonZeroU64> since this is only used for the `precision` constraint
+pub type U64Range = Range<u64>;
```
popematt, line 239: What are your thoughts on this TODO?

ion-schema/src/isl/util.rs
```diff
@@ -103,6 +105,20 @@ impl TimestampPrecision {
             },
         }
     }
+
+    fn string_value(&self) -> String {
```
popematt, line 109: 🗺️ By moving the variant-to-string mapping here, we are able to call this function from the `Display` impl and the `WriteToIsl` impl.

ion-schema/src/lib.rs
```diff
@@ -233,8 +234,8 @@ impl UserReservedFields {
                 && f.text() != Some("type")
         }) {
             return invalid_schema_error(
-               "User reserved fields can only have schema_header, schema_footer or type as the field names",
-           );
+                "User reserved fields can only have schema_header, schema_footer or type as the field names",
+            );
```
popematt, line 238: 🗺️ This was a `cargo fmt` change—just fixing the indentation.

ion-schema/src/result.rs
```diff
@@ -91,3 +91,19 @@ pub fn unresolvable_schema_error_raw<S: AsRef<str>>(description: S) -> IonSchema
         description: description.as_ref().to_string(),
     }
 }
+
+/// A macro that checks some condition required to be valid ISL.
+///
+/// If invalid, returns an InvalidSchemaErr with the given error message.
+#[macro_export]
+macro_rules! isl_require {
+    ($expression:expr => $fmt_string:literal $(, $($tt:tt)*)?) => {
```
popematt, line 100: 🗺️ Nifty little macro to make it more concise to validate the ISL.

Example usage:
```rust
isl_require!(foo.is_bar() => "Foo must be bar, but was {foo}")?;
```

I'm not very keen on the `=>` to separate the condition from the format args, but there's a very limited set of characters that are allowed to follow an `expr` designator. (I originally wanted `else`, but the other options are `,` and `;`.)

Feel free to tell me that it should be `,` or `;` instead if you have an opinion on the matter.

ion-schema/src/types.rs
```diff
@@ -780,50 +778,22 @@ mod type_definition_tests {
...
-        */
-        anonymous_type([valid_values_with_values(vec![2.into(), Decimal::new(35, -1).into(), 5e7.into(), "hello".to_owned().into(), Symbol::from("hi").into()]).unwrap()]),
-    TypeDefinitionKind::anonymous([Constraint::valid_values_with_values(vec![2.into(), Decimal::new(35, -1).into(), 5e7.into(), "hello".to_owned().into(), Symbol::from("hi").into()], IslVersion::V1_0).unwrap(), Constraint::type_constraint(34)])
-    ),
-    case::valid_values_with_range_constraint(
-        /* For a schema with valid_values constraint as below:
-            { valid_values: range::[1, 5.5] }
-        */
-        anonymous_type(
-            [valid_values_with_range(
```
popematt, line 812: 🗺️ The `valid_values_with_range` function no longer exists, so these test cases are no longer necessary.

ion-schema/src/isl/isl_constraint.rs
```diff
@@ -938,14 +866,14 @@ impl IslConstraintImpl {
                             .map(|e| {
                                 if e.is_null() {
                                     return invalid_schema_error(
-                                    "`timestamp_offset` values must be non-null strings, found null"
-                                );
+                                        "`timestamp_offset` values must be non-null strings, found null",
+                                    );
```
popematt, line 870: 🗺️ Automated change from `cargo fmt`.

ion-schema/src/isl/isl_constraint.rs
```diff
@@ -938,14 +866,14 @@ impl IslConstraintImpl {
...
                                 }
 
                                 if e.ion_type() != IonType::String {
                                     return invalid_schema_error(format!(
-                                    "`timestamp_offset` values must be non-null strings, found {e}"
-                                ));
+                                        "`timestamp_offset` values must be non-null strings, found {e}"
+                                    ));
```
popematt, line 876: 🗺️ Automated change from `cargo fmt`.

ion-schema/src/isl/isl_constraint.rs
```diff
@@ -1372,54 +1298,28 @@ pub struct IslValidValuesConstraint {
...
-        let valid_values: IonSchemaResult<Vec<ValidValue>> = valid_values
-            .iter()
-            .map(|v| match v {
-                ValidValue::Range(r) => Ok(v.to_owned()),
-                ValidValue::Element(e) => ValidValue::from_ion_element(e, isl_version),
-            })
-            .collect();
-        Ok(Self {
-            valid_values: valid_values?,
-        })
```
popematt, line 1387: 🗺️ `ValidValue::Element` can no longer have annotations, so all of this logic is no longer necessary.

ion-schema/src/isl/mod.rs
```diff
@@ -368,33 +368,25 @@ impl IslSchemaImpl {
...
     use crate::system::SchemaSystem;
     use ion_rs::element::Element;
     use ion_rs::types::Decimal;
-    use ion_rs::types::Int as IntegerValue;
-    use ion_rs::types::Timestamp;
     use ion_rs::Symbol;
     use ion_rs::{IonType, TextWriterBuilder};
     use rstest::*;
-    use std::io::Write;
     use std::path::{Path, MAIN_SEPARATOR};
```
desaikd, line 390: Clippy suggestion to remove unused import for `MAIN_SEPARATOR`.

ion-schema/src/isl/ranges.rs
```diff
@@ -0,0 +1,471 @@
...
+use ion_rs::element::writer::ElementWriter;
+use ion_rs::element::Element;
+use ion_rs::Decimal;
+use ion_rs::{IonType, IonWriter, Timestamp};
+use std::fmt::{Display, Formatter};
+
+/// The end of a [`Range`].
+#[derive(Debug, PartialEq, Clone)]
+pub enum Limit<T> {
+    Unbounded,
```
desaikd, line 13: What does unbound represent here? It would be nice to add some document comments for these enum variants and how they are used.

ion-schema/src/isl/isl_type_reference.rs
desaikd: Nice

ion-schema/src/constraint.rs
desaikd: Instead of `new_single_value(1)` `one()` might make sense here and it also aligns with `zero_or_one()`

ion-schema/src/isl/ranges.rs
```diff
@@ -0,0 +1,471 @@
...
+        match self {
+            Limit::Unbounded => f.write_str(unbounded_text),
+            Limit::Closed(value) => value.fmt(f),
+            Limit::Open(value) => write!(f, "exclusive::{value}"),
+        }
+    }
+}
+
+/// Defines a type that can be "ranged over" using a [`Range`].
+pub trait RangeOverType: PartialEq + PartialOrd + Clone
```
desaikd, line 50: Nice

ion-schema/src/isl/ranges.rs
desaikd: I did not understand what its suggesting. But I think even if its only used by `precision` we should still keep it.

ion-schema/src/types.rs
desaikd: I think the test cases would still be necessary if we have different way to construct `valid_values` constraint with range, we should verify it here.

desaikd:
Overall this is a really nice improvement! Just added few minor comments and related questions.

ion-schema/src/isl/ranges.rs
popematt: The TODO is suggesting this:
```suggestion
pub type NonZeroU64Range = Range<NonZeroU64>;
```

desaikd: Oh yes, we should change it to `NonZeroU64`.

ion-schema/src/constraint.rs
```diff
@@ -281,13 +274,6 @@ impl Constraint {
         }))
     }
 
-    /// Creates a [Constraint::ValidValues] using the [Range] specified inside it
-    pub fn valid_values_with_range(value: Range) -> Constraint {
```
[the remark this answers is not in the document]
popematt: There are now two types of ranges for valid_values, so it would need to be changed to `valid_values_with_number_range` and `valid_values_with_timestamp_range`.

... or instead, you should be able to do this:
```rust
let my_range: NumberRange = something();
let cons1 = valid_values(vec![my_range.into()]);

let my_ts_range: TimestampRange = something_else();
let cons2 = valid_values(vec![my_ts_range.into()]);
```

Basically, it's extra surface area in our APIs, and I don't think it adds a lot of value.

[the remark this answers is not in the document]
desaikd: Yes, I think keep the `valid_values()` and using it as you mentioned sounds good to me.

ion-schema/src/constraint.rs
popematt: The benefit of `zero_or_one()` (and the only reason I created it) is that it returns a `UsizeRange` whereas `new_inclusive(0, 1)` returns a `IonSchemaResult<UsizeRange>`. Both `new_single_value(1)` and `one()` would return a `UsizeRange`, so the same motivation isn't there.

I'd like to defer a `one()` function for now (if you don't mind), but I could be convinced otherwise.

desaikd: I think having `new_single_value()` is fine but when we have something like `zero_or_one` user might expect a similar `one` which helps create required occurs. But this isn't a huge issue as we atleast have an alternative API `new_single_value().

ion-schema/src/constraint.rs
```diff
@@ -204,26 +203,22 @@ impl Constraint {
     }
 
     /// Creates a [Constraint::Precision] from a [Range] specifying a precision range.
-    pub fn precision(precision: RangeImpl<usize>) -> Constraint {
-        Constraint::Precision(PrecisionConstraint::new(Range::NonNegativeInteger(
-            precision,
-        )))
+    pub fn precision(precision: U64Range) -> Constraint {
```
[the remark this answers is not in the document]
popematt: It's not obvious, but here's why. All of the `*_length` constraints check a `len()` function that returns a `usize`. However, `Decimal::precision()` returns a `u64`. Rather than deal with a potentially lossy conversion if `usize` is not 64 bits, I'm just using both. I'll add a comment to that effect.

Now I'm curious to find out whether `u64` is the the best return type for `precision()`. (Maybe one of the non-zero types, even.)

ion-schema/src/ion_extension.rs
```diff
@@ -0,0 +1,45 @@
...
+    fn any_number_as_decimal(&self) -> Option<Decimal>;
+}
+impl ElementExtensions for Element {
+    fn as_usize(&self) -> Option<usize> {
+        match self.value() {
+            Value::Int(Int::I64(i)) => i.to_usize(),
+            Value::Int(Int::BigInt(i)) => i.to_usize(),
+            _ => None,
+        }
+    }
```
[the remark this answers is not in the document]
popematt: > Non-blocking concern: for these conversions, note that it will not be possible to tell the difference between "The Element was not an integer" and "The Element was an integer that couldn't fit in a usize."

I think that scenario is unlikely to be a problem—it will only occur if people are using numbers that are too large for `u64` and `usize` as arguments to constraints in their ISL. We can revisit it once the `expect_TYPE()` functions are available. For now, I'll make the documentation a little more explicit.

ion-schema/src/isl/ranges.rs
popematt: Yes. I'll fix that.

ion-schema/src/isl/ranges.rs
```diff
@@ -1,28 +1,46 @@
...
+
+//! Range types used in Ion Schema.
+//!
+//! ### Why so many range types?
+//! - [`UsizeRange`] is used for `*_length` constraints and `occurs`
+//! - [`U64Range`] is used for the `precision` constraint because [`Decimal`] precision is measured
+//!   using `u64`. It would be ideal to use [`NonZeroU64`][std::num::NonZeroU64] for this case, but
+//!   `NonZeroU64` is difficult to work with because it doesn't implement core ops (such as `Add`)
+//!   and as a result, it cannot even implement the checked ops traits (such as `CheckedAdd`) from
+//!   the `num_traits` crate. See https://github.com/rust-num/num-traits/issues/274.
```
zslayton, line 12: The "It would be ideal [...]" portion of this seems like it might be better suited to a developer comment, but I don't feel strongly about it.

ion-schema/src/isl/ranges.rs
```diff
@@ -345,16 +192,208 @@ impl WriteToIsl for U64Range {
...
+                | (Limit::Inclusive(lower), Limit::Exclusive(upper)) => lower >= upper,
+                _ => false,
+            }
+        }
+    }
+
+    /// Checks if two limits would result in an empty range of integers. Returns true if there are
+    /// no integer values between `start` and `end`, taking into consideration the exclusivity of
+    /// the limits.
+    pub fn _is_int_range_empty<T: CheckedAdd + One + PartialOrd>(
```
zslayton, line 217: In Rust, names with a leading `_` indicate that the item is never used but you wanted to give it a meaningful name anyway. Unless there's another convention of which I'm unaware, consider renaming this.

ion-schema/src/types.rs
```diff
@@ -795,6 +795,13 @@ mod type_definition_tests {
         anonymous_type([timestamp_precision(TimestampPrecisionRange::new_single_value(TimestampPrecision::Month))]),
     TypeDefinitionKind::anonymous([Constraint::timestamp_precision(TimestampPrecision::Month.into()), Constraint::type_constraint(34)])
     ),
+    case::valid_values_constraint(
+    /* For a schema with valid_values constraint as below:
+        { valid_values: [2, 3.5, 5e7, "hello", hi, range::[2, 3]] }
+    */
+    anonymous_type([valid_values(vec![2.into(), ion_rs::Decimal::new(35, -1).into(), 5e7.into(), "hello".to_owned().into(), Symbol::from("hi").into(), NumberRange::new_inclusive(2.into(), 3.into()).unwrap().into()]).unwrap()]),
+    TypeDefinitionKind::anonymous([Constraint::valid_values(vec![2.into(), ion_rs::Decimal::new(35, -1).into(), 5e7.into(), "hello".to_owned().into(), Symbol::from("hi").into(), NumberRange::new_inclusive(2.into(), 3.into()).unwrap().into()], IslVersion::V1_0).unwrap(), Constraint::type_constraint(34)])
+    ),
```
zslayton: As a library `TODO` (if/when we finish the versioning-via-generics refactoring), we should look at making constraint macros that coalesce failure modes and elide all of the `.into()`s that are currently required at the call site, turning this:

```rust
valid_values(
  vec![
    2.into(), 
    ion_rs::Decimal::new(35, -1).into(),
    5e7.into(),
    "hello".to_owned().into(),
    Symbol::from("hi").into(), 
    NumberRange::new_inclusive(2.into(), 3.into()).unwrap().into()
  ]
)
```
into something more like:
```rust
valid_values! [
  2,
  Decimal::new(35, -1), 
  5e7,
  "hello",
  Symbol::from("hi")
  inclusive_number_range! [2, 3]
],
```

zslayton (approved):

ion-schema/src/isl/ranges.rs
popematt: Oops... too much python recently. I'll fix it.

ion-schema/src/types.rs
popematt: Created #195 to track that.

ion-schema/src/isl/ranges.rs
desaikd: I still think we should add some comments about what `Unbounded` represents here and when to use it? I see that you have sued it to represent `min`/`max` but it would be helpful to add some doc comments for it so the usecase is clear.

ion-schema/src/isl/ranges.rs
```diff
@@ -0,0 +1,502 @@
...
+            isl_require!(!Self::is_empty(&start, &end) => "")?;
+            Ok(Self {
+                lower: start,
+                upper: end,
+            })
+        }
+
+        /// Creates a new range with inclusive endpoints.
+        /// [start] must be less than or equal to [end].
+        pub fn new_inclusive(start: T, end: T) -> IonSchemaResult<Self> {
```
desaikd, line 268: (Suggestion) I should have probably brought this up in the first revision but should we name this `new_*()` methods without the new keyword. e.g. (`new_inclusive()` -> `inclusive()`)? For constraints/types I have used similar convention (e.g. `IslConstraint::byte_length()`, `IslType::named()`). Thoughts?

desaikd (approved):

ion-schema/src/isl/ranges.rs
popematt: 😅 I actually think that `IslConstraint::byte_length()` should cease to exist. Instead, to get a byte length variant of `IslConstraint`, you should use `IslConstraint::ByteLength(ByteLength::new(...))` or `ByteLength::new(...).into()`. Same thing with `IslType`.

That being said, if you feel strongly about getting rid of the `new_` prefix, I will do it. I just figured it would be a helpful way to signal that it is a constructor.

desaikd: I was trying to follow this [design pattern example](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html)(although this is for a builder pattern ): 
```rust
impl FooBuilder {
    pub fn new(/* ... */) -> FooBuilder {
        // Set the minimally required fields of Foo.
        FooBuilder {
            ...
        }
    }

    pub fn name(mut self, bar: String) -> FooBuilder {
        // Set the name on the builder itself, and return the builder by value.
        ...
    }
}
```
I don't think this needs to block the PR though, but I do think whatever we decide to do it should be consistent overall. I have created [this issue](https://github.com/amazon-ion/ion-schema-rust/issues/197) to make changes for it according to what we decide.

PR #191, opened 2023-06-28
Streamlines ISL model using version maker traits 

desaikd:
### Description of changes:
This PR works on streamlining the ISL model using version maker traits. 

### List of changes:
* adds version marker traits which will be used to construct schema,
types and constraints for a particular ISL version
* adds logic constraints implementations
* adds type reference, type and import implementations based on version
marker trait
* adds example usage of this new change for the ISL model

_Note for reviewers_: 
- This PR creates a new module `version_based_isl` which will eventually be placed as the current `isl` module. This is because the change is too big and involves changing mostly entire ISL model. In order to divide the change into pieces created `version_based_isl`. 
- All the structs/enums for ISL model are same as previous ISL model except they now accept a generic `V`(version marker trait). 

----

### DIFF
--- a/ion-schema/src/isl/version_based_isl/constraints/all_of.rs
+++ b/ion-schema/src/isl/version_based_isl/constraints/all_of.rs
@@ -10,7 +10,7 @@ impl<V: IslVersionTrait> AllOf<V> {
         Self { type_args }
     }
 
-    pub fn type_args(&self) -> &Vec<IslTypeArgument<V>> {
+    pub fn type_args(&self) -> &[IslTypeArgument<V>] {
         &self.type_args
     }
 }
--- a/ion-schema/src/isl/version_based_isl/constraints/all_of.rs
+++ b/ion-schema/src/isl/version_based_isl/constraints/all_of.rs
@@ -1,16 +1,16 @@
-use crate::isl::version_based_isl::{IslTypeRef, IslVersionTrait};
+use crate::isl::version_based_isl::{IslTypeArgument, IslVersionTrait};
 
 #[derive(Debug, Clone, PartialEq)]
 pub struct AllOf<V: IslVersionTrait> {
-    type_refs: Vec<IslTypeRef<V>>,
+    type_args: Vec<IslTypeArgument<V>>,
 }
 
 impl<V: IslVersionTrait> AllOf<V> {
-    pub fn new(type_refs: Vec<IslTypeRef<V>>) -> Self {
-        Self { type_refs }
+    pub fn new(type_args: Vec<IslTypeArgument<V>>) -> Self {
+        Self { type_args }
     }
 
-    pub fn type_refs(&self) -> &Vec<IslTypeRef<V>> {
-        &self.type_refs
+    pub fn type_args(&self) -> &Vec<IslTypeArgument<V>> {
+        &self.type_args
     }
 }
--- a/ion-schema/src/isl/version_based_isl/constraints/all_of.rs
+++ b/ion-schema/src/isl/version_based_isl/constraints/all_of.rs
@@ -1,0 +1,16 @@
+use crate::isl::version_based_isl::{IslTypeRef, IslVersionTrait};
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct AllOf<V: IslVersionTrait> {
+    type_refs: Vec<IslTypeRef<V>>,
+}
+
+impl<V: IslVersionTrait> AllOf<V> {
+    pub fn new(type_refs: Vec<IslTypeRef<V>>) -> Self {
+        Self { type_refs }
+    }
+
+    pub fn type_refs(&self) -> &Vec<IslTypeRef<V>> {
+        &self.type_refs
+    }
+}
--- a/ion-schema/src/isl/version_based_isl/constraints/any_of.rs
+++ b/ion-schema/src/isl/version_based_isl/constraints/any_of.rs
@@ -1,0 +1,16 @@
+use crate::isl::version_based_isl::{IslTypeRef, IslVersionTrait};
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct AnyOf<V: IslVersionTrait> {
+    type_refs: Vec<IslTypeRef<V>>,
+}
+
+impl<V: IslVersionTrait> AnyOf<V> {
+    pub fn new(type_refs: Vec<IslTypeRef<V>>) -> Self {
+        Self { type_refs }
+    }
+
+    pub fn type_refs(&self) -> &Vec<IslTypeRef<V>> {
+        &self.type_refs
+    }
+}
--- a/ion-schema/src/isl/version_based_isl/constraints/any_of.rs
+++ b/ion-schema/src/isl/version_based_isl/constraints/any_of.rs
@@ -10,7 +10,7 @@ impl<V: IslVersionTrait> AnyOf<V> {
         Self { type_args }
     }
 
-    pub fn type_args(&self) -> &Vec<IslTypeArgument<V>> {
+    pub fn type_args(&self) -> &[IslTypeArgument<V>] {
         &self.type_args
     }
 }
--- a/ion-schema/src/isl/version_based_isl/constraints/any_of.rs
+++ b/ion-schema/src/isl/version_based_isl/constraints/any_of.rs
@@ -1,16 +1,16 @@
-use crate::isl::version_based_isl::{IslTypeRef, IslVersionTrait};
+use crate::isl::version_based_isl::{IslTypeArgument, IslVersionTrait};
 
 #[derive(Debug, Clone, PartialEq)]
 pub struct AnyOf<V: IslVersionTrait> {
-    type_refs: Vec<IslTypeRef<V>>,
+    type_args: Vec<IslTypeArgument<V>>,
 }
 
 impl<V: IslVersionTrait> AnyOf<V> {
-    pub fn new(type_refs: Vec<IslTypeRef<V>>) -> Self {
-        Self { type_refs }
+    pub fn new(type_args: Vec<IslTypeArgument<V>>) -> Self {
+        Self { type_args }
     }
 
-    pub fn type_refs(&self) -> &Vec<IslTypeRef<V>> {
-        &self.type_refs
+    pub fn type_args(&self) -> &Vec<IslTypeArgument<V>> {
+        &self.type_args
     }
 }
--- a/ion-schema/src/isl/version_based_isl/constraints/mod.rs
+++ b/ion-schema/src/isl/version_based_isl/constraints/mod.rs
@@ -21,7 +21,7 @@ pub enum IslConstraint<V: IslVersionTrait> {
     OneOf(OneOf<V>),
     Not(Not<V>),
     Type(Type<V>),
-    Unknown(OpenContent),
+    OpenContent(OpenContent),
 }
 
 impl<V: IslVersionTrait> IslConstraint<V> {
@@ -51,9 +51,9 @@ impl<V: IslVersionTrait> IslConstraint<V> {
         IslConstraint::Not(Not::new(isl_type_ref))
     }
 
-    /// Creates an `open_content` using the field name and value referenced inside it
+    /// Creates open content using the field name and value referenced inside it
     /// Note: This open content has no effect
     pub fn open_content(field_name: String, field_value: Element) -> IslConstraint<V> {
-        IslConstraint::Unknown(OpenContent::new(field_name, field_value))
+        IslConstraint::OpenContent(OpenContent::new(field_name, field_value))
     }
 }
--- a/ion-schema/src/isl/version_based_isl/constraints/mod.rs
+++ b/ion-schema/src/isl/version_based_isl/constraints/mod.rs
@@ -4,7 +4,7 @@ use crate::isl::version_based_isl::constraints::not::Not;
 use crate::isl::version_based_isl::constraints::one_of::OneOf;
 use crate::isl::version_based_isl::constraints::open_content::OpenContent;
 use crate::isl::version_based_isl::constraints::r#type::Type;
-use crate::isl::version_based_isl::{IslTypeRef, IslVersionTrait};
+use crate::isl::version_based_isl::{IslTypeArgument, IslVersionTrait};
 use ion_rs::element::Element;
 
 pub mod all_of;
@@ -21,33 +21,33 @@ pub enum IslConstraint<V: IslVersionTrait> {
     OneOf(OneOf<V>),
     Not(Not<V>),
     Type(Type<V>),
-    Unknown(OpenContent<V>),
+    Unknown(OpenContent),
 }
 
 impl<V: IslVersionTrait> IslConstraint<V> {
-    /// Creates a `type` constraint using the [IslTypeRef] referenced inside it
+    /// Creates a `type` constraint using the [IslTypeArgument] referenced inside it
     // type is rust keyword hence this method is named type_constraint unlike other ISL constraint methods
-    pub fn type_constraint(isl_type_ref: IslTypeRef<V>) -> IslConstraint<V> {
+    pub fn type_constraint(isl_type_ref: IslTypeArgument<V>) -> IslConstraint<V> {
         IslConstraint::Type(Type::new(isl_type_ref))
     }
 
-    /// Creates an `all_of` constraint using the [IslTypeRef] referenced inside it
-    pub fn all_of<A: Into<Vec<IslTypeRef<V>>>>(isl_type_refs: A) -> IslConstraint<V> {
+    /// Creates an `all_of` constraint using the [IslTypeArgument] referenced inside it
+    pub fn all_of<A: Into<Vec<IslTypeArgument<V>>>>(isl_type_refs: A) -> IslConstraint<V> {
         IslConstraint::AllOf(AllOf::new(isl_type_refs.into()))
     }
 
-    /// Creates an `any_of` constraint using the [IslTypeRef] referenced inside it
-    pub fn any_of<A: Into<Vec<IslTypeRef<V>>>>(isl_type_refs: A) -> IslConstraint<V> {
+    /// Creates an `any_of` constraint using the [IslTypeArgument] referenced inside it
+    pub fn any_of<A: Into<Vec<IslTypeArgument<V>>>>(isl_type_refs: A) -> IslConstraint<V> {
         IslConstraint::AnyOf(AnyOf::new(isl_type_refs.into()))
     }
 
-    /// Creates a `one_of` constraint using the [IslTypeRef] referenced inside it
-    pub fn one_of<A: Into<Vec<IslTypeRef<V>>>>(isl_type_refs: A) -> IslConstraint<V> {
+    /// Creates a `one_of` constraint using the [IslTypeArgument] referenced inside it
+    pub fn one_of<A: Into<Vec<IslTypeArgument<V>>>>(isl_type_refs: A) -> IslConstraint<V> {
         IslConstraint::OneOf(OneOf::new(isl_type_refs.into()))
     }
 
-    /// Creates a `not` constraint using the [IslTypeRef] referenced inside it
-    pub fn not(isl_type_ref: IslTypeRef<V>) -> IslConstraint<V> {
+    /// Creates a `not` constraint using the [IslTypeArgument] referenced inside it
+    pub fn not(isl_type_ref: IslTypeArgument<V>) -> IslConstraint<V> {
         IslConstraint::Not(Not::new(isl_type_ref))
     }
 
--- a/ion-schema/src/isl/version_based_isl/constraints/mod.rs
+++ b/ion-schema/src/isl/version_based_isl/constraints/mod.rs
@@ -1,0 +1,59 @@
+use crate::isl::version_based_isl::constraints::all_of::AllOf;
+use crate::isl::version_based_isl::constraints::any_of::AnyOf;
+use crate::isl::version_based_isl::constraints::not::Not;
+use crate::isl::version_based_isl::constraints::one_of::OneOf;
+use crate::isl::version_based_isl::constraints::open_content::OpenContent;
+use crate::isl::version_based_isl::constraints::r#type::Type;
+use crate::isl::version_based_isl::{IslTypeRef, IslVersionTrait};
+use ion_rs::element::Element;
+
+pub mod all_of;
+pub mod any_of;
+pub mod not;
+pub mod one_of;
+pub mod open_content;
+pub mod r#type;
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum IslConstraint<V: IslVersionTrait> {
+    AllOf(AllOf<V>),
+    AnyOf(AnyOf<V>),
+    OneOf(OneOf<V>),
+    Not(Not<V>),
+    Type(Type<V>),
+    Unknown(OpenContent<V>),
+}
+
+impl<V: IslVersionTrait> IslConstraint<V> {
+    /// Creates a `type` constraint using the [IslTypeRef] referenced inside it
+    // type is rust keyword hence this method is named type_constraint unlike other ISL constraint methods
+    pub fn type_constraint(isl_type_ref: IslTypeRef<V>) -> IslConstraint<V> {
+        IslConstraint::Type(Type::new(isl_type_ref))
+    }
+
+    /// Creates an `all_of` constraint using the [IslTypeRef] referenced inside it
+    pub fn all_of<A: Into<Vec<IslTypeRef<V>>>>(isl_type_refs: A) -> IslConstraint<V> {
+        IslConstraint::AllOf(AllOf::new(isl_type_refs.into()))
+    }
+
+    /// Creates an `any_of` constraint using the [IslTypeRef] referenced inside it
+    pub fn any_of<A: Into<Vec<IslTypeRef<V>>>>(isl_type_refs: A) -> IslConstraint<V> {
+        IslConstraint::AnyOf(AnyOf::new(isl_type_refs.into()))
+    }
+
+    /// Creates a `one_of` constraint using the [IslTypeRef] referenced inside it
+    pub fn one_of<A: Into<Vec<IslTypeRef<V>>>>(isl_type_refs: A) -> IslConstraint<V> {
+        IslConstraint::OneOf(OneOf::new(isl_type_refs.into()))
+    }
+
+    /// Creates a `not` constraint using the [IslTypeRef] referenced inside it
+    pub fn not(isl_type_ref: IslTypeRef<V>) -> IslConstraint<V> {
+        IslConstraint::Not(Not::new(isl_type_ref))
+    }
+
+    /// Creates an `open_content` using the field name and value referenced inside it
+    /// Note: This open content has no effect
+    pub fn open_content(field_name: String, field_value: Element) -> IslConstraint<V> {
+        IslConstraint::Unknown(OpenContent::new(field_name, field_value))
+    }
+}
--- a/ion-schema/src/isl/version_based_isl/constraints/not.rs
+++ b/ion-schema/src/isl/version_based_isl/constraints/not.rs
@@ -1,0 +1,16 @@
+use crate::isl::version_based_isl::{IslTypeRef, IslVersionTrait};
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct Not<V: IslVersionTrait> {
+    type_ref: IslTypeRef<V>,
+}
+
+impl<V: IslVersionTrait> Not<V> {
+    pub fn new(type_ref: IslTypeRef<V>) -> Self {
+        Self { type_ref }
+    }
+
+    pub fn type_ref(&self) -> &IslTypeRef<V> {
+        &self.type_ref
+    }
+}
--- a/ion-schema/src/isl/version_based_isl/constraints/not.rs
+++ b/ion-schema/src/isl/version_based_isl/constraints/not.rs
@@ -1,16 +1,16 @@
-use crate::isl::version_based_isl::{IslTypeRef, IslVersionTrait};
+use crate::isl::version_based_isl::{IslTypeArgument, IslVersionTrait};
 
 #[derive(Debug, Clone, PartialEq)]
 pub struct Not<V: IslVersionTrait> {
-    type_ref: IslTypeRef<V>,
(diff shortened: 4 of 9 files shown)


ion-schema/src/isl/version_based_isl/constraints/all_of.rs
```diff
@@ -0,0 +1,16 @@
+use crate::isl::version_based_isl::{IslTypeRef, IslVersionTrait};
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct AllOf<V: IslVersionTrait> {
```
desaikd, line 4: 🗺️ All the constraints are now placed inside `constraints` module. Each constraint will now accept a generic `V` that represents the ISL version using version marker trait.

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,396 @@
...
+
+#[derive(Debug, Clone, Default, PartialEq)]
+pub struct UserReservedFields<V: IslVersionTrait> {
+    schema_header_fields: Vec<String>,
+    schema_footer_fields: Vec<String>,
+    type_fields: Vec<String>,
+    phantom: PhantomData<V>,
+}
+
+impl UserReservedFields<IslV2_0> {
```
desaikd, line 125: 🗺️  This is only defined for ISL 2.0 and hence would result in a compile time error when user tries to create this with ISL 1.0 schema. Please look at the `IslSchema#with_user_reserved_fields()`.

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,396 @@
...
+    /// Creates an [IslVariablyOccurringTypeRef] using the [IslConstraint]s and [Range] referenced inside it
+    pub fn variably_occurring(
+        type_ref: IslTypeRef<V>,
+        occurs: Range,
+    ) -> IslVariablyOccurringTypeRef<V> {
+        IslVariablyOccurringTypeRef::new(type_ref, occurs)
+    }
+
+    /// Creates a nullable [IslTypeRef] using the name of the type referenced inside it for ISL 2.0
+    pub fn null_or_named_type_ref<A: Into<String>>(name: A) -> IslTypeRef<IslV2_0> {
```
desaikd, line 348: 🗺️ This will only be allowed to use with ISL 2.0 schema. Using it with an ISL 1.0 schema would result in compile time error.
Example:
```rust
 let isl_type = IslType::named(
     // represents the `name` of the defined type
     "my_type_name".to_owned(),
     vec![
         // represents the `type: $null_or::int` constraint
         IslConstraint::type_constraint(
            IslTypeRef::null_or_named_type_ref("int")
        ),
     ]
 );

 // create an ISL schema using above IslType
 let mut  isl_schema = IslSchema::<IslV1_0>::new("my_schema", vec![], vec![isl_type], vec![], vec![]);
```
Output:
```
   | let mut  isl_schema = IslSchema::<IslV1_0>::new("my_schema", vec![], vec![isl_type], vec![], vec![]);
   |                                                                           ^^^^^^^^ expected `IslType<IslV1_0>`, found `IslType<IslV2_0>`
```

popematt (commented):
I have a few high level comments. I'd like to hear your thoughts on these before I dive into the actual code.

* Eventually, we really need to address the fact that some constraints have their own structs, and others are just an enum variant that's a tuple. If you're rewriting the model, this seems like an ideal time to fix it.
* I'd recommend something like this to eliminate some of the boilerplate for constraints. https://gist.github.com/popematt/11b1e632ed487d11ba50eda1bb43a78a (Note that this will not work exactly because it doesn't have the version it it.) As a bonus, a macro like this will mean that we also get a proper struct for each constraint.
* What other solutions did you consider? (Have you considered creating versioned builders for a versionless model?) What made you choose a versioned model over those other solutions? (I'd like to understand any tradeoffs, etc.)

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,396 @@
...
+    /// type::{ name: foo, codepoint_length: 1 }
+    /// ```
+    ///
+    /// ```ion
+    /// type::{ name: foo, codepoint_length: 1 }
+    /// $foo
+    /// $bar
+    /// ```
+    open_content: Vec<Element>,
+    phantom: PhantomData<V>,
```
popematt, line 184: I _think_ it's conventional to use `_` for unused variables/fields, and I think it's a good idea to give the phantom data a name.
```suggestion
    _version: PhantomData<V>,
```

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,396 @@
...
+    /// Provides open content that is there in the type definition
+    pub fn open_content(&self) -> Vec<(&String, &Element)> {
+        let mut open_contents = vec![];
+        for constraint in &self.constraints {
+            if let IslConstraint::Unknown(open_content) = constraint {
+                open_contents.push((open_content.field_name(), open_content.field_value()))
+            }
+        }
+        open_contents
+    }
```
popematt, line 298: I think this is a hint that open content should not be modeled as a constraint. Perhaps something like this instead?

```rust
pub struct IslType<V: IslVersionTrait> {
    name: Option<String>,
    constraints: Vec<IslConstraint<V>>,
    open_content: Vec<OpenContent<V>>,
    pub(crate) isl_type_struct: Option<Element>,
    phantom: PhantomData<V>,
}
```

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,396 @@
...
+
+/// Represents both named and anonymous [IslType]s and can be converted to a resolved type definition
+/// Named ISL type grammar: `type:: { name: <NAME>, <CONSTRAINT>...}`
+/// Anonymous ISL type grammar: `{ <CONSTRAINT>... }`
+#[derive(Debug, Clone, PartialEq)]
+pub struct IslType<V: IslVersionTrait> {
+    name: Option<String>,
+    constraints: Vec<IslConstraint<V>>,
+    // Represents the ISL type struct in string format for anonymous type definition
+    // For named type definition & programmatically created type definition, this will be `None`
```
popematt, line 255:
```suggestion
    /// Represents the ISL type struct in string format for anonymous type definition
    /// For named type definition & programmatically created type definition, this will be `None`
```

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,396 @@
...
+    }
+}
+
+/// Represents both named and anonymous [IslType]s and can be converted to a resolved type definition
+/// Named ISL type grammar: `type:: { name: <NAME>, <CONSTRAINT>...}`
+/// Anonymous ISL type grammar: `{ <CONSTRAINT>... }`
+#[derive(Debug, Clone, PartialEq)]
+pub struct IslType<V: IslVersionTrait> {
+    name: Option<String>,
+    constraints: Vec<IslConstraint<V>>,
```
popematt, line 253: suggestion—I recommend using a Bag data type for this, but I am okay with deferring that for now.

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,396 @@
...
+        &self.constraints
+    }
+}
+
+/// Provides an internal representation of a schema type reference.
+/// The type reference grammar is defined in the [Ion Schema Spec]
+///
+/// [Ion Schema spec]: https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#grammar
+#[derive(Debug, Clone, PartialEq)]
+pub enum IslTypeRef<V: IslVersionTrait> {
```
popematt, line 311:
```suggestion
pub enum IslTypeArgument<V: IslVersionTrait> {
```

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,396 @@
...
+impl<V: IslVersionTrait> IslTypeRef<V> {
+    /// Creates a named [IslTypeRef] using the name of the type referenced inside it
+    pub fn named<A: Into<String>>(name: A) -> IslTypeRef<V> {
+        IslTypeRef::Named(name.into(), NullabilityModifier::Nothing)
+    }
+
+    /// Creates a nullable [IslTypeRef] using the [IonType] referenced inside it
+    pub fn nullable_built_in(name: IonType) -> IslTypeRef<IslV1_0> {
+        IslTypeRef::Named(format!("{name}"), NullabilityModifier::Nullable)
+    }
```
popematt, line 329: This function should be moved to a separate impl block. I.e.:
```rust
impl IslTypeRef<IslV1_0> {
    /// Creates a nullable [IslTypeRef] using the [IonType] referenced inside it
    pub fn nullable_built_in(name: IonType) -> Self {
        IslTypeRef::Named(format!("{name}"), NullabilityModifier::Nullable)
    }
}
```

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,396 @@
...
+    Named(String, NullabilityModifier),
+    /// Represents a type reference defined as an inlined import of a type from another schema
+    TypeImport(IslImportType, NullabilityModifier),
+    /// represents an unnamed type definition reference
+    Anonymous(IslType<V>, NullabilityModifier),
+}
+
+impl<V: IslVersionTrait> IslTypeRef<V> {
+    /// Creates a named [IslTypeRef] using the name of the type referenced inside it
+    pub fn named<A: Into<String>>(name: A) -> IslTypeRef<V> {
```
popematt, line 322:
```suggestion
    pub fn named<A: Into<String>>(name: A) -> Self {
```

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,396 @@
...
+        )
+    }
+
+    /// Creates an [IslVariablyOccurringTypeRef] using the [IslConstraint]s and [Range] referenced inside it
+    pub fn variably_occurring(
+        type_ref: IslTypeRef<V>,
+        occurs: Range,
+    ) -> IslVariablyOccurringTypeRef<V> {
+        IslVariablyOccurringTypeRef::new(type_ref, occurs)
+    }
```
popematt, line 345: I don't think this belongs here because it is a constructor for `IslVariableOccuringTypeRef`.

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,396 @@
...
+        IslTypeRef::Named(name.into(), NullabilityModifier::Nothing)
+    }
+
+    /// Creates a nullable [IslTypeRef] using the [IonType] referenced inside it
+    pub fn nullable_built_in(name: IonType) -> IslTypeRef<IslV1_0> {
+        IslTypeRef::Named(format!("{name}"), NullabilityModifier::Nullable)
+    }
+
+    /// Creates an anonymous [IslTypeRef] using the [IslConstraint]s referenced inside it
+    pub fn anonymous<A: Into<Vec<IslConstraint<V>>>>(constraints: A) -> IslTypeRef<V> {
```
popematt, line 332:
```suggestion
    pub fn anonymous<A: Into<Vec<IslConstraint<V>>>>(constraints: A) -> Self {
```

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,396 @@
...
+///
+///         <OCCURS> ::= occurs: <INT>
+///            | occurs: <RANGE_INT>
+///            | occurs: optional
+///            | occurs: required
+/// ```
+///
+/// [variably occurring type reference]: https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#variably-occurring-type-arguments
+#[derive(Debug, Clone, PartialEq)]
+pub struct IslVariablyOccurringTypeRef<V: IslVersionTrait> {
```
popematt, line 367:
```suggestion
pub struct IslVariablyOccurringTypeArgument<V: IslVersionTrait> {
```

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,396 @@
...
+}
+
+/// Represents an [import] in an ISL schema.
+///
+/// [import]: https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#imports
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum IslImport {
+    Schema(String),
+    Type(IslImportType),
+    TypeAlias(IslImportType),
```
popematt, line 385:
```suggestion
    Type(IslImportType),
```
`TypeAlias` is redundant since both enum variants contain the same thing.

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,396 @@
...
+}
+
+/// Represents typed and type aliased [IslImport]s
+/// Typed import grammar: `{ id: <ID>, type: <TYPE_NAME> }`
+/// Type aliased import grammar: `{ id: <ID>, type: <TYPE_NAME>, as: <TYPE_ALIAS> }`
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct IslImportType {
+    id: String,
```
popematt, line 393: suggestion—slightly more explicit about what sort of ID it is.
```suggestion
    schema_id: String,
```

ion-schema/src/isl/version_based_isl/constraints/open_content.rs
```diff
@@ -0,0 +1,27 @@
+use crate::isl::version_based_isl::IslVersionTrait;
+use ion_rs::element::Element;
+use std::marker::PhantomData;
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct OpenContent<V: IslVersionTrait> {
```
popematt, line 6: question—does `OpenContent` really need to be versioned?

desaikd:
Overall versioned ISL model gives an ease of use, maintainability and gives compile time checks for ISL version on types, constraints and schema.

>    * Eventually, we really need to address the fact that some constraints have their own structs, and others are just an enum variant that's a tuple. If you're rewriting the model, this seems like an ideal time to fix it.

Previous ISL model had tuples for enum variants for all the constraints. I have created a struct for each constraint so that we have the same pattern throughout for all constraints be it a constraint thats same for both ISL versions or its different for different ISL versions. It gives us a nice distinction between how 2 different versions would work for a particular constraint. Also eventually if we decide to have single model that will be used for representation as well as validation then it makes sense to create different structs for all the constraints as each one will have its own validation.

>    * I'd recommend something like this to eliminate some of the boilerplate for constraints. https://gist.github.com/popematt/11b1e632ed487d11ba50eda1bb43a78a (Note that this will not work exactly because it doesn't have the version it it.) As a bonus, a macro like this will mean that we also get a proper struct for each constraint.

Adding a macro to create the structs for constraints does sound like good idea. I could create an issue for it.

>    * What other solutions did you consider? (Have you considered creating versioned builders for a versionless model?) 
What made you choose a versioned model over those other solutions? (I'd like to understand any tradeoffs, etc.)

I have chosen to go with versioned model using the version maker trait because that gives a better API to construct ISL model constraints/types and schema. This allows us to have compile time errors for when a user tries to mix two different versions of schema. I think this provides an ease of use for customers as well as ensures the different version constraints aren't mixed.

ion-schema/src/isl/version_based_isl/constraints/all_of.rs
```diff
@@ -0,0 +1,16 @@
...
+    type_refs: Vec<IslTypeRef<V>>,
```
popematt, line 5: suggestion—this should be a Bag data type, but I'm fine with punting on it for now.

ion-schema/src/isl/version_based_isl/constraints/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+    /// Creates a `not` constraint using the [IslTypeRef] referenced inside it
+    pub fn not(isl_type_ref: IslTypeRef<V>) -> IslConstraint<V> {
+        IslConstraint::Not(Not::new(isl_type_ref))
+    }
+
+    /// Creates an `open_content` using the field name and value referenced inside it
+    /// Note: This open content has no effect
+    pub fn open_content(field_name: String, field_value: Element) -> IslConstraint<V> {
+        IslConstraint::Unknown(OpenContent::new(field_name, field_value))
+    }
```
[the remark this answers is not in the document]
desaikd: Makes sense. 
These builder functions give a less verbose API then using the constructor for each constraint but its not a huge difference overall. 
```rust
IslCosntraint::Type(Type::new(IslTypeRef::named("int"))
            // vs.
IslCosntraint::type_constraint(IslTypeRef::named("int"))
```

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,396 @@
...
+    }
+
+    pub fn type_fields(&self) -> &Vec<String> {
+        &self.type_fields
+    }
+}
+
+/// Provides an internal representation of an schema file
+#[derive(Debug, Clone)]
+pub struct IslSchema<V: IslVersionTrait> {
```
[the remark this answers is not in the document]
desaikd: Good catch! I will add structs for header and footer.

ion-schema/src/isl/version_based_isl/constraints/open_content.rs
desaikd: No, its not required to be versioned.

ion-schema/src/isl/version_based_isl/mod.rs
desaikd: Changed this to be part of `IslVariableOccuringTypeRef`

ion-schema/src/isl/version_based_isl/constraints/any_of.rs
```diff
@@ -0,0 +1,16 @@
...
+pub struct AnyOf<V: IslVersionTrait> {
+    type_args: Vec<IslTypeArgument<V>>,
+}
+
+impl<V: IslVersionTrait> AnyOf<V> {
+    pub fn new(type_args: Vec<IslTypeArgument<V>>) -> Self {
+        Self { type_args }
+    }
+
+    pub fn type_args(&self) -> &Vec<IslTypeArgument<V>> {
```
zslayton, line 13: For all of these of accessor methods, return a slice (`&[IslTypeArgument]`) instead of a reference to the `Vec` to avoid double indirection.

ion-schema/src/isl/version_based_isl/constraints/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+    /// Creates a `not` constraint using the [IslTypeArgument] referenced inside it
+    pub fn not(isl_type_ref: IslTypeArgument<V>) -> IslConstraint<V> {
+        IslConstraint::Not(Not::new(isl_type_ref))
+    }
+
+    /// Creates an `open_content` using the field name and value referenced inside it
+    /// Note: This open content has no effect
+    pub fn open_content(field_name: String, field_value: Element) -> IslConstraint<V> {
+        IslConstraint::Unknown(OpenContent::new(field_name, field_value))
+    }
```
zslayton, line 58: I think there's some potential for confusion around the naming here. I suggest sticking with `OpenContent` for both the enum variant and the nested struct. I think `Unknown` makes it sound like an error even though it's not.

I'd also change the doc comment from "Creates an `open_content`...", which makes it sound like `open_content` is a named entity in the spec (as in the other constraint doc comments), to just "Creates open content using..."

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,480 @@
...
+//!                     vec![
+//!                         IslConstraint::type_constraint(
+//!                             IslTypeArgument::named("bool")
+//!                         )
+//!                     ]
+//!                 )
+//!             ]
+//!         ),
+//!         // represents some open content which will be ignored while doing validation
+//!         IslConstraint::open_content("open_content".to_owned(), Element::string("This is an open content!"))
```
zslayton, line 49:
```suggestion
//!         IslConstraint::open_content("open_content".to_owned(), Element::string("This is open content!"))
```

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,480 @@
...
+//! // below code represents an ISL type:
+//! // type:: {
+//! //      name:my_type_name,
+//! //      type: int,
+//! //      all_of: [
+//! //          { type: bool }
+//! //      ],
+//! //      open_content: "This is an open content!"
+//! //  }
+//! let isl_type: IslType<IslV1_0> = IslType::named(
```
zslayton, line 28: Neat!

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,480 @@
...
+//!     // represents the `name` of the defined type
+//!     "my_type_name".to_owned(),
```
zslayton, line 30: As a `TODO`, all our constructors should accept `impl Into<String>` and similar to avoid requiring boilerplate at the callsites.

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,480 @@
...
+//!         // represents the `type: int` constraint
+//!         IslConstraint::type_constraint(
+//!             IslTypeArgument::named("int")
+//!         ),
+//!     ]
+//! );
+//!
+//! let mut isl_schema_header =  IslSchemaHeader::<IslV2_0>::new(vec![], vec![]); // no open content or imports are there in schema header
+//! // Performing the following operation with ISL 1.0 schema would result in a compile time error.
+//! isl_schema_header = isl_schema_header.with_user_reserved_fields(vec!["foo".to_string(), "bar".to_string()], vec![], vec![]);
```
zslayton, line 92: I don't think `cargo fmt` will ever change the formatting of doc examples. Consider manually wrapping the text if it gets to be longer than 120 or so columns.

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,480 @@
...
+use std::marker::PhantomData;
+
+pub mod constraints;
+
+pub trait IslVersionTrait {}
+
+pub struct IslV1_0 {}
+impl IslVersionTrait for IslV1_0 {}
+
+// TODO: This can be a trait if we have get a new minor version of ISL
```
zslayton, line 116:
```suggestion
// TODO: This can be a trait if we get a new minor version of ISL
```

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,480 @@
...
+        &self.open_content
+    }
+
+    pub fn imports(&self) -> &[IslImport] {
+        &self.imports
+    }
+}
+
+impl IslSchemaHeader<IslV1_0> {
+    pub fn new(open_content: Vec<Element>, imports: Vec<IslImport>) -> Self {
```
zslayton, line 179: If/when we change the constructors to take generics, we should take any `IntoIterator<Item=Element>` instead of specifically `Vec`. That way users can write `[]` instead of `vec![]`, for example.

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,480 @@
...
+    pub fn with_schema_footer(self, schema_footer: IslSchemaFooter) -> Self {
+        Self {
+            schema_footer: Some(schema_footer),
+            ..self
+        }
+    }
+
+    pub fn id(&self) -> String {
+        self.id.to_owned()
+    }
```
zslayton, line 302: Let the caller decide whether the allocation/copy is necessary.
```suggestion
    pub fn id(&self) -> &str {
        &self.id
    }
```

ion-schema/src/isl/version_based_isl/mod.rs
zslayton: Just curious, do we already have a `Bag` type implemented somewhere? Or did you have a 3P impl in mind?

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,480 @@
...
+    /// Creates an anonymous [IslType] using the [IslConstraint] defined within it
+    pub fn anonymous<A: Into<Vec<IslConstraint<V>>>>(constraints: A) -> IslType<V> {
+        IslType::new(None, constraints.into())
+    }
+
+    /// Provides a name if the ISL type is named type definition
+    /// Otherwise returns None
+    pub fn name(&self) -> &Option<String> {
+        &self.name
+    }
```
zslayton, line 368:
```suggestion
    pub fn name(&self) -> Option<&str> {
        self.name.as_ref()
    }
```

ion-schema/src/isl/version_based_isl/mod.rs
zslayton: +1, I encountered some mental dissonance when I found `OpenContent` in the `IslConstraint` enum earlier in my review. Separating it would be helpful.

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,480 @@
...
+            open_content,
+            imports,
+            user_reserved_fields: UserReservedFields::default(),
+            _version: Default::default(),
+        }
+    }
+}
+
+impl IslSchemaHeader<IslV2_0> {
+    pub fn new(open_content: Vec<Element>, imports: Vec<IslImport>) -> Self {
```
zslayton, line 190: Since it'll be relatively rare for users to add open content, I think we should make it an optional builder-style method that can be called after the constructor.
```rust
// Before
IslSchemaHeader::new(vec![], vec![])
```

```rust
IslSchemaHeader::new(vec![])
    // Optional
    .with_open_content(vec![])
```
`Vec::new()` on its own doesn't allocate, so I think having `IslSchemaHeader::new()` create an empty `Vec` that `with_open_content` can replace is fine.

We can apply this pattern to any of the types that have optional constructor arguments.

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,480 @@
...
+            _version: Default::default(),
+        }
+    }
+
+    pub fn with_user_reserved_fields(
+        self,
+        schema_header_fields: Vec<String>,
+        schema_footer_fields: Vec<String>,
+        type_fields: Vec<String>,
+    ) -> Self {
```
zslayton, line 204: At a glance, I expected this to be two methods (`with_user_reserved_header_fields()` and `with_user_reserved_footer_fields()`). I'd only keep them as a unified method if it's illegal to add one without the other.

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,480 @@
...
+    /// Represents the schema footer if its present which contains any open content provided in the footer.
+    /// If the schema footer is not present this is None.
+    schema_footer: Option<IslSchemaFooter>,
+    /// Represents all the IslType defined in this schema file.
+    /// For more information: https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#type-definitions
+    types: Vec<IslType<V>>,
+    /// Represents all the inline IslImportTypes in this schema file.
+    inline_imported_types: Vec<IslImportType>,
+    /// Represents open content as `Element`s
+    /// Note: This doesn't preserve the information about where the open content is placed in the schema file.
```
zslayton, line 251: This is reasonable, but has interesting effects on how we serialize a schema later. We should talk through the impact it'll have. No `TODO` here.

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,480 @@
...
+    /// For open content defined within type definitions use IslType#open_content()
+    pub fn open_content(&self) -> &Vec<Element> {
+        &self.open_content
+    }
+
+    /// Provides the schema header value if present,
+    /// Otherwise returns None.
+    pub fn schema_header(&self) -> &Option<IslSchemaHeader<V>> {
+        &self.schema_header
+    }
```
zslayton, line 322: If you return a `&Option<_>`, the user can't `unwrap()` it because the `Option` itself is borrowed. Use `Option::as_ref()` to get an `Option<&_>` instead.
```suggestion
    pub fn schema_header(&self) -> Option<&IslSchemaHeader<V>> {
        self.schema_header.as_ref()
    }
```

ion-schema/src/isl/version_based_isl/mod.rs
```diff
@@ -0,0 +1,480 @@
...
+
+    /// Provides the schema footer value if present,
+    /// Otherwise returns None.
+    pub fn schema_footer(&self) -> &Option<IslSchemaFooter> {
+        &self.schema_footer
+    }
```
zslayton, line 328:
```suggestion
    pub fn schema_footer(&self) -> Option<&IslSchemaFooter> {
        self.schema_footer.as_ref()
    }
```

ion-schema/src/isl/version_based_isl/constraints/not.rs
```diff
@@ -0,0 +1,16 @@
+use crate::isl::version_based_isl::{IslTypeArgument, IslVersionTrait};
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct Not<V: IslVersionTrait> {
```
popematt, line 4: Add `impl <V: IslVersionTrait> From<Not<V>> for IslConstraint<V>`, and similarly for all other constraints.

ion-schema/src/isl/version_based_isl/mod.rs
popematt: We don't have a bag type. There are plenty of 3P bag/multiset types available, but I don't have a particular one in mind. Either way, we need it so that we have proper equality since the order of the constraints is unimportant.

If open content was modeled separately from the constraints, we could use a set instead of a bag since two completely identical constraints are redundant and functionally equivalent to having only one of them.

popematt (approved):
I am unsure about the maintainability of having a separate file for each individual constraint—I suspect that we may end up repeating a lot of very similar code across all of the files—but it's better than having a single file with thousands of lines of code.

I did leave a comment about adding trivial `From` implementations to convert from the inner types of the enum variants to the enum type for `IslConstraint`. I think it should be done, but it doesn't need to be done immediately.

Overall though, this is looking good.

PR #200, opened 2023-09-19
adds ion-schema-tests and related fixes for ISL 2.0

desaikd:
### Description of changes:
This PR adds from remaining ion-schema-tests for ISL 2.0 and related bug fixes.

----

### DIFF
--- a/ion-schema/src/isl/isl_constraint.rs
+++ b/ion-schema/src/isl/isl_constraint.rs
@@ -584,7 +584,7 @@ impl IslConstraintImpl {
                 }
 
                 if value.annotations().contains("range") {
-                    return invalid_schema_error("contains constraint can not have ranges");
+                    return invalid_schema_error("contains list can not have any annotations");
                 }
 
                 let values: Vec<Element> = value
--- a/ion-schema/src/isl/isl_constraint.rs
+++ b/ion-schema/src/isl/isl_constraint.rs
@@ -584,9 +584,7 @@ impl IslConstraintImpl {
                 }
 
                 if value.annotations().contains("range") {
-                    return invalid_schema_error(format!(
-                        "contains constraint can not have ranges"
-                    ));
+                    return invalid_schema_error("contains constraint can not have ranges");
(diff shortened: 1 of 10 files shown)


ion-schema/src/isl/isl_constraint.rs
```diff
@@ -583,6 +583,10 @@ impl IslConstraintImpl {
                     ));
                 }
 
+                if value.annotations().contains("range") {
+                    return invalid_schema_error("contains constraint can not have ranges");
+                }
```
popematt, line 588: You could generalize this. In ISL 2.0, the `contains` list cannot have any annotations.

ion-schema/src/isl/ranges.rs
```diff
@@ -132,7 +132,51 @@ pub type TimestampRange = base::Range<Timestamp>;
...
+                    Nanosecond => 9,
+                    OtherFractionalSeconds(scale) => *scale,
+                };
+
+                // Checking for e.g. range::[exclusive::1, exclusive::2] which is empty.
+                let adjusted_lower = start_value.checked_add(1);
+                // If the _lower_ bound wraps around when we add one, then we know it's empty.
+                if adjusted_lower.is_none() {
+                    return true;
+                }
```
popematt, line 173: I'm not sure I understand this. Why would the lower bound wrap around by adding one? The lower bound cannot wrap around unless it's already the _max_ value, in which case we can detect that it's empty by `lower >= upper`.

desaikd: Thats right, thanks for catching it. I forgot about the `Limit::Unbound` enum variant you added recently.
Since this isn't really a max value doesn't require to check for `is_none` here. Maybe can just be `start_value + 1` in that case.

popematt (approved):
Sorry for the delay. I started a review, and forgot to publish it.

There's one minor change that _needs_ to be made (for which I've added a suggested change), and there's a spot where you _could_ DRY up the code a bit.

PR #208, opened 2024-02-21
Modifies `PartialEq` for `Violation`

desaikd:
### Description of changes:
This PR works on fixing `PartialEq` for `Violation` and modifies `flattened_violations` return sorted violations which can be used for comparison.

### List of changes:
* Changes `PartialEq` for `Violation` to defer to flattened violations
for equivalence rather then `Violation`s tree
* Modified `flattened_violations` to return sorted vector absed on
constraint
* Modified `flattened_violations` to not return the violation itself
when there are no children
* adds clippy and rustfmt changes

### Test: 
Adds test for violations equivalence.

----

### DIFF
--- a/ion-schema/src/violation.rs
+++ b/ion-schema/src/violation.rs
@@ -92,6 +92,7 @@ impl PartialEq for Violation {
         self.constraint == other.constraint
             && self.code == other.code
             && self.message == other.message
+            && self.ion_path == other.ion_path
             && self.flattened_violations() == other.flattened_violations()
     }
 }
@@ -156,12 +157,12 @@ impl fmt::Display for ViolationCode {
 
 #[cfg(test)]
 mod violation_tests {
-    use crate::ion_path::IonPath;
+    use crate::ion_path::{IonPath, IonPathElement};
     use crate::violation::{Violation, ViolationCode};
+    use rstest::rstest;
 
-    #[test]
-    fn violation_equivalence() {
-        let violation1: Violation = Violation::with_violations(
+    #[rstest(violation1, violation2,
+    case::unordered_violations(Violation::with_violations(
             "type_constraint",
             ViolationCode::TypeMismatched,
             "type mismatched",
@@ -180,8 +181,7 @@ mod violation_tests {
                     &mut IonPath::default(),
                 ),
             ],
-        );
-        let violation2: Violation = Violation::with_violations(
+        ), Violation::with_violations(
             "type_constraint",
             ViolationCode::TypeMismatched,
             "type mismatched",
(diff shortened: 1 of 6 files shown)


ion-schema/src/violation.rs
```diff
@@ -146,3 +153,59 @@ impl fmt::Display for ViolationCode {
...
+                ),
+                Violation::new(
+                    "regex",
+                    ViolationCode::RegexMismatched,
+                    "regex mismatched",
+                    &mut IonPath::default(),
+                ),
+            ],
+        );
+        assert_ne!(violation1.violations(), violation2.violations());
```
popematt, line 204: question—why are we asserting that this are not equal? This seems like it's not relevant to the test.

ion-schema/src/violation.rs
```diff
@@ -146,3 +153,59 @@ impl fmt::Display for ViolationCode {
...
     }
 }
+
+#[cfg(test)]
+mod violation_tests {
+    use crate::ion_path::IonPath;
+    use crate::violation::{Violation, ViolationCode};
+
+    #[test]
+    fn violation_equivalence() {
```
popematt, line 163: Can you add another test case with multiple levels of nesting?

ion-schema/src/violation.rs
desaikd: Err: this was part of debugging. I will remove this.

ion-schema/src/violation.rs
desaikd: Sure. I will add that.

desaikd: `PartialEq` had missing `ion_path` equivalence check. Added a condition check for that along with more tests.

ion-schema/src/violation.rs
```diff
@@ -61,13 +61,11 @@ impl Violation {
...
         let mut flattened_violations = Vec::new();
         self.flatten_violations(&mut flattened_violations);
+        flattened_violations.sort_by(|a, b| (a.constraint).cmp(&b.constraint));
         flattened_violations
     }
 
     fn flatten_violations<'a>(&'a self, flattened: &mut Vec<&'a Violation>) {
-        if self.violations.is_empty() {
-            flattened.push(self);
-        }
```
popematt, line 70: I'm not sure if this is correct. If there is any violation, and you flatten them, you should always get a non-empty list of violations, right?

ion-schema/src/violation.rs
```diff
@@ -61,13 +61,11 @@ impl Violation {
     pub fn flattened_violations(&self) -> Vec<&Violation> {
         let mut flattened_violations = Vec::new();
         self.flatten_violations(&mut flattened_violations);
+        flattened_violations.sort_by(|a, b| (a.constraint).cmp(&b.constraint));
```
popematt, line 64: I think the fields used for sorting needs to be consistent with the fields used for equality, or else we could end up with weird bugs.

ion-schema/src/violation.rs
desaikd: That seems right. I guess I was thinking more about returning child violations for the current violation. More like a flattened version of `violations()` method which returns a violation tree.

ion-schema/src/violation.rs
desaikd: So equality uses flattened_violations itself. If we use it for sorting as well it would end up in infinite recursive calls.

ion-schema/src/violation.rs
```diff
@@ -146,3 +154,217 @@ impl fmt::Display for ViolationCode {
...
+            vec![],
+        ),
+        Violation::with_violations(
+            "type_constraint",
+            ViolationCode::TypeMismatched,
+            "type mismatched",
+            &mut IonPath::default(),
+            vec![],
+        )
+    ))]
```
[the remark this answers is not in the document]
desaikd: `case::unordered_violations()` already has nested violations with multiple violations.

ion-schema/src/violation.rs
popematt: Sure, but I think we could get away with not sorting by flattened violations because if you flatten it once, all of the returned violations have no children (and flattening a childless violation is equivalent to an identity function).

All of the other fields need to be included in the sort.

ion-schema/src/violation.rs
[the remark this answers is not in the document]
popematt: These ones are different though. They have violations with the same constraint name. This isn't guaranteed to pass unless we have a more complete sorting of the flattened violations.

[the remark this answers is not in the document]
desaikd: I see. This is based on adding more checks for sorting. Will add that.

ion-schema/src/violation.rs
desaikd: Returning a `HashSet` instead.

ion-schema/src/violation.rs
desaikd: Changed this behavior back to return the current violation as well. `flattened_violations` will always return atleast 1 violation.

ion-schema/src/violation.rs
```diff
@@ -146,3 +154,217 @@ impl fmt::Display for ViolationCode {
...
+            ViolationCode::TypeMismatched,
+            "type mismatched",
+            &mut IonPath::default(),
+            vec![
+                Violation::with_violations(
+                    "regex",
+                    ViolationCode::RegexMismatched,
+                    "regex mismatched",
+                    &mut IonPath::default(),
+                    vec![
```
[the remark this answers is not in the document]
desaikd: Modified `case::nested_violations` to include multiple levels of nesting for unordered violations.

ion-schema/src/violation.rs
```diff
@@ -93,12 +97,29 @@ impl PartialEq for Violation {
...
+
+impl Hash for Violation {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.code.hash(state);
+        self.constraint.hash(state);
+        self.message.hash(state);
+        self.ion_path.hash(state);
+        for violation in self.violations.iter() {
+            violation.hash(state);
+        }
```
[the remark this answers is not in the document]
desaikd: Removed `PartialEq`, added this macro and document comments on how to use these macro for equivalence.

wasm-schema-sandbox/src/lib.rs
```diff
@@ -193,9 +194,9 @@ pub fn validate(
 
     log!("validation complete!");
 
-    let violations = match &result {
+    let violations: Vec<&Violation> = match &result {
         Ok(_) => vec![],
-        Err(violation) => violation.flattened_violations(),
+        Err(violation) => violation.flattened_violations().into_iter().collect(),
```
popematt, line 199: Doesn't `flattened_violations()` already return a `Vec<&Violation>`?
```suggestion
        Err(violation) => violation.flattened_violations(),
```

ion-schema/src/violation.rs
```diff
@@ -146,3 +150,336 @@ impl fmt::Display for ViolationCode {
         )
     }
 }
+
+#[macro_export]
+macro_rules! assert_equivalent_violations {
```
popematt, line 155: Please add a brief doc comment for this and for `assert_non_equivalent_violations`.

popematt (approved):
Two minor comments. Otherwise it looks good.

PR #215, opened 2024-08-13
Adds support for `get_imported_type`, `get_type` and `get_built_in_or_defined_type`

desaikd:
### Description of changes:
This PR adds support for `get_imported_type`, `get_type` and `get_built_in_or_defined_type`.

### List of changes:
- Adds `get_imported_type` API which can help get a particular type that is being imported to given schema
- Renames previous `get_type` -> `get_built_in_or_defined_type` as it only provides types that are defined within the schema or a built in type.
- Adds `get_type` which

----

### DIFF
--- a/ion-schema-tests-runner/src/generator.rs
+++ b/ion-schema-tests-runner/src/generator.rs
@@ -321,7 +321,7 @@ fn generate_preamble(root_dir_path: &Path) -> TokenStream {
         /// Asserts that a value is or is not valid for a given ISL type.
         fn __assert_value_validity_for_type(value_ion: &str, schema_id: &str, type_id: &str, expect_valid: bool) -> Result<(), String> {
             let schema = __new_schema_system().load_schema(schema_id).unwrap();
-            let isl_type = schema.get_local_or_imported_type(type_id).unwrap();
+            let isl_type = schema.get_type(type_id).unwrap();
             let value: ion_rs::element::Element = ion_rs::element::Element::read_one(value_ion.as_bytes()).unwrap();
             let prepared_value: ion_schema::IonSchemaElement = if value.annotations().contains("document") && value.ion_type() == ion_rs::IonType::SExp {
                 let element_vec = value.as_sequence()
--- a/ion-schema-tests-runner/src/generator.rs
+++ b/ion-schema-tests-runner/src/generator.rs
@@ -321,7 +321,7 @@ fn generate_preamble(root_dir_path: &Path) -> TokenStream {
         /// Asserts that a value is or is not valid for a given ISL type.
         fn __assert_value_validity_for_type(value_ion: &str, schema_id: &str, type_id: &str, expect_valid: bool) -> Result<(), String> {
             let schema = __new_schema_system().load_schema(schema_id).unwrap();
-            let isl_type = schema.get_type(type_id).unwrap();
+            let isl_type = schema.get_local_or_imported_type(type_id).unwrap();
             let value: ion_rs::element::Element = ion_rs::element::Element::read_one(value_ion.as_bytes()).unwrap();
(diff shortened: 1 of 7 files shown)


ion-schema/src/schema.rs
```diff
@@ -53,15 +53,29 @@ impl Schema {
...
+    pub fn get_built_in_or_defined_type<A: AsRef<str>>(&self, name: A) -> Option<TypeDefinition> {
         let type_id = self
             .types
             .get_built_in_type_id_or_defined_type_id_by_name(name.as_ref())?;
         Some(TypeDefinition::new(*type_id, Arc::clone(&self.types)))
     }
 
+    /// Returns the requested type, if present in this schema or a a built in type or imported in the schema;
+    /// otherwise returns None.
+    pub fn get_type<A: AsRef<str>>(&self, name: A) -> Option<TypeDefinition> {
```
popematt, line 74: I know that the library is still considered unstable, but it is not a great idea to re-use a function name to mean something different.

Perhaps keep the existing `get_type` as is, and call this function `get_local_or_imported_type()` (built-in types are implicitly imported).

ion-schema/src/schema.rs
```diff
@@ -53,15 +53,29 @@ impl Schema {
         SchemaTypeIterator::new(Arc::clone(&self.types), self.types.get_imports())
     }
 
+    /// Returns the requested type, if imported in this schema;
+    /// otherwise returns None.
```
popematt, line 57: Documentation should mention that this includes built-in types.

ion-schema/src/system.rs
```diff
@@ -543,6 +543,14 @@ impl TypeStore {
         self.imported_type_ids_by_name.values().cloned().collect()
     }
 
+    /// Provides the [`TypeId`] associated with given name if it exists in the [`TypeStore`] as an imported type;
+    /// Otherwise returns None
```
popematt, line 547: Comment should mention built in types.

popematt (commented):
Few minor comments.

On a more general note, if there's a big formatting change like this, can you please put that into a separate commit in the same PR (or a different PR) so that I can look at them separately?

ion-schema/src/schema.rs
desaikd: I think this should only include the imported types for built-in types there is already `get_local_or_imported_type` API. This is just to keep the imports separate. I have modified the logic for it in the `TypeStore`.

ion-schema/src/system.rs
desaikd: I have modified this method to only include imported types. For built-in types, there is already `get_type_id_by_name` API available.

ion-schema/src/schema.rs
desaikd: Makes sense. Will change this accordingly.

ion-schema-tests-runner/src/generator.rs
```diff
@@ -321,7 +321,7 @@ fn generate_preamble(root_dir_path: &Path) -> TokenStream {
         /// Asserts that a value is or is not valid for a given ISL type.
         fn __assert_value_validity_for_type(value_ion: &str, schema_id: &str, type_id: &str, expect_valid: bool) -> Result<(), String> {
             let schema = __new_schema_system().load_schema(schema_id).unwrap();
-            let isl_type = schema.get_type(type_id).unwrap();
+            let isl_type = schema.get_local_or_imported_type(type_id).unwrap();
```
popematt, line 324: Is it _necessary_ to change this name? Is the benefit worth it for the churn that users would have to deal with?

popematt (approved):

ion-schema-tests-runner/src/generator.rs
desaikd: I think we can just keep the name same for now(i.e. `get_type`). The method will also return imported type now, which seems to be reasonable. User can rely on `get_built_in_or_defined_type` for previous behavior.

Issue #137, opened 2023-01-13
`ion_schema` should not re-export `ion_rs`


popematt:
**The current situation:**

If you're already using `ion_rs`, and then you want to start using `ion_schema`, you may have to change a lot of your imports from `ion_rs::` to `ion_schema::external::ion_rs::`. As long as the versions are the same, this should be a fairly mechanical change that an IDE can handle for you. If the versions are different, then it may be more complex to switch.

In addition, whenever there is a new version of `ion_rs`, and you are using `ion_schema`, you have to wait for `ion_schema` to pick up the latest version of `ion_rs` and release a new version of its own before you can get the benefits of the latest version of `ion_rs`.

**Some alternatives to consider:**

These are just ideas. I'm not sure of the feasibility of all of them.

* Stop exporting `ion_rs` in `ion_schema` and pin to a range of acceptable versions
* ~~Make `ion_rs` export a gated feature in `ion_schema`~~
* Make `ion_schema` a feature in the `ion_rs` crate

This might not be a possibility while both of the libraries are at a `0.x` version, but I think it is something to re-evaluate before releasing a `1.0.0` version of either one of them.

dlurton:
+1 for dropping the re-export of `ion_rs`.  I think that the use of `ion_schema` in applications that don't also directly use `ion_rs` would be rare.  Please don't lock us into a specific version of `ion_rs` and require us to use `ion_schema__external::ion_rs`  just so can also use `ion_schema`.

popematt:
@dlurton do you care to opine on any of the alternatives I've suggested or any ideas you might have?

dlurton:
-1 for: "Make ion_rs export a gated feature in ion_schema" -- this doesn't seem to help the issue at all.

Between the other two, "Stop exporting ion_rs in ion_schema", "Make ion_schema a feature in the ion_rs crate"--either of these I think would be a lot better and it's just a matter of packaging.

It's not good to force the application that wants to use `ion-schema-rust` to use `ion_schema::external::ion_rs`, but it's kind of required right now and leads to one of two somewhat unhappy scenarios when reading Ion from external sources:

[...]

PR #220, opened 2024-10-07
Update to ion-rs@v1.0.0.rc-7 and remove re-export of ion-rs types

popematt:
### Issue #, if available:

#137 

### Description of changes:

* Updates to `ion-rs@v1.0.0-rc-7`
* This update broke the implementations of `WriteToIsl`, so I removed the `WriteToIsl` trait in favor of implementing `WriteAsIon`
* Updates `ion-schema-tests` to the latest commit. (Required because `ion-rs` supports only up to u128 for integers.)
* Minor refactoring in `ranges.rs` to use `Min` and `Max` instead of `Unbounded`. That made it cleaner to implement `WriteAsIon`.

----


ion-schema/src/ion_extension.rs
```diff
@@ -19,25 +18,21 @@ pub(crate) trait ElementExtensions {
 impl ElementExtensions for Element {
     fn as_usize(&self) -> Option<usize> {
         match self.value() {
-            Value::Int(Int::I64(i)) => i.to_usize(),
-            Value::Int(Int::BigInt(i)) => i.to_usize(),
+            Value::Int(i) => i.as_i128()?.to_usize(),
```
zslayton, line 21: There's an [`Int::as_usize()`](https://docs.rs/ion-rs/1.0.0-rc.7/ion_rs/struct.Int.html#method.as_usize) method you could use here.

ion-schema/src/ion_extension.rs
```diff
@@ -19,25 +18,21 @@ pub(crate) trait ElementExtensions {
...
             _ => None,
         }
     }
     fn as_u64(&self) -> Option<u64> {
         match self.value() {
-            Value::Int(Int::I64(i)) => i.to_u64(),
-            Value::Int(Int::BigInt(i)) => i.to_u64(),
+            Value::Int(i) => i.as_i128()?.to_u64(),
```
zslayton, line 27: ...but weirdly, there's _not_ an `Int::as_u64()` method.

ion-schema/src/isl/mod.rs
```diff
@@ -173,48 +183,203 @@ impl Display for IslVersion {
...
+        } else {
+            None
+        }
+    }
+    fn expect_header(&self) -> &SchemaHeader {
+        if let SchemaContent::Header(value) = self {
+            value
+        } else {
+            panic!("illegal state encountered; this should be unreachable. Expected to find a Header, but found: {:?}", self)
+        }
```
zslayton, line 243: For all of these, you could do:
```suggestion
        let SchemaContent::Header(value) = self else {
            unreachable!("expected to find a Header, but found: {:?}", self)
        };
        value
```

ion-schema/src/isl/isl_type.rs
```diff
@@ -101,8 +99,8 @@ impl IslType {
         }
     }
 
-    pub fn name(&self) -> &Option<String> {
-        &self.name
+    pub fn name(&self) -> Option<&String> {
+        self.name.as_ref()
```
zslayton, line 103: If it's a trivial change, it'd be nice for this to return `Option<&str>` instead.

ion-schema/src/isl/mod.rs
```diff
@@ -707,26 +889,25 @@ mod isl_tests {
...
         // initiate an Ion pretty text writer
-        let mut buffer = Vec::new();
-        let mut writer = TextWriterBuilder::pretty().build(&mut buffer).unwrap();
+        let buffer = Vec::new();
+        let config = WriteConfig::<v1_0::Text>::new(TextFormat::Pretty);
+        let mut writer = Writer::new(config, buffer)?;
 
         // write the previously constructed ISL model into a schema file using `write_to`
-        let write_schema_result = expected_schema.write_to(&mut writer);
+        let write_schema_result = expected_schema.write_as_ion(&mut writer);
```
zslayton, line 900: This works, though it can also be written `writer.write(expected_schema)`.

ion-schema/src/isl/ranges.rs
```diff
@@ -38,44 +38,63 @@ pub enum Limit<T> {
...
             Limit::Inclusive(value) => value.fmt(f),
             Limit::Exclusive(value) => write!(f, "exclusive::{value}"),
         }
     }
 }
 
+impl<T: WriteAsIon> WriteAsIon for Limit<T> {
+    fn write_as_ion<V: ValueWriter>(&self, writer: V) -> IonResult<()> {
+        match self {
+            // TODO: change Unbounded to Min and Max
```
zslayton, line 89: Outdated?

ion-schema/src/isl/ranges.rs
```diff
@@ -379,50 +337,22 @@ mod base {
...
-                        Limit::Exclusive(value) => {
-                            writer.set_annotations(["exclusive"]);
-                            writer.write_element(&value.clone().into())?;
-                        }
-                    }
-                    writer.step_out()?;
-                    Ok(())
-                }
+                Limit::Inclusive(value) if self.lower == self.upper => value.fmt(f),
+                _ => f.write_fmt(format_args!("range::[{},{}]", self.lower, self.upper)),
```
zslayton, line 344: This is my first time seeing a manual invocation of `format_args!` in the wild. 🤔

ion-schema/src/types.rs
```diff
@@ -131,7 +131,7 @@ impl BuiltInTypeDefinition {
         let mut constraints = vec![];
 
         // parses an isl_type to a TypeDefinition
-        let type_name = isl_type.name();
+        let type_name = isl_type.name().map(|x| x.to_owned());
```
zslayton, line 134: TIL that `Option<&str>` doesn't implement `ToOwned`.

zslayton (approved):
🚀

ion-schema/src/ion_extension.rs
zslayton: Tracking this as https://github.com/amazon-ion/ion-rust/issues/840.

ion-schema/src/isl/mod.rs
popematt: It can't be written that way because `IslSchema` does not implement `WriteAsIon`. This is because `WriteAsIon` implementations are allowed to write a single value, but an `IslSchema` is a sequence of top-level values.

PR #222, opened 2024-10-14
Makes IonSchemaElement use borrowed Elements

### LINKED ISSUE #137: `ion_schema` should not re-export `ion_rs`

### LINKED ISSUE #214: Validate API should borrow the data it is validating
Right now,the API to validate some data takes ownership of the data. It really should be immutably borrowing the `Element` to avoid unnecessary cloning. 

This is a big usability issue, and IMO, this issue must be resolved before the 1.0.0 release.

popematt:
### Issue #, if available:

Fixes #214 
Fixes #137

### Description of changes:

Changes `IonSchemaElement` so that it holds references to `Element`s rather than owning `Element`s. Details in PR tour 🗺️  comments.

----

### DIFF
--- a/ion-schema-tests-runner/src/generator.rs
+++ b/ion-schema-tests-runner/src/generator.rs
@@ -321,18 +321,15 @@ fn generate_preamble(root_dir_path: &Path) -> TokenStream {
 
         /// Asserts that a value is or is not valid for a given ISL type.
         fn __assert_value_validity_for_type(value_ion: &str, schema_id: &str, type_id: &str, expect_valid: bool) -> Result<(), String> {
-            let schema = __new_schema_system().load_schema(schema_id).unwrap();
-            let isl_type = schema.get_type(type_id).unwrap();
-            let value: ion_rs::Element = ion_rs::Element::read_one(value_ion.as_bytes()).unwrap();
+            let schema = __new_schema_system().load_schema(schema_id).expect(&format!("Expected to load schema: {}", schema_id));
+            let isl_type = schema.get_type(type_id).expect(&format!("Expected to get type: {}", type_id));
+            let value: ion_rs::Element = ion_rs::Element::read_one(value_ion.as_bytes()).expect(&format!("Expected to be able to read value: {}", value_ion));
             let prepared_value: ion_schema::IonSchemaElement = if value.annotations().contains("document") && value.ion_type() == ion_rs::IonType::SExp {
-                let element_vec = value.as_sequence()
-                    .unwrap_or_else(|| unreachable!("We already confirmed that this is a s-expression."))
-                    .elements()
-                    .map(|it| it.to_owned())
-                    .collect::<Vec<_>>();
-                ion_schema::IonSchemaElement::Document(element_vec)
+                let values = value.as_sequence()
+                    .expect("We already confirmed that this is a s-expression.");
+                ion_schema::IonSchemaElement::from(ion_schema::AsDocumentHint::as_document(values))
             } else {
-                ion_schema::IonSchemaElement::SingleElement(value)
+                ion_schema::IonSchemaElement::from(&value)
             };
             let validation_result = isl_type.validate(prepared_value);
             if validation_result.is_ok() == expect_valid {
--- a/ion-schema/src/nfa.rs
+++ b/ion-schema/src/nfa.rs
@@ -6,7 +6,6 @@ use crate::IonSchemaElement;
 use ion_rs::Element;
 use std::collections::{HashMap, HashSet};
 use std::iter::Peekable;
-use std::slice::Iter;
 use std::sync::Arc;
 
 /// Represents an id for a state in NFA
@@ -156,8 +155,11 @@ impl NfaEvaluation {
     }
 
     /// Validates provided ordered elements against referenced [Nfa]
-    pub fn validate_ordered_elements(&mut self, elements: &[Element], type_store: &TypeStore) {
-        let mut elements_iter = elements.iter().peekable();
+    pub fn validate_ordered_elements<'a, T: Iterator<Item = &'a Element>>(
+        &mut self,
+        mut elements_iter: Peekable<T>,
+        type_store: &TypeStore,
+    ) {
         // given elements are actually events for the `Nfa` referenced in this `NfaEvaluation`.
         // iterate through all elements and update state-visit count(`NfaRun`) for all possible transitions for given element(event).
         while let Some(element) = elements_iter.next() {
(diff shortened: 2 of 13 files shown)


ion-schema/src/nfa.rs
```diff
@@ -156,8 +155,11 @@ impl NfaEvaluation {
     }
 
     /// Validates provided ordered elements against referenced [Nfa]
-    pub fn validate_ordered_elements(&mut self, elements: &[Element], type_store: &TypeStore) {
-        let mut elements_iter = elements.iter().peekable();
+    pub fn validate_ordered_elements<'a, T: Iterator<Item = &'a Element>>(
```
popematt, line 158: 🗺️ The NFA implementation now accepts any iterator over `Element`s, rather than requiring a slice of `Element`s or being tied to a particular iterator implementation.

ion-schema/src/system.rs
```diff
@@ -437,16 +438,13 @@ impl PendingTypes {
...
+static DERIVED_ISL_TYPES: [&str; 9] = [
     "type::{ name: lob, one_of: [ blob, clob ] }",
     "type::{ name: number, one_of: [ decimal, float, int ] }",
     "type::{ name: text, one_of: [ string, symbol ] }",
     "type::{ name: $lob, one_of: [ $blob, $clob ] }",
     "type::{ name: $number, one_of: [ $decimal, $float, $int ] }",
     "type::{ name: $text, one_of: [ $string, $symbol ] }",
-    // this is just a place holder for document type,
-    // IonSchemaElement::Document(_) type is used to verify the correctness on the validation side
-    "type::{ name: document }",
```
popematt, line 449: 🗺️ `document` is an "atomic" type instead of a "derived" type now that we have `IonSchemaElementType::Document`.

ion-schema-tests-runner/src/generator.rs
```diff
@@ -321,18 +321,15 @@ fn generate_preamble(root_dir_path: &Path) -> TokenStream {
 
         /// Asserts that a value is or is not valid for a given ISL type.
         fn __assert_value_validity_for_type(value_ion: &str, schema_id: &str, type_id: &str, expect_valid: bool) -> Result<(), String> {
-            let schema = __new_schema_system().load_schema(schema_id).unwrap();
-            let isl_type = schema.get_type(type_id).unwrap();
-            let value: ion_rs::Element = ion_rs::Element::read_one(value_ion.as_bytes()).unwrap();
+            let schema = __new_schema_system().load_schema(schema_id).expect(&format!("Expected to load schema: {}", schema_id));
+            let isl_type = schema.get_type(type_id).expect(&format!("Expected to get type: {}", type_id));
+            let value: ion_rs::Element = ion_rs::Element::read_one(value_ion.as_bytes()).expect(&format!("Expected to be able to read value: {}", value_ion));
```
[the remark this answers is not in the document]
zslayton: I'm not very familiar with this part of the code--it looks like this is a method that only gets run in testing, so performance may not matter. If so, you can disregard.

These `expect(&format!(...))` calls will allocate/format a `String` even if the previous method calls succeed. We should use something like `Result::unwrap_or_else(|| panic!("Expected ..."))` to sidestep the cost on the happy path.

https://doc.rust-lang.org/std/result/enum.Result.html#method.unwrap_or_else

ion-schema-tests-runner/src/generator.rs
```diff
@@ -321,18 +321,15 @@ fn generate_preamble(root_dir_path: &Path) -> TokenStream {
 
         /// Asserts that a value is or is not valid for a given ISL type.
         fn __assert_value_validity_for_type(value_ion: &str, schema_id: &str, type_id: &str, expect_valid: bool) -> Result<(), String> {
```
zslayton, line 323: Do you happen to know why we're using a `__` prefix rather than, say, making a new submodule to house these functions?

ion-schema/src/nfa.rs
```diff
@@ -156,8 +155,11 @@ impl NfaEvaluation {
...
+        &mut self,
+        mut elements_iter: Peekable<T>,
```
zslayton, line 160: TIL that `Peekable`'s impl of `peekable()` is not specialized to avoid producing a `Peekable<Peekable<T>>`.

ion-schema/src/nfa.rs
```diff
@@ -267,12 +269,12 @@ impl NfaEvaluation {
     // this method iterates through elements to satisfy minimum required occurrences for given transition
     // It will return false if an invalid Ion value is found which doesn't satisfy minimum occurrence requirement for given transition
     // Otherwise it will return true
-    fn evaluate_transition_to_self(
+    fn evaluate_transition_to_self<'a, T: Iterator<Item = &'a Element>>(
```
zslayton, line 272: Food for thought: you could create a new trait for this to avoid having to spell out the bounds each time.
```rust
// Make an empty extension trait called `ElementIter`...
trait ElementIter<'a>: Iterator<Item = &'a Element> {}
impl<'a, T> ElementIter<'a> for T where T: Iterator<Item = &'a Element> {}

// ...and then replace this:
fn evaluate_transition_to_self<'a, T: Iterator<Item = &'a Element>> (...) {...}

// ...with this:
fn evaluate_transition_to_self<'a, T: ElementIter<'a>>(...) {...}
```

zslayton (approved):

ion-schema-tests-runner/src/generator.rs
popematt: I wanted to make sure that we didn't have any name conflicts with a generated test case, so I added a `__` prefix. I could have added a new submodule, but that probably also would have needed something to reduce the likelihood of a name conflict with a generated test function.

ion-schema-tests-runner/src/generator.rs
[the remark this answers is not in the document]
popematt: It's only in testing.

ion-schema/src/nfa.rs
popematt: Yeah, I guess I could. TBH, though, I'm hoping to do a fairly complete rewrite the NFA to get better error messages like in https://github.com/amazon-ion/ion-schema-kotlin/pull/228

PR #225, opened 2024-10-25
Fixes ordered_elements implementation

### LINKED ISSUE #56: Implementation of a `StateMachine` for validation
As part of the validation of `ordered_elements` constraint, it would require to implement a State Machine which can keep track of `occurs` constraint and move to another state as per events(e.g. `type:int`) .  

**Example:**
For a constraint `ordered_elements: [int, decimal]`, State Machine constructs an NFA like below:
```
 InitialState --event1--> IntermediateState --event2--> FinalState
 S1:required --type:int--> S2:required --type:decimal--> S3:optional
```
*Note::By default the `occurs` constraint would be `required` meaning `range::[1,1]`*

Given below is a summary tale representing the states and events for a value to be validated:

 value: [1, 2.0]
 
| Event            | State set (each state represented by state:visitCount)|
| -------------|------------------------------------------------------|
|                      | (initial state) |
| type:int         | S1:1 |
| type:decimal     | S2:1 --> matches! |

value: [2.0, 1]

 | Event             | State set (each state represented by state:visitCount)|
 |--------------|-----------------------------------------------------|
 |                       | (initial state)                                     |
 | type:int          | S1:0 --> no matches! expected int found decimal value

Prerequisite PR: #55 (`contains` implementation is required to be used within `occurs` constraint while `ordered_elements` constraint validation)

### LINKED ISSUE #105: implemnattion of `Display` for `Violation`
Implement a `Display` for `Violation` which could represent all the violations in a nested struct form. 
(Adding the type name for which the violation occurred should help direct to the root cause)

e.g.
```
Schema:
type::{
  name: my_int,
  type: int,
}
```
```
Input(value to be validated):
[1, 2]
```
```
{ 
   message: "value didn't satisfy type constraint(s)", 
   constraint: my_int, 
   violations: [ 
       { 
          message: "expected type Integer, found List", 
          constraint: type_constraint, 
          violations: [  ], 
          code: type_mismatched 
       } 
   ], 
   code: type_constraints_unsatisfied 
}
```

### LINKED ISSUE #221: Issue with `ordered_elements` constraint when first element has a min occurrence >= 1
Currently `ordered_elements` constraint uses NFA to evaluate if the given elements are valid or not. It is not taking into account the first element that has a minimum occurrence >=1. It instead allows even the invalid values to be valid for that first element. This bug is in the NFA evaluator and might have to do with the first transition that is to the state itself.

Here's an example through the sandbox, 
For valid values it works fine and gives the result as valid, example [here](https://amazon-ion.github.io/ion-schema/sandbox?schema=type%3A%3A%7B%0A%20%20%20%20name%3A%20ordered_elements_type%2C%0A%20%20%20%20ordered_elements%3A%20%5B%7Btype%3A%20symbol%2C%20occurs%3A%20required%7D%2C%20%7Btype%3A%20int%2C%20occurs%3A%20range%3A%3A%5B1%2C%203%5D%7D%2C%20%7Btype%3A%20bool%2C%20occurs%3A%20required%7D%5D%2C%0A%7D&value=(%3F%201%20true)&type=ordered_elements_type).
But for it even results into "valid" for invalid first element:, example [here](https://amazon-ion.github.io/ion-schema/sandbox?schema=type%3A%3A%7B%0A%20%20%20%20name%3A%20ordered_elements_type%2C%0A%20%20%20%20ordered_elements%3A%20%5B%7Btype%3A%20symbol%2C%20occurs%3A%20required%7D%2C%20%7Btype%3A%20int%2C%20occurs%3A%20range%3A%3A%5B1%2C%203%5D%7D%2C%20%7Btype%3A%20bool%2C%20occurs%3A%20required%7D%5D%2C%0A%7D&value=(5%201%20true)&type=ordered_elements_type).

popematt:
### Issue #, if available:

Fix #56
Fix #221 
Fix #105

### Description of changes:

Complete rewrite of the state machine for ordered elements. This fixes all the correctness issues with `ordered_elements`; now all of the `ion-schema-tests` cases for `ordered_elements` are enabled and passing. It also brings improved violation messaging for the `ordered_elements` constraint—although not quite as nice as https://github.com/amazon-ion/ion-schema-kotlin/pull/228 because the source ISL isn't available by the time we're constructing the ordered elements NFA, and I didn't want to tackle that as part of this PR since I think the violations may end up being refactored before v1.0.0 anyway.

That being said, I did go and update the implementation of `Display` for `Violation` because it was printing stuff on one line in a completely unusable way. The new implementation is a little bit more like `ion-schema-kotlin` in that it prints a multi-line tree of violations.

----


ion-schema/src/ordered_elements_nfa.rs
```diff
@@ -0,0 +1,443 @@
...
+            let mut invalid_transitions: HashSet<TraversalError> = HashSet::new();
+
+            ion_path.push(IonPathElement::Index(input_index));
+
+            // For each state in the set of current states...
+            for &(from_state_id, num_visits) in &current_state_set {
+                let from_state = self.states.get(from_state_id).unwrap();
+
+                // Need to clone the range because strangely &RangeInclusive doesn't
+                // implement Copy or IntoIterator.
```
zslayton, line 170: Happily, this weirdness is [getting addressed in edition 2024](https://github.com/rust-lang/rfcs/blob/master/text/3550-new-range.md).

ion-schema/src/ordered_elements_nfa.rs
```diff
@@ -0,0 +1,443 @@
...
+                let edges = self.edges.get(from_state_id).unwrap().clone();
+
+                // For each edge out of the current state we are inspecting...
+                for to_state_id in edges {
+                    let to_state: &State = self.states.get(to_state_id).unwrap();
```
zslayton, line 175: Here and above, you could use `[]` for concision instead of `get(...).unwrap()`.

ion-schema/src/violation.rs
```diff
@@ -86,12 +86,30 @@ impl Violation {
...
+        f.write_str(self.message.as_str())?;
+
+        let mut stack = vec![];
+        let mut violations_iter = self.violations.iter();
+        let mut violation = violations_iter.next();
+
+        while let Some(v) = violation {
+            f.write_fmt(format_args!(
+                "\n{}- {}",
+                "  ".repeat(stack.len() + 1),
```
zslayton, line 100: Consider adding a `TODO` for a reusable indentation `String` rather than re-allocating/formatting one in each iteration.

zslayton (commented):
Looks good! My comments are mostly comment clarifications.

ion-schema/src/ordered_elements_nfa.rs
```diff
@@ -0,0 +1,443 @@
...
+                }
+                j += 1;
+            }
+            edges.push(min_transition..=j)
+        }
+
+        // We make `Final` have fake edges to `Initial`. `Initial` denies all entry and `Final`
+        // doesn't allow exiting, so it's not actually traversable, and besides, you can't enter
+        // `Final` unless you're at the end-of-sequence event, so there's no more input to handle
+        // that would make it even try any further transitions.
```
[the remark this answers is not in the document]
popematt: It turns out I don't actually need them. I added them because I'm used to thinking about the way arrays work in Java, and in that case it simplifies things to have a fake entry so that the `edges` and `states` have the same length. However, a `Vec::get` returns an `Option`, so I don't actually need them here.

ion-schema/src/ordered_elements_nfa.rs
```diff
@@ -0,0 +1,443 @@
...
+    ) -> ValidationResult {
+        let mut current_state_set: HashSet<StateVisitCount> = HashSet::new();
+        let mut new_states: HashSet<StateVisitCount> = HashSet::new();
+
+        let mut input_index = 0;
+        current_state_set.insert((0usize, 1usize));
+
+        // Essentially, for-each input, but we want to capture the `Option::None` at the end of the iterator.
+        loop {
+            let element: Event = iter.next();
```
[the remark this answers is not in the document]
popematt: Changing to `ElementOrSequenceEnd` to be very explicit about what is represented.

ion-schema/src/violation.rs
```diff
@@ -86,12 +86,37 @@ impl Violation {
...
-            write!(f, "  {v}")?;
+        f.write_str(self.message.as_str())?;
+
+        let mut stack = vec![];
+        let mut violations_iter = self.violations.iter();
+        let mut violation = violations_iter.next();
+
+        let one_indent = "  ";
+        // TODO: Consider extracting this into a static or thread local variable.
+        let mut indentation = vec![one_indent.to_string()];
```
zslayton, line 99: This works, but why have a `Vec` of indentation `String`s instead of a single `String` that you add spacing to as the stack gets deeper?

zslayton (approved):
One thing to consider RE: indentation before you merge.

ion-schema/src/violation.rs
popematt: Lol, it's because I've been using too much Java recently, and I forgot that strings can be mutated in Rust. I'll fix it.

Issue #206, opened 2024-02-07
Validation doesn’t produce violations in a deterministic order


jobarr-amzn:
As reported through another channel by an `ion-schema-rust` user:

```rust
#[test]
fn test_non_determinism() {
    use ion_rs::element::Element;
    use ion_schema::{
        system::SchemaSystem,
        isl::{
            IslSchema,
            isl_constraint::v_1_0::fields,
            isl_type::v_1_0::named_type,
            isl_type_reference::v_1_0::{
                variably_occurring_type_ref,
                named_type_ref,
            },
            isl_range::Range
        }
    };

    let element = Element::read_one("{str: 123, number: \"fox\"}").unwrap();

    assert!((0..20).map(|_| {
        let struct_type = named_type("my_type", vec![fields([
            ("str".into(), variably_occurring_type_ref(named_type_ref("string"), Range::required())),
            ("number".into(), variably_occurring_type_ref(named_type_ref("int"), Range::required())),
        ].into_iter())]);
        let isl_schema = IslSchema::schema_v_1_0("my_schema", vec![], vec![struct_type], vec![], vec![]);
        let mut schema_system = SchemaSystem::new(vec![]);
        let schema = schema_system.load_schema_from_isl_schema_v1_0(isl_schema).unwrap();
        let type_definition = schema.get_type("my_type").unwrap();
        let violation = type_definition.validate(&element).unwrap_err();
        violation
            .flattened_violations()
            .into_iter()
            .cloned()
            // Uncomment to make this test pass
            // .sorted_by(|left, right| left.message().cmp(right.message()))
            .collect::<Vec<_>>()
    }).all_equal());
}
```

As far as I can tell this is because `HashMap` is used to store field constraints and its hash function is randomly seeded.

jobarr-amzn:
Version information for the above:
```toml
[[package]]
name = "ion-schema"
version = "0.9.1"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "f4b66b5bf004a664fce2936505d4fdced4e7e50f8e911c7306b9aa2d4c754eab"
dependencies = [
 "half",
 "ion-rs",
 "num-bigint 0.3.3",
 "num-traits",
 "regex 1.10.3",
 "thiserror",
]
```

popematt:
> As far as I can tell this is because HashMap is used to store field constraints and its hash function is randomly seeded.

[...]

Issue #187, opened 2023-06-19
Use version marker trait for ISL model


desaikd:
Current APIs for creating an ISL model allows mixing different versioned ISL constraints. When we try to resolve this ISL model by loading the schema(`load_schema()`) it returns error if the constraints are from different ISL versions. If we have something that returns a compile-time error while constructing the ISL model that gives an ease of use for the API.

We can create a marker trait for ISL versioning and have it implement for different ISL versions.
e.g.
```
trait IslVersion {} // ISL version marker trait

struct IslV1_0 {}
struct IslV2_0 {}
```
This can be used to define ISL model structs over the generic marker trait `IslVersion`. We can then have a particular version struct implementation for a constraint. This allows us to define constraints based on its version and gives a compile time error when a constraint doesn't exist for the given ISL version.

PR #230, opened 2025-02-19
Adds IslVersion marker and ViolationRecorder types

### LINKED ISSUE #229: Define `ViolationRecorder`, `IslVersion` traits and related types

popematt:
### Issue #, if available:

progress on #187, #206
resolves #229 

### Description of changes:

* Adds `IslVersion` trait, which is public, but cannot be implemented outside of this crate
* Adds empty types representing Ion Schema 1.0 and 2.0
* Adds `ViolationInfo` struct
* Adds `ViolationRecorder` trait with implementations for `Option<ViolationInfo>` and `Vec<ViolationInfo>`.

----

### DIFF
--- a/ion-schema/src/ion_schema_version.rs
+++ b/ion-schema/src/ion_schema_version.rs
@@ -1,0 +1,44 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+mod private {
+    /// Used to prevent a public trait from being implemented outside of this crate.
+    /// See https://predr.ag/blog/definitive-guide-to-sealed-traits-in-rust/
+    pub trait IslVersion {}
+}
+
+/// Trait for ISL version marker types.
+pub trait IslVersion: private::IslVersion {
+    const MAJOR_MINOR: (u8, u8);
+}
+
+/// Ion Schema Language 1.0
+#[allow(non_camel_case_types)]
+pub enum ISL_1_0 {}
+
+impl private::IslVersion for ISL_1_0 {}
+
+impl IslVersion for ISL_1_0 {
+    const MAJOR_MINOR: (u8, u8) = (1, 0);
+}
+
+/// Ion Schema Language 2.0
+#[allow(non_camel_case_types)]
+pub enum ISL_2_0 {}
+
+impl private::IslVersion for ISL_2_0 {}
+
+impl IslVersion for ISL_2_0 {
+    const MAJOR_MINOR: (u8, u8) = (2, 0);
+}
+
+/*
+When a new version is added, we can add pub(crate) traits to represent ranges of versions.
+
+/// Internal-only marker trait to make it simple to implement a trait once
+/// for things that are supported the same way in Isl 2.0 and 2.1.
+pub(crate) trait SinceISL_2_0: IslVersion { }
+
+impl SinceISL_2_0 for ISL_2_0 {}
+
+ */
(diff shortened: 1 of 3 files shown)


ion-schema/src/ion_schema_version.rs
```diff
@@ -0,0 +1,44 @@
...
+}
+
+/// Trait for ISL version marker types.
+pub trait IslVersion: private::IslVersion {
+    const MAJOR_MINOR: (u8, u8);
+}
+
+/// Ion Schema Language 1.0
+#[allow(non_camel_case_types)]
+pub enum ISL_1_0 {}
```
desaikd, line 17: I am not sure I understand the purpose for this enum. Why do we need an enum here, why not struct? Does this type represent the ISL model? It would be nice to add some comments on the purpose of these types or what they represent.

ion-schema/src/violation_recorder.rs
```diff
@@ -0,0 +1,176 @@
...
+pub struct ViolationInfo<'a> {
+    /// The constraint that was violated.
+    constraint: &'a Constraint,
+    /// The value that failed to satisfy the constraint.
+    value: IonSchemaElement<'a>,
+    /// A human-readable description of why the value violated the constraint.
+    message: String,
+    // TODO: Wire these in later
+    //   path: IonPath
+}
```
desaikd, line 71: Should we add a violation code here?

popematt: We could, but I'm not convinced of the utility of doing so. While it may be easier for a programmatic use case to interact with the violation code than it is to interact with the message, I'm not sure that the violation code provides enough information to be useful given that a reference to the constraint is also present in the `ViolationInfo`. For example, a violation code such as `InvalidValue` does not provide more useful information that simply checking that the violated constraint is a `valid_values` constraint. Another example—a `ContainerLengthTooLong` violation code would tell you exactly what the problem is, but you would still have to look at the constraint to see what is allowable range.

I am open to revisiting this later, though.

ion-schema/src/ion_schema_version.rs
popematt: This is a type that represents the Ion Schema Language version `1.0`. It's an enum with no variants so that you cannot create an instance of it—it's purpose is just to represent a specific ISL version in type signatures.

This will allow us to have functions such as `write_as_isl<V: IslVersion>(&self, writer: IonWriter) -> IonSchemaResult`, which can be called like this: `my_schema.write_as_isl::<ISL_2_0>(writer)?`, as well as enabling builder classes for the model that include the version in the type signature so that e.g. if you have a `TypeDefinitionBuilder<ISL_1_0>` you can add a `scale` constraint, but not an `exponent` constraint.

Here's a more complete example for the decimal-specific constraints.
```rust
// ISL 1.0
impl TypeDefinitionBuilder<ISL_1_0> {
    pub fn scale(mut self, range: Range<isize>) -> Self {
        let c = Scale { range } ;
        self.constraints.push(AnyConstraint::Scale(c));
        self
    }
}

// ISL 2.x
impl <V: SINCE_ISL_2_0> TypeDefinitionBuilder<V> {
    pub fn exponent(mut self, range: Range<isize>) -> Self {
        let c = Exponent { range } ;
        self.constraints.push(AnyConstraint::Exponent(c));
        self
    }
}

// All ISL Versions
impl <V: ISLVersion> TypeDefinitionBuilder<V> {
    pub fn precision(mut self, range: Range<usize>) -> Self {
        let c = Precision { range } ;
        self.constraints.push(AnyConstraint::Precision(c));
        self
    }
}
```

desaikd: Thanks for the detailed explanation. LGTM!

desaikd (approved):

Issue #228, opened 2025-01-17
Release v1.0.0 of `ion-schema-rust`


popematt:
As the Ion team and other users have used `ion-schema-rust` since its initial release in 2022, we have learned more about what customers need from `ion-schema-rust` and discovered bugs and other sharp edges.

In order to continue developing tools that depend on `ion-schema-rust` we need to have a stable release of the library. (If we don't, then there will eventually end up being a critical mass of things depending on the current state of `ion-schema-rust` and that will become the _de facto_ stable release.)

Items that need to be addressed:
* Tech Debt—there are several correctness issues that need to be fixed 
  * #209
  * #205
  * #224 
  * #199 
  * #197 
  * #44
* Correctness
  * #184 
  * #147 
  * #27 
* Architectural Challenges
  * The use of `IonSchemaSystem` makes certain combinations of operations impossible. You cannot load a schema and all its dependencies and then try to manipulate or inspect those schemas. It's impossible to load schemas, add a new type and then use for validation.
  * The codebase exhibits tight coupling between validation and type resolution processes
  * Schema manipulation capabilities are limited
  * The violation reporting system lacks flexibility
  * The maintenance of two parallel models introduces complexity, as changes often need to be implemented and synchronized across both representations.
* API Limitations
  * #206 
  * #138 
  * #187 
  * #207 
  * #212

popematt:
To address the architectural challenges, we will make the following changes from the current design.

### 1. Validation Output: ViolationRecorder Trait

Rather than maintaining a violation accumulator in the validation logic, we can decouple it with a `ViolationRecorder` trait that is essentially a visitor over all violations. The visitor function would be called once for each constraint/value combination that is violated. The trait implementation can be provided by the user, and the user can decide what to do when a violation is encountered. They could store as little or as much information about the violation as they choose, in whatever data structure they choose. The trait-based approach makes it easy to support fast-fail validation because it may return `Continue(())` to continue validation, or `Break(())` to end validation prematurely.

```rust
pub trait ViolationRecorder {
    /// Accepts a violation.
    fn accept(&mut self, violation: &ViolationInfo) -> ControlFlow<()>;
}


impl ResolvedType {
    pub fn validate<'a, V, R, F>(&self, value: V, recorder: F) -> Result<(), R>
        where V: AsRef<IonSchemaElement<'a>>, 
              R: ViolationRecorder,
              F: FnOnce() -> R
    {
[...]

PR #231, opened 2025-02-20
Adds stubs for TypeDefinitionBuilder, internal traits, and implementations for AnnotationsV2Simple

popematt:
### Issue #, if available:

#228 

### Description of changes:

* Adds definitions for internal traits with stubs of supporting structs
* Adds stub for `TypeDefinitionBuilder`
* Set up `AnyConstraint` and `AnyConstraintRef` enum
* Implements `annotations` for ISL 2.0 simplified syntax
* Updates `ViolationRecorder` to use `AnyConstraintRef`
* Updates `ion-rs` dependency to `1.0.0-rc.11`

----

### DIFF
--- a/ion-schema/src/internal_traits.rs
+++ b/ion-schema/src/internal_traits.rs
@@ -5,10 +5,10 @@
 //       instead of being clobbered together.
 
 use crate::result::IonSchemaResult;
+use crate::type_reference::TypeReference;
 use crate::{IonSchemaElement, IslVersion, ViolationRecorder};
 use ion_rs::{Element, ValueWriter};
 use std::ops::ControlFlow;
-use std::sync::Arc;
 
 /// For internal implementation of validation.
 ///
@@ -25,14 +25,24 @@ pub(crate) trait ValidateInternal {
     ) -> ControlFlow<()>
     where
         R: ViolationRecorder<'a>;
+}
 
-    /// Resolve any type references in preparation for validation.
-    fn resolve(&self, schema_store: &Arc<InvisibleSchemaStore>) -> IonSchemaResult<()>;
+/// Trait for visiting all the type references of things that could have type references.
+///
+/// This can be implemented for things with no type references, and if done so, the implementations
+/// of each function should be a no-op.
+trait HasTypeReferences {
+    /// Recursively visit each [TypeReference] held by self or by members of self.
+    ///
+    /// Used for:
+    ///   - Identifying references to other schemas that may need to be loaded
+    ///   - Ensuring that all references are resolvable
+    fn visit_type_references<F: FnMut(&TypeReference)>(&self, visitor: F);
 
-    /// Un-resolve all type references.
-    /// This is not strictly required, but if type reference resolution were to fail,
-    /// we should undo any changes before returning ownership to the user.
-    fn unresolve(&self);
+    /// Recursively visit each [TypeReference] held by self or by members of self.
+    ///
+    /// Used for resolving the type references.
+    fn visit_mut_type_references<F: FnMut(&mut TypeReference)>(&mut self, visitor: F);
 }
 
 // TODO: fields/functions to support
@@ -41,7 +51,7 @@ pub(crate) trait ValidateInternal {
 //  - "validate" configuration options
 pub(crate) struct ValidationContext {}
 
-pub(crate) struct InvisibleSchemaStore {
+pub(crate) struct SchemaStore {
     // TODO: This is a placeholder
 }
 
--- a/ion-schema/src/internal_traits.rs
+++ b/ion-schema/src/internal_traits.rs
@@ -1,0 +1,64 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// TODO: This file is a placeholder. These things will eventually be moved to more sensible locations
+//       instead of being clobbered together.
+
+use crate::result::IonSchemaResult;
+use crate::{IonSchemaElement, IslVersion, ViolationRecorder};
+use ion_rs::{Element, ValueWriter};
+use std::ops::ControlFlow;
+use std::sync::Arc;
+
+/// For internal implementation of validation.
+///
+/// Should be implemented by TypeDefinition, TypeArgument, etc. and all constraints.
+pub(crate) trait ValidateInternal {
+    /// Validate a value.
+    /// Panics if any unresolved type references are encountered. This crate must
+    /// ensure that resolve() is called before calling validate.
+    fn validate_internal<'a, R>(
+        &'a self,
+        value: &'a IonSchemaElement,
+        ctx: &ValidationContext,
+        recorder: &'a mut R,
+    ) -> ControlFlow<()>
+    where
+        R: ViolationRecorder<'a>;
+
+    /// Resolve any type references in preparation for validation.
+    fn resolve(&self, schema_store: &Arc<InvisibleSchemaStore>) -> IonSchemaResult<()>;
+
+    /// Un-resolve all type references.
+    /// This is not strictly required, but if type reference resolution were to fail,
+    /// we should undo any changes before returning ownership to the user.
+    fn unresolve(&self);
+}
+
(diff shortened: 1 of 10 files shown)


ion-schema/src/internal_traits.rs
```diff
@@ -0,0 +1,64 @@
...
+    where
+        R: ViolationRecorder<'a>;
+
+    /// Resolve any type references in preparation for validation.
+    fn resolve(&self, schema_store: &Arc<InvisibleSchemaStore>) -> IonSchemaResult<()>;
+
+    /// Un-resolve all type references.
+    /// This is not strictly required, but if type reference resolution were to fail,
+    /// we should undo any changes before returning ownership to the user.
+    fn unresolve(&self);
```
zslayton: This is fine for the purposes of this PR, I'm just thinking out loud.

* `resolve` and `unresolve` take `&self`. Can they make changes that way? Or is the plan to use interior mutability?
* Absent any details, I find myself wondering if this effect couldn't be achieved in a more straightforward way by making sure that type reference resolution succeeds before modifying the schema store.

ion-schema/src/internal_traits.rs
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// TODO: fields/functions to support
+//  - looking up references in the InvisibleSchemaStore
+//  - any other state or needed for validation
+//  - "validate" configuration options
+pub(crate) struct ValidationContext {}
+
+pub(crate) struct InvisibleSchemaStore {
```
zslayton: The `Invisible` prefix is accurate but perhaps redundant? All private implementation-detail types are invisible to users.

ion-schema/src/internal_traits.rs
```diff
@@ -0,0 +1,64 @@
...
+/// For internal implementation of serialization.
+///
+/// Implementations of `WriteAsIon` may delegate to this when possible.
+pub(crate) trait WriteAsIsl<V: IslVersion> {
+    fn write_as_isl<W: ValueWriter>(&self, writer: W) -> IonSchemaResult<()>;
+}
+
+/// For internal implementation of reading Ion Schema Language.
+pub(crate) trait ReadFromIsl<V: IslVersion>: Sized {
+    fn try_read(ion: &Element, ctx: &LoaderContext) -> IonSchemaResult<Self>;
```
zslayton, line 67: Again, thinking out loud.

I've been thinking about introducing an `IonElement` trait that is implemented by `Element`, `LazyElement`, and `LazyValue`. If we centered the schema API around that trait rather than `Element`, we'd get skip-scans and perf improvements.

ion-schema/src/model/constraints/annotations_v2_simple.rs
```diff
@@ -0,0 +1,260 @@
...
+        Ok(())
+    }
+
+    fn unresolve(&self) { /* no-op */
+    }
+}
+
+impl WriteAsIsl<ISL_2_0> for AnnotationsV2Simple {
+    fn write_as_isl<W: ValueWriter>(&self, writer: W) -> IonSchemaResult<()> {
+        let mut modifiers = vec![];
```
zslayton: This could be done with a `list_writer` instead of allocating a `Vec`.

ion-schema/src/model/constraints/any_constraint.rs
```diff
@@ -0,0 +1,99 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+use crate::model::constraints::*;
+
+macro_rules! any_of_these {
+    (#[$derives:meta] enum $super_:ident { $($name:ident,)+ }) => {
+        #[$derives]
```
zslayton, line 7: Cool!

zslayton (approved):

ion-schema/src/internal_traits.rs
popematt: I was hoping that you would have feedback on this. I was a bit handwave-y about this because I know it's possible, but I probably need to mess around with it to find the best way.

> Or is the plan to use interior mutability?

Yes.

> Absent any details, I find myself wondering if this effect couldn't be achieved in a more straightforward way by making sure that type reference resolution succeeds before modifying the schema store.

The general idea is that a schema store can be created and destroyed, but cannot be modified (almost). Any modification that happens is part of the `resolve(schema: Map<String, Schema>)` function that creates the schema store in the first place.

I could change this so that there's a `find_type_references(&self) -> Vec<&TypeReference>`, then we check that they're all resolvable, then we put everything into the schema store, and then finally call `resolve()`, which is infallible.

My original plan was (and I know there are some ownership/lifetime things to sort out, but I know that it's possible) to do something like this:

```rust
type SchemaResolutionError = (Map<String, Schema>, InvalidSchemaError);
type ResolvedSchemas = Map<String, ResolvedSchema>;

pub fn resolve(schemas: Map<String, Schema>) -> Result<ResolvedSchemas, SchemaResolutionError>  {
    let schema_store = SchemaStore::from(schemas);
    if let Err(e) = update_type_references(schema_store) {
        unresolve_schema_references(schema_store);
        return Err((schema_store.into(), e))
    };
    // Create a Map<String, ResolvedSchema> and return that
}

fn resolve_type_references(schema_store: &SchemaStore) -> Result<(), InvalidSchemaError) {
    for (_, schema) in schema_store.schemas() {
        for (_, type) in schema.types() {
            // recursively traverse the content of each type, using interior mutability to update the type references
            type.resolve(schema_store)?;
        }
    }
}

fn unresolve_type_references(schema_store: &SchemaStore) -> {
    for (_, schema) in schema_store.schemas() {
        for (_, type) in schema.types() {
            // recursively traverse the content of each type, using interior mutability to update the type references
            type.unresolve();
        }
    }
}
```

I'm open to suggestions, of course.

ion-schema/src/model/constraints/any_constraint.rs
popematt: I had an earlier version that would take an enum like this...

```rust
enum Foo {
    Bar { a: u8, b: bool },
    Baz { c: String, d: Vec<Element> },
}
```
...and generate structs for `Bar` and `Baz` with all of the relevant `derive`s and `getters`. It was really great until I realized that it was returning `&String` and `&Vec<_>` for things, and I'd get more idiomatic APIs if I returned `&str` and `&[_]` instead.

ion-schema/src/internal_traits.rs
desaikd: What you have here looks good to me. I also like the idea of having `find_type_references` here, it seems much simpler.

ion-schema/src/model/constraints/annotations_v2_simple.rs
```diff
@@ -0,0 +1,260 @@
...
+/// Represents the [annotations] constraint, simple syntax, in Ion Schema 2.0
+///
+/// [annotations]: https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#annotations
+#[derive(Debug, PartialEq, Clone)]
+pub struct AnnotationsV2Simple {
+    pub(crate) modifier: AnnotationsV2Modifier,
+    pub(crate) annotations: Vec<Symbol>,
+}
+
+impl AnnotationsV2Simple {
```
desaikd, line 38: Are we only thinking about having immutable constraints here right now? I was wondering if we should add something like `add_annotation` or `with_annotations` to push a new annotation to this constraint. And maybe even remove. All the APIs to modify the annotation constraint.

desaikd (approved):

ion-schema/src/internal_traits.rs
popematt: > we'd get skip-scans and perf improvements.

We might if we used it for `IonSchemaElement`. We won't really be able to use skip-scanning for parsing Ion Schema Language because there's nothing that we're going to skip.

ion-schema/src/model/constraints/annotations_v2_simple.rs
popematt: I've created https://github.com/amazon-ion/ion-schema-rust/issues/233

PR #234, opened 2025-02-28
Adds TypeDefinition, TypeArgument, TypeReference, and TypeConstraint

popematt:
### Issue #, if available:

#228 

### Description of changes:

* Adds `TypeReference`, `TypeDefinition`, `TypeArgument`, and `TypeConstraint`
* Updates `LoadContext` to be parameterized with an `IslVersion` so that we can avoid having lots of `<SomeClass as ReadFromIsl<V>>::try_read(...)`.
* Adds `WriteContext` with a parameterized `IslVersion` for the same reason
* Updates the lifetimes in the `ValidateInternal` trait in order to solve a problem with multiple borrows of a mutable reference when attempting to call `validate_internal` for all constraints within a `TypeDefinition`.
* Adds some helper functions to `ion_extension.rs` for `Struct` and `Symbol`.
* Moves `TypeDefinitionBuilder` to the same file as `TypeDefinition`. This aligns with other types, which also have their builders in the same file as the type that they build.
* Adds a `ConstraintName` trait so that it's easy to access the keyword for a constraint.
* Adds `impl ValidateInternal`, `impl WriteAsIsl`, and `fn constraint_keyword()` to `AnyConstraint`

----


ion-schema/src/internal_traits.rs
```diff
@@ -59,16 +60,51 @@ pub(crate) struct SchemaStore {
...
 pub(crate) trait WriteAsIsl<V: IslVersion> {
-    fn write_as_isl<W: ValueWriter>(&self, writer: W) -> IonSchemaResult<()>;
+    fn write_as_isl<W: ValueWriter>(
+        &self,
+        writer: W,
+        ctx: &WriteContext<V>,
+    ) -> IonSchemaResult<()> {
+        let (major, minor) = V::MAJOR_MINOR;
+        invalid_schema_error(format!(
+            "Cannot write to Ion Schema Language {major}.{minor}"
```
zslayton:
```suggestion
            "Cannot write {self:?} as Ion Schema Language {major}.{minor}"
```
Or maybe even more user-friendly: `{self:?} is not supported in Ion Schema Language {major}.{minor}`.

ion-schema/src/internal_traits.rs
```diff
@@ -59,16 +60,51 @@ pub(crate) struct SchemaStore {
...
 
 // TODO: fields/functions to support
 //    * looking up types/schemas that are already loaded
 //    * queueing up types/schemas that need to be loaded
 //    * any other context that we discover that we need.
-pub(crate) struct LoaderContext {}
+pub(crate) struct LoaderContext<V> {
+    version: PhantomData<V>,
+}
+impl<V> LoaderContext<V> {
```
zslayton: Eventually I'd expect this `impl` to constrain `V: IslVersion`. At the moment it allows creating a useless `LoaderContext`, but that's fine for now.

ion-schema/src/model/ranges.rs
```diff
@@ -0,0 +1,173 @@
...
+                })
+            }
+            (Bound::Included(x), Bound::Included(y)) if x > y => {
+                Err(IonSchemaError::InvalidSchemaError {
+                    description: "Invalid range".to_string(),
+                })
+            }
+            (min, max) => Ok(IonSchemaRange { min, max }),
+        }
+    }
```
zslayton: This `match` is very repetitive, but also very resistant to refactoring for brevity! 😅

While I don't think the arms/guards can be consolidated, you can defer the bodies in the error case:

```suggestion
    pub fn try_new(min: Bound<T>, max: Bound<T>) -> Result<Self, IonSchemaError> {
        match (min, max) {
            (Bound::Unbounded, Bound::Unbounded) => {},
            (Bound::Excluded(x), Bound::Included(y)) if x >= y => {},
            (Bound::Included(x), Bound::Excluded(y)) if x >= y => {},
            (Bound::Excluded(x), Bound::Excluded(y)) if x >= y => {},
            (Bound::Included(x), Bound::Included(y)) if x > y => {}
            (min, max) => return Ok(IonSchemaRange { min, max }),
        };
        Err(IonSchemaError::InvalidSchemaError {
            description: "Invalid range".to_string(),
        })
    }
```

zslayton (approved):

ion-schema/src/model/constraints/mod.rs
```diff
@@ -3,6 +3,12 @@
 
 mod annotations_v2_simple;
 mod any_constraint;
+mod type_constraint;
 
 pub use annotations_v2_simple::*;
 pub use any_constraint::*;
+pub use type_constraint::*;
+
+pub(crate) trait ConstraintName {
```
desaikd, line 12: (Suggestion) This is fine. but it might be simple to just add this as a `name()` function to the enum `AnyConstraint`?

desaikd (approved):
There are some clippy warning not related to this change. Otherwise LGTM!

ion-schema/src/model/constraints/mod.rs
popematt: It's already present (although it's called `constraint_keyword()` right now), and it actually uses this trait in its implementation. This bit in the any constraint macro:
```rust
            pub(crate) const fn constraint_keyword(&self) -> &'static str {
                match self {
                    $($super_::$name(_) => $name::CONSTRAINT_NAME,
                    )+
                }
            }
```
Expands to
```rust
            pub(crate) const fn constraint_keyword(&self) -> &'static str {
                match self {
                    AnyConstraint::TypeConstraint(_) => TypeConstraint::CONSTRAINT_NAME,
                    AnyConstraint::TimestampPrecision(_) => TimestampPrecision::CONSTRAINT_NAME,
                    AnyConstraint::ContainerLength(_) => ContainerLength::CONSTRAINT_NAME,
                    // etc.
                }
            }
```

PR #235, opened 2025-03-04
Adds ranges, utf8_byte_length, and timestamp_precision

popematt:
### Issue #, if available:

#228 

### Description of changes:

* Consolidates the `any_of_these` and `any_of_these_refs` macros (used for `AnyConstraint`) into one macro so that we only need one list of the constraints.
* Adds `ReadFromIsl` for `i64` and `usize` to support range types
* Adds `IonSchemaRange` that is somewhat compatible with the range types in the Rust standard library.
  * In order to avoid some ugly APIs, I've make the `IonSchemaRange` implement `From` instead of `TryFrom`. That means it's possible to programmatically construct an empty range (which is not valid ISL). However, parsing ISL files will still correctly raise an error for empty ranges. I'm willing to revisit that decision.
* Adds `TimestampPrecision` and `Utf8ByteLength` constraints.
* Adds some more extension functions for `Element`

----

### DIFF
--- a/ion-schema/src/model/constraints/timestamp_precision.rs
+++ b/ion-schema/src/model/constraints/timestamp_precision.rs
@@ -13,11 +13,7 @@ use crate::{IonSchemaElement, IslVersion, ViolationInfo, ViolationRecorder};
 use ion_rs::{Element, IonResult, TimestampPrecision as TSPrecision, ValueWriter, WriteAsIon};
 use std::fmt::{Debug, Display, Formatter};
 use std::ops::ControlFlow;
-
-// For brevity within this module.
 use TimestampPrecisionValue::*;
-// Associated constants still need to be qualified, even with a * import.
-use TimestampPrecisionValue as Tpv;
 
 /// Argument for the [TimestampPrecision] constraint.
 #[derive(Copy, Clone, Debug, PartialOrd, PartialEq)]
@@ -64,14 +60,14 @@ impl TimestampPrecisionValue {
 
     fn as_str(&self) -> &'static str {
         match self {
-            Year => Tpv::YEAR,
-            Month => Tpv::MONTH,
-            Day => Tpv::DAY,
-            Minute => Tpv::MINUTE,
-            Second => Tpv::SECOND,
-            Millisecond => Tpv::MILLIS,
-            Microsecond => Tpv::MICROS,
-            Nanosecond => Tpv::NANOS,
+            Year => TimestampPrecisionValue::YEAR,
+            Month => TimestampPrecisionValue::MONTH,
+            Day => TimestampPrecisionValue::DAY,
+            Minute => TimestampPrecisionValue::MINUTE,
+            Second => TimestampPrecisionValue::SECOND,
+            Millisecond => TimestampPrecisionValue::MILLIS,
+            Microsecond => TimestampPrecisionValue::MICROS,
+            Nanosecond => TimestampPrecisionValue::NANOS,
         }
     }
 }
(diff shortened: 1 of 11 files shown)


ion-schema/src/model/constraints/timestamp_precision.rs
```diff
@@ -0,0 +1,300 @@
...
+use crate::result::{invalid_schema_error, IonSchemaResult};
+use crate::{IonSchemaElement, IslVersion, ViolationInfo, ViolationRecorder};
+use ion_rs::{Element, IonResult, TimestampPrecision as TSPrecision, ValueWriter, WriteAsIon};
+use std::fmt::{Debug, Display, Formatter};
+use std::ops::ControlFlow;
+
+// For brevity within this module.
+use TimestampPrecisionValue::*;
+// Associated constants still need to be qualified, even with a * import.
+use TimestampPrecisionValue as Tpv;
```
zslayton: I'd be ok with either, but having both seems unusual?

ion-schema/src/model/ranges.rs
```diff
@@ -0,0 +1,347 @@
...
+    fn from(value: T) -> Self {
+        IonSchemaRange {
+            min: Included(value.clone()),
+            max: Included(value),
+        }
+    }
+}
+
+// Technically, the implementations this produces can construct empty ranges, which are not valid in
+// Ion Schema Language. This only affects programmatically constructed types. If we want to
```
zslayton: Given our re-read of the spec, should this be updated?

zslayton (approved):
Looks good. There are some clippy comments to address before merging.

ion-schema/src/model/constraints/timestamp_precision.rs
popematt: Well, I like being able to use the enum variants without qualification (which is why the first one is there). If we're getting rid of one of them, I'd probably get rid of the second one.

ion-schema/src/model/ranges.rs
popematt: I've created https://github.com/amazon-ion/ion-schema/issues/140. Once that is resolved, I think we can remove this note. (I'll also update this note to mention that issue.)

ion-schema/src/model/constraints/timestamp_precision.rs
```diff
@@ -0,0 +1,296 @@
...
+    #[case::year("month", TimestampPrecision::new_unchecked(Month))]
+    #[case::year("day", TimestampPrecision::new_unchecked(Day))]
+    #[case::year("minute", TimestampPrecision::new_unchecked(Minute))]
+    #[case::year("second", TimestampPrecision::new_unchecked(Second))]
+    #[case::year("millisecond", TimestampPrecision::new_unchecked(Millisecond))]
+    #[case::year("microsecond", TimestampPrecision::new_unchecked(Microsecond))]
+    #[case::year("nanosecond", TimestampPrecision::new_unchecked(Nanosecond))]
+    #[case::year("range::[min, minute]", TimestampPrecision::new_unchecked(..=Minute))]
+    #[case::year("range::[minute, max]", TimestampPrecision::new_unchecked(Minute..))]
+    fn timestamp_precision_try_read_ok(#[case] ion: &str, #[case] expected: TimestampPrecision) {
```
desaikd, line 290: We should probably add invalid read test cases as well to make sure it errors on invalid data.

ion-schema/src/model/ranges.rs
```diff
@@ -0,0 +1,345 @@
...
+            Ok(IonSchemaRange::from(value))
+        }
+    }
+}
+
+impl<T: RangeBoundType> From<&T> for IonSchemaRange<T> {
+    fn from(value: &T) -> Self {
+        IonSchemaRange {
+            min: Included(value.clone()),
+            max: Included(value.clone()),
```
desaikd, line 201:
```suggestion
            max: Included(value),
```

ion-schema/src/model/constraints/timestamp_precision.rs
```diff
@@ -0,0 +1,296 @@
...
+/// [ISL 1.0]: https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#timestamp_precision
+/// [ISL 2.0]: https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#timestamp_precision
+#[derive(Debug, PartialEq, Clone)]
+pub struct TimestampPrecision {
+    range: IonSchemaRange<i64>,
+}
+
+impl TimestampPrecision {
+    pub(crate) fn new<T: Into<IonSchemaRange<TimestampPrecisionValue>>>(range: T) -> Self {
+        let tsp_range: IonSchemaRange<TimestampPrecisionValue> = range.into();
```
desaikd, line 114: I think I am msising something here. Is the current implementation only supporting valid ranges being constructed that are passed here. Are we only allowing ranges with same start and end. (i..e range:[year, year]).

ion-schema/src/model/ranges.rs
```diff
@@ -0,0 +1,345 @@
...
+    };
+}
+from_range!(Range);
+from_range!(RangeTo);
+from_range!(RangeFrom);
+from_range!(RangeInclusive);
+from_range!(RangeToInclusive);
+
+#[cfg(test)]
+mod tests {
```
desaikd, line 234: I couldn't find tests of ranges with timestamp precision in the timestamp_precision constraint file. Maybe we should add those different ranges here to check for read and write.(probably something like range::[month, year]).

ion-schema/src/model/ranges.rs
popematt: This one needs to be clone because we have `&T` rather than `T`.

ion-schema/src/model/constraints/timestamp_precision.rs
popematt: > Are we only allowing ranges with same start and end

No, there are are `From<_>` implementations for `IonSchemaRange` [here](https://github.com/amazon-ion/ion-schema-rust/pull/235/files#diff-b1b4a4530b9151e7a5be3879ff8c98f5ae3f7701fd84d21f0f4a1345b399b2eaR216-R231) that make it possible to do something like `TimestampPrecision::new(Year..=Day)` (equivalent to `range::[year, day]`) or `builder.timestamp_precision(Second..)` (equivalent to `range::[second, max]`).

ion-schema/src/model/ranges.rs
popematt: Sure, I will add some test cases for that to `timestamp_precision.rs`.

ion-schema/src/model/constraints/timestamp_precision.rs
popematt: Will do.

PR #236, opened 2025-03-10
Adds ValidValues constraint

popematt:
### Issue #, if available:

#228 

### Description of changes:

* Adds `valid_values` data model, builder functions, and read and write implementations. `ValidateInternal` impl is `todo!()`.
* Removes some of the trait bounds from `IonSchemaRange` to make it easier to work with.
* Adds a field to `WriteContext` that sets whether ranges are allowed to be written out as a single value.
* Switches the `ion-rs` dependency to depend directly on the `main` branch of `amazon-ion/ion-rust` so that we can make small fixes in `ion-rust` and immediately use them without having to create a new release of `ion-rust`.
* Adds a test harness for `WriteAsIsl` and `ReadFromIsl` test cases.

See PR tour comments (🗺️) for more details.

----

### DIFF
--- a/ion-schema/src/internal_traits.rs
+++ b/ion-schema/src/internal_traits.rs
@@ -79,13 +79,18 @@ pub(crate) trait WriteAsIsl<V: IslVersion>: Debug {
 /// In the future, it may include things such as:
 ///  - options to allow lossy conversion between ISL versions
 ///  - stylistic choices for writing as Ion
+#[derive(Copy, Clone, Debug)]
 pub(crate) struct WriteContext<V> {
     version: PhantomData<V>,
+    /// Indicates whether ranges may be "minimized" into single values.
+    /// E.g.: writing `range::[1, 1]` as `1`.
+    pub(crate) minimize_ranges: bool,
 }
 impl<V: IslVersion> WriteContext<V> {
     pub fn new() -> Self {
         WriteContext {
             version: PhantomData::<V>,
+            minimize_ranges: true,
         }
     }
 }
@@ -107,6 +112,8 @@ macro_rules! read_from_isl {
 }
 read_from_isl!(usize, expect_usize);
 read_from_isl!(i64, expect_i64);
+read_from_isl!(ion_rs::Decimal, expect_decimal);
+read_from_isl!(ion_rs::Timestamp, expect_timestamp);
 
 // TODO: fields/functions to support
 //    * looking up types/schemas that are already loaded
--- a/ion-schema/src/ion_schema_version.rs
+++ b/ion-schema/src/ion_schema_version.rs
@@ -8,11 +8,12 @@ mod private {
 }
 
 /// Trait for ISL version marker types.
-pub trait IslVersion: private::IslVersion {
+pub trait IslVersion: private::IslVersion + Clone {
     const MAJOR_MINOR: (u8, u8);
 }
 
 /// Ion Schema Language 1.0
+#[derive(Copy, Clone, Debug)]
 #[allow(non_camel_case_types)]
 pub enum ISL_1_0 {}
 
@@ -23,6 +24,7 @@ impl IslVersion for ISL_1_0 {
 }
 
 /// Ion Schema Language 2.0
+#[derive(Copy, Clone, Debug)]
 #[allow(non_camel_case_types)]
 pub enum ISL_2_0 {}
 
--- a/ion-schema/src/model/constraints/timestamp_precision.rs
+++ b/ion-schema/src/model/constraints/timestamp_precision.rs
@@ -185,7 +185,7 @@ impl ValidateInternal for TimestampPrecision {
             return recorder.accept(ViolationInfo::new(
                 self.into(),
                 value.clone(),
-                format!("expected a timestamp precision of {range}; found {timestamp}",),
+                format!("expected a timestamp precision of {range:?}; found {timestamp:?}",),
             ));
         }
         ControlFlow::Continue(())
--- a/ion-schema/src/model/constraints/utf8_byte_length.rs
+++ b/ion-schema/src/model/constraints/utf8_byte_length.rs
@@ -71,7 +71,7 @@ impl ValidateInternal for Utf8ByteLength {
                 self.into(),
                 value.clone(),
                 format!(
-                    "expected a UTF-8 encoded byte length of {}; found {length}",
+                    "expected a UTF-8 encoded byte length of {:?}; found {length}",
                     self.range
                 ),
             ));
@@ -112,6 +112,16 @@ mod tests {
 
     #[test]
     fn test_builder() -> IonSchemaResult<()> {
+        let type_ = TypeDefinitionBuilder::<ISL_1_0>::new()
+            .utf8_byte_length(5)
+            .build();
+        assert_eq!(
+            type_.constraints().cloned().collect::<Vec<_>>(),
+            vec![AnyConstraint::Utf8ByteLength(Utf8ByteLength::new(
+                IonSchemaRange::try_new(Bound::Included(5), Bound::Included(5))?
+            ))]
+        );
+
         let type_ = TypeDefinitionBuilder::<ISL_1_0>::new()
             .utf8_byte_length(1..10)
             .build();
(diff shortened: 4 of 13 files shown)


ion-schema/src/internal_traits.rs
```diff
@@ -79,13 +79,18 @@ pub(crate) trait WriteAsIsl<V: IslVersion>: Debug {
 /// In the future, it may include things such as:
 ///  - options to allow lossy conversion between ISL versions
 ///  - stylistic choices for writing as Ion
+#[derive(Copy, Clone, Debug)]
 pub(crate) struct WriteContext<V> {
     version: PhantomData<V>,
+    /// Indicates whether ranges may be "minimized" into single values.
+    /// E.g.: writing `range::[1, 1]` as `1`.
+    pub(crate) minimize_ranges: bool,
```
popematt, line 87: 🗺️ The field is `pub(crate)` since this struct isn't supposed to have any special logic to it.

ion-schema/src/ion_schema_version.rs
```diff
@@ -8,11 +8,12 @@ mod private {
 }
 
 /// Trait for ISL version marker types.
-pub trait IslVersion: private::IslVersion {
+pub trait IslVersion: private::IslVersion + Clone {
```
popematt, line 11: 🗺️ For some reason, `PhantomData<T>` doesn't implement `Clone` unless `T` implements `Clone`, even though `PhantomData` is a zero-sized type (so it should always be trivially `Copy` and `Clone`). So now we have a `Clone` requirement for `IslVersion`.

ion-schema/src/model/constraints/timestamp_precision.rs
```diff
@@ -185,7 +185,7 @@ impl ValidateInternal for TimestampPrecision {
             return recorder.accept(ViolationInfo::new(
                 self.into(),
                 value.clone(),
-                format!("expected a timestamp precision of {range}; found {timestamp}",),
+                format!("expected a timestamp precision of {range:?}; found {timestamp:?}",),
```
popematt, line 188: 🗺️ Changed to use the `Debug` representation instead of `Display` as part of dropping `Display` from the trait bounds of `IonSchemaRange`.

ion-schema/src/model/constraints/utf8_byte_length.rs
```diff
@@ -112,6 +112,16 @@ mod tests {
 
     #[test]
     fn test_builder() -> IonSchemaResult<()> {
+        let type_ = TypeDefinitionBuilder::<ISL_1_0>::new()
```
popematt, line 115: 🗺️ Just a drive-by test improvement.

ion-schema/src/model/constraints/valid_values.rs
```diff
@@ -0,0 +1,363 @@
...
+        values: I,
+    ) -> Self {
+        let values = values.into_iter().map(|it| it.into()).collect();
+        let constraint = ValidValues { values };
+        self.constraints.push(constraint.into());
+        self
+    }
+
+    /// Adds a `valid_values` constraint with a single range argument to this type definition.
+    pub fn valid_number_range<
```
popematt, line 48: 🗺️ I think that restricting the range of a number is a common enough use case to justify having a specialized function for it.

ion-schema/src/model/constraints/valid_values.rs
```diff
@@ -0,0 +1,363 @@
...
+            }
+            .into()]
+        );
+    }
+
+    #[test]
+    fn test_builder_valid_number_range() {
+        let type_ = TypeDefinitionBuilder::<ISL_1_0>::new()
+            .valid_number_range(1..=10)
+            .build();
```
popematt, line 259: 🗺️ This is why I thought that the `valid_number_range` would be a useful function.

ion-schema/src/model/constraints/valid_values.rs
```diff
@@ -0,0 +1,363 @@
...
+            .into()]
+        );
+    }
+
+    test_harness!(write_as_isl:
+        ISL_1_0, ISL_2_0 {
+            #[case::empty(
+                Ok("[]"),
+                ValidValues { values: vec![] }
+            )]
```
popematt, line 276: 🗺️ Here is a test case for the test harness. I've modeled it after the `#[case]` syntax from `rstest` so that it looks familiar, but if you think it could be easier to read with some other syntax, I can easily change it.

I also considered `test_case_name => (arg_1, arg_2),` for the test cases, but I don't have a particular preference either way.

ion-schema/src/model/constraints/valid_values.rs
```diff
@@ -0,0 +1,363 @@
...
+            type_.constraints().cloned().collect::<Vec<_>>(),
+            vec![ValidValues {
+                values: vec![IonSchemaRange::from(Decimal::from(1)..=Decimal::from(10)).into()]
+            }
+            .into()]
+        );
+    }
+
+    test_harness!(write_as_isl:
+        ISL_1_0, ISL_2_0 {
```
popematt: 🗺️ This is the test harness in action. I didn't want to bike-shed with myself on the syntax, so it's pretty minimal. If you think that a few more keywords would make it more self-describing, I'm happy to add some. (For example, it could be more like `test WriteAsIsl for ISL_1_0, ISL_2_0 { ...` if that's preferable.

ion-schema/src/model/ranges.rs
```diff
@@ -75,7 +74,9 @@ impl<T: RangeBoundType> IonSchemaRange<T> {
         };
         is_above_min && is_below_max
     }
+}
 
+impl<T> IonSchemaRange<T> {
```
popematt, line 79: 🗺️ This bit confused me for a moment when I saw it as a diff on GitHub, so here's an explanation. I moved `fn map_bounds()` to a separate `impl` block that has less type bounds on `T` so that I could use it with `NumericRangeType` in `valid_values.rs`.

ion-schema/src/ion_schema_version.rs
zslayton: I think it's supposed to be part of `PhantomData<T>`'s contract. If a type contains a `PhantomData<T>`, it must behave like it contains a `T`.

ion-schema/src/model/ranges.rs
```diff
@@ -88,65 +89,31 @@ impl<T: RangeBoundType> IonSchemaRange<T> {
...
-
-impl<V: IslVersion, T: RangeBoundType> WriteAsIsl<V> for IonSchemaRange<T> {
+impl<V: IslVersion, T: RangeBoundType + WriteAsIon> WriteAsIsl<V> for IonSchemaRange<T> {
     fn write_as_isl<W: ValueWriter>(
         &self,
         writer: W,
         ctx: &WriteContext<V>,
     ) -> IonSchemaResult<()> {
-        if self.min == self.max {
+        if ctx.minimize_ranges && self.min == self.max {
```
zslayton, line 98: Does this need to be `.ion_eq` instead of `==`? A comment earlier in the diff explained that single values check for Ion equality while ranges use mathematical/number-line equality.

ion-schema/src/test_harness.rs
```diff
@@ -0,0 +1,113 @@
...
+    (__goto__ __loop__: $version:ident $(,$other_versions:ident)* { $(#[case::$case_name:ident($($args:expr),+)])+ }$(;$($tt:tt)*)?) => {
+        test_harness!(__goto__ __emitfn__: $($case_name, $version, $($args),+)+);
+        test_harness!(__goto__ __loop__: $($other_versions),* { $(#[case::$case_name($($args),+)])+ $(;$($tt)*)?} );
+    };
+    (__goto__ __loop__: { $(#[case::$case_name:ident($($args:expr),+)])+ }$(;$($tt:tt)*)?) => {
+        test_harness!(__goto__ __loop__: $($($tt)*)?);
+    };
+    (__goto__ __loop__:) => {
+        // No more groups
+    };
```
zslayton: This is a bit brain melting to read, at least outside an IDE. Maybe add some dev comments describing the cases?

zslayton (approved):

ion-schema/src/model/ranges.rs
popematt: No, this doesn't need to be `.ion_eq`. That comment you saw was in `impl WriteAsIsl for ValidValues`, and is only true for `valid_values`, whereas this trait implementation applies to other ranges, such as the `IonSchemaRange<usize>` (used in `utf8_byte_length`) and even `IonSchemaRange<TimestampPrecisionValue>` (and `TimestampPrecisionValue` doesn't implement `IonEq`).

PR #237, opened 2025-03-13
Adds models for AllOf, Fields, and OrderedElements

popematt:
### Issue #, if available:

#228 

### Description of changes:

Adds data models, builder methods, and stubbed trait implementations for `Fields`, `AllOf`, and `OrderedElements`

As I was working on that change, I ran into some issues that prompted some refactoring. There were things that were unergonomic and there was a circular type requirement that was causing infinite recursion in the type checker.

* Adds `Versioned`, a wrapper for other things that can pass the ISL version as part of the type.
* Updates `TypeArgument` significantly so that it's considerable more ergonomic to use, and gets rid of a bunch of unneeded code.
* Updates `TypeDefinitionBuilder` to be better encapsulated.
* Simplified `ReadFromIsl` for `TypeDefinition` by introducing a `ReadConstraint` trait, which can return `Ok(None)` if the constraint is not supported for a given ISL version.

----

### DIFF
--- a/ion-schema/src/ion_schema_version.rs
+++ b/ion-schema/src/ion_schema_version.rs
@@ -13,7 +13,7 @@ pub trait IslVersion: private::IslVersion + Clone {
 }
 
 /// Ion Schema Language 1.0
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone, Debug, PartialEq)]
 #[allow(non_camel_case_types)]
 pub enum ISL_1_0 {}
 
@@ -24,7 +24,7 @@ impl IslVersion for ISL_1_0 {
 }
 
 /// Ion Schema Language 2.0
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone, Debug, PartialEq)]
 #[allow(non_camel_case_types)]
 pub enum ISL_2_0 {}
 
@@ -44,3 +44,40 @@ pub(crate) trait SinceISL_2_0: IslVersion { }
 impl SinceISL_2_0 for ISL_2_0 {}
 
  */
+
+use std::marker::PhantomData;
+use std::ops::Deref;
+
+/// Wrapper for data that needs to be passed around with an ISL version—for example, builder methods
+/// for some constraints can accept `Versioned<TypeArgument>`s.
+///
+/// Technically, this is a smart pointer that (rather than managing the ownership or memory of the
+/// wrapped value) encodes Ion Schema version information along with the wrapped value.
+#[derive(Copy, Clone, PartialEq, Debug)]
+pub struct Versioned<T, V: IslVersion>(T, PhantomData<V>);
+impl<T, V: IslVersion> Versioned<T, V> {
+    /// Creates a new [`Versioned`] instance, wrapping the given value.
+    pub(crate) fn new(value: T) -> Self {
+        Self(value, PhantomData)
+    }
+
+    /// Maps the value contained by this [`Versioned`] instance.
+    pub(crate) fn map<U, F: FnOnce(T) -> U>(
+        this: Versioned<T, V>,
+        transform: F,
+    ) -> Versioned<U, V> {
+        Versioned(transform(this.0), this.1)
+    }
+
+    /// Consumes this [`Versioned`] instance, return the value contained within.
+    pub(crate) fn into_inner(this: Versioned<T, V>) -> T {
+        this.0
+    }
+}
+impl<T, V: IslVersion> Deref for Versioned<T, V> {
+    type Target = T;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
--- a/ion-schema/src/model/constraints/all_of.rs
+++ b/ion-schema/src/model/constraints/all_of.rs
@@ -1,0 +1,173 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::internal_traits::{
+    LoaderContext, ValidateInternal, ValidationContext, WriteAsIsl, WriteContext,
+};
+use crate::ion_schema_version::Versioned;
+use crate::model::constraints::{ConstraintName, ReadConstraint};
+use crate::model::type_argument::TypeArgument;
+use crate::model::{TypeDefinitionBuilder, VersionedTypeArgumentList};
+use crate::result::IonSchemaResult;
+use crate::{IonSchemaElement, IslVersion, ViolationRecorder};
+use ion_rs::{Element, ValueWriter};
+use std::ops::ControlFlow;
+
+impl ConstraintName for AllOf {
+    const CONSTRAINT_NAME: &'static str = "all_of";
+}
+
+/// Represents the `all_of` constraint (ref. [ISL 1.0], [ISL 2.0]).
+///
+/// [ISL 1.0]: https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#all_of
+/// [ISL 2.0]: https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#all_of
+#[derive(Debug, PartialEq, Clone)]
+pub struct AllOf {
+    type_arguments: Vec<TypeArgument>,
+}
+
+impl AllOf {
+    fn new(mut type_arguments: Vec<TypeArgument>) -> Self {
+        // Sorting the vec allows us to get Bag-like equality semantics.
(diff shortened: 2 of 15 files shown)


ion-schema/src/ion_schema_version.rs
```diff
@@ -44,3 +44,40 @@ pub(crate) trait SinceISL_2_0: IslVersion { }
...
  */
+
+use std::marker::PhantomData;
+use std::ops::Deref;
+
+/// Wrapper for data that needs to be passed around with an ISL version—for example, builder methods
+/// for some constraints can accept `Versioned<TypeArgument>`s.
+///
+/// Technically, this is a smart pointer that (rather than managing the ownership or memory of the
+/// wrapped value) encodes Ion Schema version information along with the wrapped value.
```
popematt, line 55: 🗺️ I'm not completely sure whether this should be an opaque type or a pseudo-smartpointer. I'm open to suggestions. I'm calling it a smart pointer for now because it implements `Deref`, so I also tried to follow the other best practices for functions on smart pointers.

ion-schema/src/model/constraints/all_of.rs
```diff
@@ -0,0 +1,173 @@
...
+    #[test]
+    fn test_builder() {
+        let type_ = TypeDefinitionBuilder::<ISL_1_0>::new()
+            .all_of((
+                // This is a heterogeneous "list" (really a tuple) of things that can become a type arg.
+                "foo_type",
+                TypeDefinitionBuilder::new().build(),
+                TypeReference::imported("bar.isl", "baz"),
+                "quux_type".or_null(true),
+            ))
```
popematt, line 100: 🗺️ I just want to point this out because I think it's cool that we were able to make heterogeneous argument lists work here.

ion-schema/src/model/constraints/annotations_v2_simple.rs
popematt, line 1: 🗺️ Changes here are just switching from `ReadFromIsl` -> `ReadConstraint` and switching from using `builder.constraint` (the field) to `builder.with_constraint()` (the new method). No actual logic has changed.

ion-schema/src/model/constraints/timestamp_precision.rs
popematt, line 1: 🗺️ Changes here are just switching from `ReadFromIsl` -> `ReadConstraint` and switching from using `builder.constraint` (the field) to `builder.with_constraint()` (the new method). No actual logic has changed.

ion-schema/src/model/constraints/type_constraint.rs
popematt, line 1: 🗺️ With one exception (noted below) changes here are just switching from `ReadFromIsl` -> `ReadConstraint` and switching from using `builder.constraint` (the field) to `builder.with_constraint()` (the new method). No actual logic has changed.

ion-schema/src/model/constraints/type_constraint.rs
```diff
@@ -42,10 +43,11 @@ impl ConstraintName for TypeConstraint {
 }
 
 impl<V: IslVersion> TypeDefinitionBuilder<V> {
-    pub fn type_constraint(mut self, type_argument: TypeArgument) -> Self {
-        let constraint = TypeConstraint { type_argument };
-        self.constraints.push(constraint.into());
-        self
+    pub fn type_constraint(self, type_argument: VersionedTypeArgument<V>) -> Self {
```
popematt, line 46: 🗺️ This is the significant change—`TypeArgument` has been replaced with `VersionedTypeArgument<V>`.

ion-schema/src/model/constraints/utf8_byte_length.rs
popematt, line 1: 🗺️ Changes here are just switching from `ReadFromIsl` -> `ReadConstraint` and switching from using `builder.constraint` (the field) to `builder.with_constraint()` (the new method). No actual logic has changed.

ion-schema/src/model/constraints/valid_values.rs
popematt, line 1: 🗺️ Changes here are just switching from `ReadFromIsl` -> `ReadConstraint` and switching from using `builder.constraint` (the field) to `builder.with_constraint()` (the new method). No actual logic has changed.

ion-schema/src/model/type_definition.rs
```diff
@@ -46,7 +49,8 @@ impl TypeDefinition {
 /// A version-safe builder for Ion Schema type definitions.
 #[derive(Debug, Clone)]
 pub struct TypeDefinitionBuilder<V: IslVersion> {
-    pub(crate) constraints: Vec<AnyConstraint>,
+    constraints: Vec<AnyConstraint>,
+    // FIXME: Should probably be (String, IonData<Element>)
     open_content: Vec<(Symbol, IonData<Element>)>,
```
popematt: 🗺️ I left a note here (rather than change it) because I'm not convinced enough yet to change it. Do you have any thoughts on what this should be?

ion-schema/src/model/type_definition.rs
```diff
@@ -60,11 +64,17 @@ impl<V: IslVersion> TypeDefinitionBuilder<V> {
...
+    /// This is pub(crate) so that users cannot add a combination of constraints that is not
+    /// valid for any ISL version.
+    pub(crate) fn with_constraint(mut self, constraint: AnyConstraint) -> Self {
+        self.constraints.push(constraint);
+        self
+    }
+
+    pub fn build(self) -> VersionedTypeDefinition<V> {
+        Versioned::new(TypeDefinition::new(self.constraints, self.open_content))
     }
```
popematt, line 77: 🗺️ Now that `build()` returns `VersionedTypeDefinition`, it can be `pub`!

ion-schema/src/model/constraints/all_of.rs
zslayton: This is cool! I want to steal this for `ion-rust`'s `ValueWriter`.

ion-schema/src/model/constraints/all_of.rs
```diff
@@ -0,0 +1,173 @@
...
+#[derive(Debug, PartialEq, Clone)]
+pub struct AllOf {
+    type_arguments: Vec<TypeArgument>,
+}
+
+impl AllOf {
+    fn new(mut type_arguments: Vec<TypeArgument>) -> Self {
+        // Sorting the vec allows us to get Bag-like equality semantics.
+        // TODO: Replace the vec with a bag or set, or come up with a less hacky way to sort the type arguments.
+        type_arguments.sort_by_cached_key(|item| format!("{item:?}"));
```
zslayton: Since you're already sorting the `Vec`, you could use something like [`slice::partition_dedup`](https://doc.rust-lang.org/std/primitive.slice.html#method.partition_dedup) to get from Bag to Set.

ion-schema/src/model/constraints/all_of.rs
```diff
@@ -0,0 +1,173 @@
...
+                .build(),
+        );
+        assert_ne!(
+            TypeDefinitionBuilder::<ISL_1_0>::new()
+                .all_of(("a", "b", "c", "a"))
+                .build(),
+            TypeDefinitionBuilder::<ISL_1_0>::new()
+                .all_of(("a", "b", "c", "b"))
+                .build(),
+        );
```
zslayton, line 172: Maybe add a comment highlighting that this is checking for Bag (not Set) equivalence?

ion-schema/src/model/constraints/fields.rs
```diff
@@ -0,0 +1,252 @@
...
+        let name = value.0.into();
+        let type_ = value.1.into();
+        Versioned::map(type_, |t| (name, t))
+    }
+}
+
+/// A list of [`VersionedFieldTypeArgument`]s.
+///
+/// This type can be constructed from `Vec<T>`, `[T; N]`, or `(T₀, T₁, ...)`,
+/// where `T` (or `Tₙ`) implements `Into<VersionedFieldTypeArgument<V>>`.
```
zslayton, line 60: 🎩 Fancy subscripts.

ion-schema/src/model/type_definition.rs
zslayton: Why would it be `String` rather than `Symbol`? (Presumably ISL allows `$0` field names as open content, even though: Yuck.)

zslayton (approved):
Thanks for calling out the mechanical changes. There are a few clippy warnings to look at before merging.

ion-schema/src/model/constraints/all_of.rs
popematt: Using a set would be possible, but it would be a compromise rather than the desired state.

PR #238, opened 2025-03-17
Adds data models and builder functions for remaining constraints

popematt:
### Description of changes:

Adds all the remaining constraints, except for the variants of `annotations` (which still needs a little bit of work). Much of this PR is copy/paste from other constraints with some renaming done. It's probably fine to skim over them—just double check that there are no copy/paste mistakes where things should have been renamed but weren't.
* `AnyOf` and `OneOf` are copied from `AllOf`
* `Not` is copied from `TypeConstraint`
* `Scale`, `Precision`, `Exponent`, `ContainerLength`, `CodepointLength`, and `ByteLength` are copied from `Utf8ByteLength`

These constraints are entirely new:
* `Regex`
* `TimestampOffset`
* `Contains`
* `Element` and `FieldNames`
* `Ieee754Float`

I also fixed the visibility of the fields in the `utf8_byte_length` and `type` constraints.
Finally, `convert_to_pattern` was made `pub(crate)` so that it could be reused in the `Regex` constraint.

There are some doc comments, but I plan to fill that out in a later PR to avoid blowing up the cope of this one.

----


popematt:
When I started self-reviewing it to write the PR tour comments, I hit upon a solution to one of the things that was bothering me. Don't bother reviewing this yet.

ion-schema/src/model/constraints/element.rs
```diff
@@ -0,0 +1,146 @@
...
+        &self.type_argument
+    }
+
+    pub fn distinct(&self) -> bool {
+        self.distinct
+    }
+}
+
+impl<V: IslVersion> TypeDefinitionBuilder<V> {
+    pub fn element(self, type_argument: VersionedTypeArgument<V>) -> Self {
```
popematt, line 48: 🗺️ I chose to make this method available for both ISL versions, so that customers can switch from ISL 1.0 to 2.0 without having to refactor any existing code that does programmatic construction. There's also `element_distinct` with accepts a `bool` flag indicating whether the elements should be distinct or not.

ion-schema/src/model/constraints/field_names.rs
```diff
@@ -0,0 +1,110 @@
...
+        &self.type_argument
+    }
+
+    pub fn distinct(&self) -> bool {
+        self.distinct
+    }
+}
+
+impl TypeDefinitionBuilder<ISL_2_0> {
+    pub fn field_names(self, type_argument: VersionedTypeArgument<ISL_2_0>) -> Self {
```
popematt, line 45: 🗺️ Like `element`, `field_names` has two builder methods, but in this case, both are only available for ISL 2.0.

ion-schema/src/constraint.rs
```diff
@@ -1140,7 +1140,7 @@ impl ConstraintValidator for ContainerLengthConstraint {
         let size = if let Some(element_iter) = value.as_sequence_iter() {
             element_iter.count()
         } else if let Some(strukt) = value.as_struct() {
-            strukt.fields().map(|(k, v)| v).count()
+            strukt.fields().count()
```
popematt: 🗺️ This is a clippy warning that keeps showing up. Since I'm touching this file, I figured I'd fix it now.

zslayton: Is this different from `strukt.len()`?

ion-schema/src/model/constraints/any_of.rs
```diff
@@ -0,0 +1,174 @@
...
+/// [ISL 2.0]: https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#any_of
+#[derive(Debug, PartialEq, Clone)]
+pub struct AnyOf {
+    type_arguments: Vec<TypeArgument>,
+}
+
+impl AnyOf {
+    fn new(mut type_arguments: Vec<TypeArgument>) -> Self {
+        // Sorting the vec allows us to get Bag-like equality semantics.
+        // TODO: Replace the vec with a HashBag or come up with a less hacky way to sort the type arguments.
```
zslayton, line 32: Since this is used in a few places, consider making a naive `Bag` type wrapping a sorted `Vec<TypeArgument>` that you can optimize later.

ion-schema/src/model/constraints/any_of.rs
```diff
@@ -0,0 +1,174 @@
...
+                ),
+                TypeArgument::from_type_ref(
+                    NullabilityModifier::Nothing,
+                    TypeReference::imported("bar.isl", "baz")
+                ),
+                TypeArgument::from_type_ref(
+                    NullabilityModifier::Nullable,
+                    TypeReference::from("quux_type")
+                ),
+            ],)
```
zslayton, line 122: These explicitly-constructed `vec!`s would be good candidates for the tuple-of-`Into<_>`s trick. (Not in this PR.)

ion-schema/src/model/constraints/ieee754_float.rs
```diff
@@ -0,0 +1,118 @@
...
+            None
+        };
+        use FloatingPointNumberFormat::*;
+        match text {
+            Some("binary16") => Ok(Binary16),
+            Some("binary32") => Ok(Binary32),
+            Some("binary64") => Ok(Binary64),
+            _ => invalid_schema_error(format!("Not a valid IEEE754 format symbol: {value}")),
+        }
+    }
```
zslayton, line 99: I was curious to see if the compiler would do something clever with this, like lifting out the comparison of the string prefix. To my surprise, [it did not](https://godbolt.org/z/Tvdf4sz9M).

image

ion-schema/src/model/constraints/ieee754_float.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+impl From<&FloatingPointNumberFormat> for Value {
```
zslayton: This type is `Copy`. Is there a reason you're implementing this for a reference? Or just being consistent with other types' impls?

ion-schema/src/model/constraints/regex.rs
```diff
@@ -0,0 +1,296 @@
...
+        self.regex.as_str()
+    }
+}
+
+impl<V: IslVersion> TypeDefinitionBuilder<V> {
+    /// Adds a `regex` constraint to the type being built.
+    ///
+    /// This will panic if the input is invalid. For a safe alternative, see [`Self::try_regex`].
+    pub fn regex(self, regex: impl IonSchemaRegexSource<V>) -> Self {
+        self.with_constraint(Versioned::into_inner(regex.build_regex().unwrap()).into())
```
zslayton: This could be `self.try_regex(regex).unwrap()`.

ion-schema/src/model/constraints/regex.rs
```diff
@@ -0,0 +1,296 @@
...
+                case_insensitive: true,
+            })]
+        );
+
+        let type_ = TypeDefinitionBuilder::<ISL_1_0>::new()
+            .regex(
+                "abc"
+                    .to_regex_builder()
+                    .case_insensitive(true)
+                    .multiline(true),
```
zslayton: Not important: you could have `IonSchemaRegexSource` provide its own `case_insensitive` and `multiline` methods that were implemented by calling `.to_regex_builder().METHOD()`. Then you could do:

```rust
.regex("abc".case_insensitive(true).multiline(true))
```

zslayton (approved):

ion-schema/src/constraint.rs
popematt: No. I'll change it.

ion-schema/src/model/constraints/any_of.rs
popematt: It could be, but I'm intentionally not using it here so that if there's something wrong with the tuple-of-`Into<_>`s logic, it won't silently pass these tests. In other words, I'm intentionally constructing these the verbose way in order to make the tests more thorough.

ion-schema/src/model/constraints/ieee754_float.rs
popematt: That is weird. Good catch!

ion-schema/src/model/constraints/regex.rs
popematt: I like this.

PR #240, opened 2025-03-18
Adds Bag type

popematt:
### Issue #, if available:

#228 
Suggested in https://github.com/amazon-ion/ion-schema-rust/pull/238#discussion_r2001290273

### Description of changes:

Adds a naive `Bag` implementation and switches `ValidValues`, `AnyOf`, `OneOf`, `AllOf`, and `TypeDefinition` to use rather than having them all sort their own `Vec` or have incorrect equality.

I tried to minimize the diff as best as I could. The interesting changes are entirely in `bag.rs`. The rest of the changes are switching `vec` -> `bag`, updating related comments, and updating some constructor signatures.

----

### DIFF
--- a/ion-schema/src/model/bag.rs
+++ b/ion-schema/src/model/bag.rs
@@ -40,7 +40,7 @@ impl<T: PartialEq + Debug> Bag<T> {
 }
 
 impl<T> Bag<T> {
-    /// Returns the number of elements in the vector, also referred to
+    /// Returns the number of elements in the bag, also referred to
     /// as its 'length'.
     pub fn len(&self) -> usize {
         self.items.len()
--- a/ion-schema/src/model/bag.rs
+++ b/ion-schema/src/model/bag.rs
@@ -1,0 +1,113 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::fmt::Debug;
+use std::slice;
+
+/// An (internal only) bag implementation.
+///
+/// This naive implementation is backed by a sorted vec to have the same equality semantics as a bag.
+/// However, this relies on having a total ordering that is consistent with `Eq`.
+///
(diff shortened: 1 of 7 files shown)


ion-schema/src/model/bag.rs
```diff
@@ -0,0 +1,113 @@
...
+/// However, this relies on having a total ordering that is consistent with `Eq`.
+///
+/// The current implementation sorts by the `Debug` representation of the value. This is not
+/// particularly efficient, but until we have a better total ordering we can use, it will do.
+///
+/// Because of the sortedness invariant, adding or removing items one at a time is very inefficient,
+/// and unsupported. Any modifications should be done by trivially converting to a `Vec`, applying
+/// the modifications, and then converting back to a `Bag`.
+///
+/// TODO: Replace the vec with a HashBag or come up with a less hacky way to sort the type arguments.
```
zslayton, line 19: Outdated?

ion-schema/src/model/bag.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// This is required to be performed after any add/remove operation in order to maintain the
+    /// invariant that the items are sorted.
+    fn sort_items(&mut self) {
+        self.items.sort_by_key(|item| format!("{item:?}"));
+    }
+}
+
+impl<T> Bag<T> {
+    /// Returns the number of elements in the vector, also referred to
+    /// as its 'length'.
```
zslayton:
```suggestion
    /// Returns the number of elements in the bag, also referred to
    /// as its 'length'.
```

zslayton (approved):
Some clippy warnings to look at before merging.

ion-schema/src/model/bag.rs
popematt: Not really. I'd like to come up with a less naive approach because using `Debug` as the sort key works by coincidence rather than by design.

PR #241, opened 2025-03-19
Fix 'annotations' to support multiple versions of the constraint

popematt:
### Issue #, if available:

#228 

### Description of changes:

I've changed the strategy for dealing with the `annotations` constraint. Instead of having 3 possible constraints, we now have one constraint which internally has 3 variants.

The tests in `violation_recorder.rs` used the `annotations` constraint; rather than updating them to use the new annotations constraint implementation, I switched them to use a simpler constraint.

I also added a `pub builder()` function to `TypeDefinition`.

----

### DIFF
--- a/ion-schema/src/model/constraints/annotations/mod.rs
+++ b/ion-schema/src/model/constraints/annotations/mod.rs
@@ -1,0 +1,126 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+mod v1;
+mod v2_simple;
+mod v2_standard;
+
+pub use v1::*;
+pub use v2_simple::*;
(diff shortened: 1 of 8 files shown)


desaikd (approved):
LGTM! I didn't see any methods that allow adjusting `required` on all the annotations. (in ISL terms using `required::` on annotations list). That maybe done as a separate PR as well and might be nice to just have TODO/issue for it)

popematt:
> LGTM! I didn't see any methods that allow adjusting required on all the annotations. (in ISL terms using required:: on annotations list). That maybe done as a separate PR as well and might be nice to just have TODO/issue for it)

I've chosen not to do that for now because you have to specify either `Optional` or `Required` for all of the annotations. It's easy enough to add a `required()` method later. It shouldn't be a breaking change, so I have no problem waiting until someone asks for it.

PR #242, opened 2025-03-19
Fixes all clippy warnings

popematt:
### Issue #, if available:

None

### Description of changes:

This fixes all of the clippy warnings in `ion-schema-rust`.
Most of them had to do with formatting in the rustdoc comments.
See PR comments for the others.

----

### DIFF
--- a/ion-schema/src/isl/mod.rs
+++ b/ion-schema/src/isl/mod.rs
@@ -3,15 +3,13 @@
 //! This module consists of three submodules that help constructing ISL types/constraint:
 //!
 //! * `isl_type` module represents a schema type [IslType] which converts given ion content in the schema file
-//! into an ISL types(not-yet-resolved types). It stores `IslConstraint`s defined within the given type.
-//!
+//!   into an ISL types(not-yet-resolved types). It stores `IslConstraint`s defined within the given type.
 //! * `isl_import` module represents a schema import [IslImport] which converts given ion content in the schema file
-//! into an ISL import. It stores schema id, an optional type that needs to be imported and an optional alias to that type.
-//!
+//!   into an ISL import. It stores schema id, an optional type that needs to be imported and an optional alias to that type.
 //! * `isl_constraint` module represents schema constraints [IslConstraint]
-//! which converts given ion content in the schema file into an ISL constraint(not-yet-resolved constraints).
-//!
+//!   which converts given ion content in the schema file into an ISL constraint(not-yet-resolved constraints).
 //! * `isl_type_reference` module provides a schema type reference.
+//!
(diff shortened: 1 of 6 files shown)


ion-schema/src/isl/mod.rs
```diff
@@ -555,6 +553,7 @@ impl<'a> Iterator for SchemaIterator<'a> {
 }
 
 #[cfg(test)]
+#[allow(unused_must_use)]
```
popematt, line 556: 🗺️ The combination of `rstest` and a test function that returns `Result` causes a warning for an unused `Result`. Unfortunately, trying to add this `allow` annotation to the test functions directly did not work (probably because it doesn't get picked up by rstest's macro), so I've applied it to this entire module.

ion-schema/src/isl/util.rs
```diff
@@ -209,9 +209,15 @@ impl ValidValue {
         if element.annotations().contains("range") {
             isl_require!(annotation.len() == 1 => "Unexpected annotation(s) on valid values argument: {element}")?;
             // Does it contain any timestamps
-            let has_timestamp = element.as_sequence().map_or(false, |s| {
```
popematt, line 212: 🗺️ Clippy complained about an unnecessary `map_or` call.

ion-schema/src/ordered_elements_nfa.rs
```diff
@@ -424,9 +424,9 @@ impl Display for State {
 /// The reason why a transition (or edge) in the state machine graph cannot be traversed.
 #[derive(Debug)]
 enum TraversalError {
-    CannotEnterState(StateId, Violation),
-    CannotExitState(StateId),
-    CannotReEnterState(StateId),
+    Entering(StateId, Violation),
+    Exiting(StateId),
+    ReEntering(StateId),
```
popematt, line 429: 🗺️ Clippy complained that the enum variants all had the same prefix, suggesting that it's an implied part of the variant name or should be part of the enum name. When I looked it up, it seemed like it would also complain about a commin suffix, so I shortened all of the names to be the gerund form of the attempted action rather than a complete description.

ion-schema/src/violation.rs
```diff
@@ -210,10 +210,9 @@ macro_rules! assert_equivalent_violations {
     };
 }
 
-#[macro_export]
-
 /// Equivalence for `Violation`s is not supported due to its tree structure of having children violations.
 /// This macro can be used for comparing if two violations are not equal and uses `flattened_violations` for the comparison.
+#[macro_export]
```
popematt, line 215: 🗺️ There shouldn't be empty lines between an attribute and the thing to which it applies.

nirosys (approved):

PR #243, opened 2025-03-24
Adds SchemaDocument and builder

popematt:
### Issue #, if available:

#228 

### Description of changes:

* Adds `SchemaDocument` and supporting types `SchemaItem`, `SchemaHeader`, `SchemaFooter`, and relevant builders.
* Drive-by improvements for the type argument list `Into` macros
* Make `impl Deref for Versioned` only available for test, and make `into_inner()` a method instead of just an associated function.
----

### DIFF
--- a/ion-schema/src/ion_schema_version.rs
+++ b/ion-schema/src/ion_schema_version.rs
@@ -46,7 +46,6 @@ impl SinceISL_2_0 for ISL_2_0 {}
  */
 
 use std::marker::PhantomData;
-use std::ops::Deref;
 
 /// Wrapper for data that needs to be passed around with an ISL version—for example, builder methods
 /// for some constraints can accept `Versioned<TypeArgument>`s.
@@ -70,10 +69,15 @@ impl<T, V: IslVersion> Versioned<T, V> {
     }
 
     /// Consumes this [`Versioned`] instance, return the value contained within.
-    pub(crate) fn into_inner(this: Versioned<T, V>) -> T {
-        this.0
+    pub(crate) fn into_inner(self) -> T {
+        self.0
     }
 }
+
+// For testing-convenience only, we make this wrapper non-opaque.
+#[cfg(test)]
+use std::ops::Deref;
+#[cfg(test)]
 impl<T, V: IslVersion> Deref for Versioned<T, V> {
     type Target = T;
 
--- a/ion-schema/src/model/constraints/fields.rs
+++ b/ion-schema/src/model/constraints/fields.rs
@@ -96,10 +96,10 @@ impl<const N: usize, V: IslVersion, T: Into<VersionedFieldTypeArgument<V>>> From
 /// https://rust-lang.github.io/rfcs/3086-macro-metavar-expr.html is stable.
 macro_rules! into_fields_arg_list {
     ($($t:ident $i:tt),+) => {
-        impl<V: IslVersion, $($t),+> From<( $($t),+ )> for VersionedFieldTypeArgumentList<V>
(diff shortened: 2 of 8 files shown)


ion-schema/src/model/constraints/fields.rs
popematt, line 1: 🗺️ This is a minor fix to the `into_field_args_list` that makes it work for one-element tuples.

ion-schema/src/model/variable_type_argument.rs
popematt, line 1: 🗺️ This is a minor fix to the `into_var_type_arg_list` that makes it work for one-element tuples.

ion-schema/src/model/type_argument.rs
popematt, line 1: 🗺️ This is a minor fix to the `into_type_arg_list` that makes it work for one-element tuples.

ion-schema/src/model/constraints/regex.rs
```diff
@@ -109,6 +109,7 @@ impl<V: IslVersion> IonSchemaRegexSource<V> for IonSchemaRegexBuilder<V> {
 
 impl<V: IslVersion> From<Versioned<Regex, V>> for IonSchemaRegexBuilder<V> {
     fn from(value: Versioned<Regex, V>) -> Self {
+        let value = value.into_inner();
```
popematt, line 112: 🗺️  Side-effect of the change to `Versioned`.

ion-schema/src/ion_schema_version.rs
```diff
@@ -70,10 +69,15 @@ impl<T, V: IslVersion> Versioned<T, V> {
...
 
     /// Consumes this [`Versioned`] instance, return the value contained within.
-    pub(crate) fn into_inner(this: Versioned<T, V>) -> T {
-        this.0
+    pub(crate) fn into_inner(self) -> T {
+        self.0
     }
 }
+
+// For testing-convenience only, we make this wrapper non-opaque.
```
popematt, line 77: 🗺️ Really, I kept this for `cfg(test)` so that I could keep the diff smaller.

ion-schema/src/ion_schema_version.rs
```diff
@@ -70,10 +69,15 @@ impl<T, V: IslVersion> Versioned<T, V> {
     }
 
     /// Consumes this [`Versioned`] instance, return the value contained within.
-    pub(crate) fn into_inner(this: Versioned<T, V>) -> T {
-        this.0
+    pub(crate) fn into_inner(self) -> T {
+        self.0
```
popematt, line 73: 🗺️ Now, instead of `Versioned::into_inner(value)`, you can call `value.into_inner()`.

ion-schema/src/model/schema.rs
```diff
@@ -0,0 +1,236 @@
...
+}
+
+/// Represents an Ion Schema document.
+///
+/// A schema is a collection of types that can be used to constrain the Ion data model.
+#[derive(Debug, Clone, PartialEq)]
+pub struct SchemaDocument {
+    /// The version in which this schema was created, but not necessarily the only version for which it is valid.
+    isl_version: (u8, u8),
+    items: Vec<SchemaItem>,
```
desaikd, line 41: Why is this a `Vec`? Wouldn't this mean we can add multiple headers and footers?

popematt: It's a `Vec` so that we can preserve the relative order of the schema items (particularly the open content). The `SchemaDocumentBuilder` uses the type-state pattern to ensure that no more than one header and one footer are present, and that they occur in the correct order relative to the type definitions.

Since users of the library can only create a `SchemaDocument` using the `SchemaDocumentBuilder` they cannot add multiple headers and footers.

desaikd (approved):

Issue #151, opened 2023-03-14
Make `SchemaSystem` thread-safe


sadderchris:
Right now, since `SchemaSystem` is `!Send` and `!Sync`, it can't be used from multiple threads concurrently or even passed between multiple threads.  If we want to validate an Ion document in a multithreaded environment, we need to instantiate a new `SchemaSystem` for each thread, which isn't great for performance.

Ideally, we should be able to load an ISL schema once and re-use it from multiple threads without having to (re)parse it every time.

popematt:
I think there is some overlap between this issue and #150.

This is pretty off-the-cuff, but I wonder if we can solve them both by decoupling the loading of a schema from the management/caching of the schemas.

Suppose we define two traits:
```rust
// The point is that this trait should be some sort of unification across things like Arc<Schema>,
// Box<Schema>, Rc<Schema>, etc. so that regardless of how the user chooses to handle the 
// memory management of the cached schemas, the other parts of the ISL implementation will still work.
trait SchemaRef: Clone {
    fn new(schema: Schema) -> Self;
    fn as_schema(self) -> &Schema
}

trait SchemaCache {
    type SchemaType: SchemaRef;

    fn get_or_put(&mut self, id: String, schema: fn() -> Schema) -> Self::SchemaType;
    fn get(&self, id: String) -> Option<Self::SchemaType>;
    fn invalidate(&mut self, id: String) -> IonSchemaResult<()>;
}
```

And we define a separate loader (function), such as this:
```rust
/// Loads the requested schema, along with any direct and transient imports
fn load_schema(id: String, authorities: &Vec<Authority>) -> IonSchemaResult<(Schema, Vec<Schema>)>;
```

Then with a bit of glue code, we can put together a `SchemaSystem` that wraps this loader function and a `SchemaCache`. Depending on the `SchemaCache` being used, the `SchemaSystem` could be `Send`/`Sync`, I think.

Obviously this idea needs some refinement if it's going to work, but it _might_ be a decent place to start.

PR #245, opened 2025-03-31
Adds resolve function, ResolvedSchema, and ResolvedType

popematt:
### Issue #, if available:

#228 

### Description of changes:

* The key piece is the `resolve` function to resolve type references.
* Also adds `ResolvedSchema`, `ResolvedType`.
* Incidental changes to `SchemaDocument` and `Import`.

See 🪧 comments for details.

----

### DIFF
--- a/ion-schema/src/model/mod.rs
+++ b/ion-schema/src/model/mod.rs
@@ -18,3 +18,4 @@ pub use schema_header::*;
 pub use type_argument::*;
 pub use type_definition::*;
 pub use type_reference::*;
+pub use variable_type_argument::*;
--- a/ion-schema/src/model/schema.rs
+++ b/ion-schema/src/model/schema.rs
@@ -84,12 +84,42 @@ impl SchemaDocument {
         self.items.iter()
     }
 
+    /// Iterate all types declared in this schema document
+    pub fn types(&self) -> impl Iterator<Item = (&str, &TypeDefinition)> {
+        self.items.iter().filter_map(|item| match item {
+            SchemaItem::Type(name, def) => Some((name.as_str(), def)),
+            _ => None,
+        })
+    }
+
+    /// Gets the [`SchemaHeader`], if this schema document has a header.
+    pub fn header(&self) -> Option<&SchemaHeader> {
+        self.items.iter().find_map(|item| {
+            if let SchemaItem::Header(header) = item {
+                Some(header)
+            } else {
+                None
+            }
+        })
+    }
+
     /// Get a type from the schema, by name.
     pub fn get_type(&self, name: &str) -> Option<&TypeDefinition> {
         self.items.iter().find_map(|item| match item {
             SchemaItem::Type(type_name, type_def) if name == type_name => Some(type_def),
             _ => None,
         })
+    }
+
+    /// Iterate all types declared in this schema definition
+    pub(crate) fn indexed_type_names(&self) -> impl Iterator<Item = (usize, &String)> {
+        self.items
+            .iter()
+            .enumerate()
+            .filter_map(|(idx, item)| match item {
+                SchemaItem::Type(name, _) => Some((idx, name)),
+                _ => None,
+            })
     }
 
     pub(crate) fn get_type_idx_by_name(&self, name: &str) -> Option<usize> {
--- a/ion-schema/src/model/schema.rs
+++ b/ion-schema/src/model/schema.rs
@@ -33,7 +33,7 @@ impl From<Element> for SchemaItem {
 }
 
 impl TypeRefWalker for SchemaItem {
-    fn walk<V: TypeRefVisitor>(&self, visitor: &mut V) {
+    fn walk<V: TypeRefVisitor>(&mut self, visitor: &mut V) {
         if let SchemaItem::Type(_, t) = self {
             t.walk(visitor)
         }
--- a/ion-schema/src/model/schema_header.rs
+++ b/ion-schema/src/model/schema_header.rs
@@ -64,6 +64,25 @@ impl Import {
             }),
         }
     }
+
+    /// The Schema ID that this [`Import`] is targeting.
+    pub fn schema_id(&self) -> &str {
+        &self.schema_id
+    }
+
+    /// The (optional) type that this [`Import`] is targeting.
+    pub fn type_name(&self) -> Option<&str> {
+        self.type_import
+            .as_ref()
+            .map(|type_import| type_import.type_name.as_str())
+    }
+
+    /// The (optional) alias for the type that this [`Import`] is targeting.
+    pub fn type_alias(&self) -> Option<&str> {
+        self.type_import
+            .as_ref()
+            .and_then(|type_import| type_import.alias.as_deref())
+    }
 }
 
 /// Builds a [`SchemaHeader`].
--- a/ion-schema/src/model/type_reference.rs
+++ b/ion-schema/src/model/type_reference.rs
@@ -18,7 +18,9 @@ pub struct TypeReference {
 }
 
 impl TypeReference {
-    fn new<S: Into<String>>(type_name: S) -> Self {
+    /// Creates a `TypeReference` for a type name that is defined in or imported to the current
+    /// [`SchemaDocument`].
+    fn local<S: Into<String>>(type_name: S) -> Self {
         TypeReference {
             schema_id: None,
             type_name: type_name.into(),
@@ -26,6 +28,7 @@ impl TypeReference {
         }
     }
 
+    /// Creates a `TypeReference` that is an inline import from another schema.
     pub fn imported<A: Into<String>, B: Into<String>>(schema_id: A, type_name: B) -> Self {
         TypeReference {
             schema_id: Some(schema_id.into()),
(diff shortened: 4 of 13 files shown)


ion-schema/src/model/mod.rs
```diff
@@ -18,3 +18,4 @@ pub use schema_header::*;
 pub use type_argument::*;
 pub use type_definition::*;
 pub use type_reference::*;
+pub use variable_type_argument::*;
```
popematt, line 21: 🪧 I forgot to add this in a prior PR.

ion-schema/src/model/schema.rs
popematt, line 1: 🪧 This just adds some new accessors for the members of `SchemaDocument`.

ion-schema/src/model/schema_header.rs
popematt, line 1: 🪧 Accessors for members of `Import`.

ion-schema/src/model/type_reference.rs
```diff
@@ -13,17 +13,24 @@ use ion_rs::{Element, StructWriter, Value, ValueWriter};
...
-    // Initially `None`, but changed to `Some(_)` when resolving the schemas
-    // TODO: resolved: Cell<Option<...>>,
+    // Not exposed in public API. Initially `None`, but changed to `Some(_)` when resolving the schemas
+    resolved_type_coordinates: Option<TypeCoordinates>,
 }
 
 impl TypeReference {
-    // TODO: Determine if this needs to be pub
-    pub(crate) fn imported<A: Into<String>, B: Into<String>>(schema_id: A, type_name: B) -> Self {
+    fn new<S: Into<String>>(type_name: S) -> Self {
```
popematt: 🪧 There's now a public constructor for local type references (instead of just using `"foo".into()` or something like that).

ion-schema/src/model/type_reference.rs
```diff
@@ -13,17 +13,24 @@ use ion_rs::{Element, StructWriter, Value, ValueWriter};
 pub struct TypeReference {
     schema_id: Option<String>,
     type_name: String,
-    // Not exposed in public API.
-    // Initially `None`, but changed to `Some(_)` when resolving the schemas
-    // TODO: resolved: Cell<Option<...>>,
+    // Not exposed in public API. Initially `None`, but changed to `Some(_)` when resolving the schemas
+    resolved_type_coordinates: Option<TypeCoordinates>,
```
popematt, line 17: 🪧 This is important for the `resolve` function. Getter and setter are below.

ion-schema/src/model/type_reference.rs
```diff
@@ -35,20 +42,18 @@ impl TypeReference {
...
+    }
 }
 
 impl<T: Into<String>> From<T> for TypeReference {
     fn from(value: T) -> Self {
-        TypeReference {
-            schema_id: None,
-            type_name: value.into(),
-        }
+        TypeReference::new(value)
```
popematt: 🪧 This change and all that follow are just switching out manual construction to using the constructor functions.

ion-schema/src/resolver/resolve_fn.rs
```diff
@@ -0,0 +1,563 @@
...
+                    })
+                    .copied()
+            };
+            match type_ref_coordinates {
+                Ok(tc) => {
+                    // SAFETY: This is safe because
+                    //  * All the data is fully owned within this function while this happens
+                    //  * We know there is no concurrent access to this
+                    //
+                    // we know that there is no concurrent access to the type references.
```
zslayton: Were these meant to be one sentence?

ion-schema/src/resolver/resolve_fn.rs
```diff
@@ -0,0 +1,563 @@
...
+                    //
+                    // This is necessary because
+                    //  * We can't just use `&mut TypeReference` because we run into conflicting borrows
+                    //    because the loop would have to iterate over mutable references, but the global
+                    //    and local coordinate maps require an immutable borrow at the same time in order
+                    //    to borrow the schema and type names.
```
zslayton: `ion-rust` has been bitten by the `&mut`-in-a-loop limitation so many times. It's the motivation for 90% of its `unsafe` code. I need Polonius to hurry up and get here. 😞

ion-schema/src/resolver/resolve_fn.rs
```diff
@@ -0,0 +1,563 @@
...
+        })
+        .collect();
+    Some(unresolved_schemas)
+}
+
+/// Cast an immutable reference to a mutable reference.
+#[allow(clippy::mut_from_ref)]
+unsafe fn as_mutable<T: ?Sized>(val: &T) -> &mut T {
+    (val as *const T as *mut T).as_mut().unwrap_unchecked()
+}
```
zslayton: This is undefined behavior. To do this without causing UB, you need to wrap the reference that you'll be manually toggling between `&` and `&mut` inside an [`UnsafeCell`](https://doc.rust-lang.org/std/cell/struct.UnsafeCell.html).

ion-schema/src/resolver/resolved_schema.rs
```diff
@@ -0,0 +1,122 @@
...
+
+    /// Consumes this [`ResolvedSchema`], and attempts to return the [`SchemaDocument`]s for this
+    /// schema and all of its dependencies.
+    ///
+    /// If all other [`ResolvedSchema`] that were resolved in the same batch have been dropped,
+    /// returns all the schemas that were used in resolving this schema. Otherwise, returns `None`.
+    ///
+    /// To get an owned copy of the [`SchemaDocument`] for just this schema, you can alternately
+    /// use `self.as_schema_document().clone()`.
+    pub fn unresolve(self) -> Option<HashMap<String, SchemaDocument>> {
```
zslayton: Before 1.0, consider making an opaque wrapper `SchemaMap` that lets you replace `HashMap` if you want to later.

zslayton (changes requested):

ion-schema/src/resolver/resolve_fn.rs
popematt: If I put it in an `UnsafeCell`, then types and schemas are no longer `Sync`, which would be undoing #151.

I think I can come up with an alternative solution.

ion-schema/src/model/schema_header.rs
```diff
@@ -64,6 +64,25 @@ impl Import {
...
         }
     }
+
+    /// The Schema ID that this [`Import`] is targeting.
+    pub fn schema_id(&self) -> &str {
+        &self.schema_id
+    }
+
+    /// The (optional) type that this [`Import`] is targeting.
+    pub fn type_name(&self) -> Option<&str> {
```
desaikd, line 74: (Question) Is this also used for inline imports or inline imports are still pending?

ion-schema/src/model/type_reference.rs
```diff
@@ -13,17 +13,24 @@ use ion_rs::{Element, StructWriter, Value, ValueWriter};
...
-    pub(crate) fn imported<A: Into<String>, B: Into<String>>(schema_id: A, type_name: B) -> Self {
+    fn new<S: Into<String>>(type_name: S) -> Self {
+        TypeReference {
+            schema_id: None,
+            type_name: type_name.into(),
+            resolved_type_coordinates: None,
+        }
+    }
+
+    pub fn imported<A: Into<String>, B: Into<String>>(schema_id: A, type_name: B) -> Self {
```
desaikd, line 32: (Suggestion) We should probably add a doc comment explaining what kinds of imports does this represent? header imports, inline imports?

ion-schema/src/model/type_reference.rs
```diff
@@ -35,20 +42,18 @@ impl TypeReference {
...
 
-    /// Returns an immutable borrow of the [TypeDefinition] to which this [TypeReference] refers.
-    pub fn get(&self) -> Option<&TypeDefinition> {
-        todo!()
+    pub(crate) fn type_coordinates(&self) -> Option<TypeCoordinates> {
+        self.resolved_type_coordinates
     }
 
-    // TODO: functions for setting the resolved type
+    pub(crate) fn resolve_type_coordinates(&mut self, type_coordinates: Option<TypeCoordinates>) {
```
desaikd: I think I am missing context on what this function is doing. IS this placing a resolved type? If so, why does it take an `Option`.

ion-schema/src/resolver/resolved_schema.rs
popematt: I've switched the output to `impl Iterator<Item=(String, SchemaDocument)>`

ion-schema/src/model/schema_header.rs
popematt: Inline imports are a kind of type reference, and are implemented there. This is only for header imports.

ion-schema/src/model/type_reference.rs
popematt: Changed it to `set_type_coordinates` so that it doesn't look confusing that it can accept `None`.

ion-schema/src/resolver/resolve_fn.rs
popematt: There's a lot of `clone()`-ing of schema and type names, but I have update this so that now it doesn't use unsafe code, and it doesn't require the use of any kind of `Cell`.

zslayton (approved):

PR #246, opened 2025-04-14
Refactor InvalidSchemaError to support multiple errors

popematt:
### Issue #, if available:

#228 

### Description of changes:

* Refactors the crate's `result` module to...
  * Support multiple errors in `InvalidSchemaError`
  * Support error location in `InvalidSchemaError`
  * More cleanly wrap `IoError`
  * Eliminate the need for a `format!` when calling `invalid_schema_error`.

This change looks big, but a lot of the diff is swapping various `invalid_schema_error` for the new `invalid_schema!` macro.

See comments in the individual files to guide your attention to the important parts. If a file does not have a  "❗"  comment, then the changes are just switching from `invalid_schema_error(format!(...`, `invalid_schema_error_raw(format!(...`, etc. to the new `invalid_schema!` macro (and consequent `cargo fmt` changes).

----


ion-schema/src/ion_schema_version.rs
```diff
@@ -49,9 +49,6 @@ use std::marker::PhantomData;
 
 /// Wrapper for data that needs to be passed around with an ISL version—for example, builder methods
 /// for some constraints can accept `Versioned<TypeArgument>`s.
-///
-/// Technically, this is a smart pointer that (rather than managing the ownership or memory of the
-/// wrapped value) encodes Ion Schema version information along with the wrapped value.
```
popematt, line 54: ❗ This was an out of date comment. `Versioned` is no longer a "smart pointer".

ion-schema/src/model/schema_header.rs
```diff
@@ -36,10 +38,16 @@ impl SchemaHeader {
...
 pub struct Import {
     schema_id: String,
     type_import: Option<TypeImport>,
+    source_location: IslSourceLocation,
+}
+impl PartialEq for Import {
+    fn eq(&self, other: &Self) -> bool {
+        self.schema_id == other.schema_id && self.type_import == other.type_import
+    }
 }
```
popematt, line 51: ❗ `source_location` is extra metadata, not part of the semantic meaning of an `Import`, so we need a custom implementation of `PartialEq` to exclude it.

ion-schema/src/model/type_definition.rs
```diff
@@ -17,10 +17,16 @@ use std::ops::ControlFlow;
...
+#[derive(Debug, Clone)]
 pub struct TypeDefinition {
     constraints: Bag<AnyConstraint>,
     open_content: Bag<(Symbol, IonData<Element>)>,
+    source_location: IslSourceLocation,
+}
+impl PartialEq for TypeDefinition {
+    fn eq(&self, other: &Self) -> bool {
+        self.constraints == other.constraints && self.open_content == other.open_content
+    }
```
popematt, line 29: ❗ As with `Import`, `source_location` is extra metadata in `TypeDefinition`, not actually part of the equality, so we need a custom implementation of `PartialEq`.

ion-schema/src/model/type_reference.rs
```diff
@@ -4,18 +4,24 @@
...
     type_name: String,
+    source_location: IslSourceLocation,
     // Not exposed in public API. Initially `None`, but changed to `Some(_)` when resolving the schemas
     resolved_type_coordinates: Option<TypeCoordinates>,
 }
+impl PartialEq for TypeReference {
+    fn eq(&self, other: &Self) -> bool {
+        self.schema_id == other.schema_id && self.type_name == other.type_name
+    }
+}
```
popematt, line 24: ❗ Again, `source_location` is extra metadata, not actually part of the equality, so we need a custom implementation of `PartialEq`. In this case, it also exposed that the `resolved_type_coordinates` are not part of the "equality" definition either—two type references should be equivalent if they refer to the same type, regardless of whether one or both have been resolved.

ion-schema/src/result/invalid_schema_error.rs
```diff
@@ -0,0 +1,165 @@
...
+        Err(IonSchemaError::InvalidSchemaError(value.build()))
+    }
+}
+
+#[derive(Debug, Error, PartialEq)]
+enum InvalidSchemaErrorQuantity {
+    One(InvalidSchemaDetails),
+    Many(BTreeSet<InvalidSchemaDetails>),
+}
+impl Display for InvalidSchemaErrorQuantity {
```
popematt, line 144: ❗ When there are multiple errors in an `InvalidSchemaError`, they are displayed like this:
```
foo/bar.isl:24:13 > range start must not be greater than range end
foo/bar.isl:105:22 > top level type missing name
foo/baz.isl:1:1 > unknown schema version 3.14
```

ion-schema/src/result/invalid_schema_error.rs
```diff
@@ -0,0 +1,165 @@
...
+                self.push_err(e);
+                None
+            }
+        }
+    }
+}
+
+#[derive(Debug, Error, PartialEq)]
+#[error(transparent)]
+pub struct InvalidSchemaError(InvalidSchemaErrorQuantity);
```
popematt, line 80: ❗ An `InvalidSchemaError` is now composed of one or more `InvalidSchemaDetails`. `InvalidSchemaErrorQuantity` has a `One` and `Many` variant so that we can avoid allocating a `BTreeSet` unless we know we're collecting multiple errors. Heap allocation for a collection is _not_ required for the common case of invoking the `invalid_schema!` macro (although it may still be required for some of the `String`s in `InvalidSchemaDetails`).

ion-schema/src/result/invalid_schema_error.rs
```diff
@@ -0,0 +1,165 @@
+use crate::result::{HasIslSourceLocation, IonSchemaError, IonSchemaResult, IslSourceLocation};
+use std::collections::BTreeSet;
+use std::fmt::{Display, Formatter, Write};
+use thiserror::Error;
+
+/// Collects multiple errors to report them together.
+///
+/// ```ignore
```
popematt, line 8: ❗ This is `ignore` because doc tests can only use things that are visible outside of the crate, and `InvalidSchemaErrorCollector` is only `pub(crate)`.

ion-schema/src/result/mod.rs
popematt, line 1: ❗ When reviewing this file, you should not view changes from all commits. Exclude the first commit of the PR (the commit that moves the file from `result.rs` to `mod/result.rs`) to see how the content of the `result` mod has actually changed.

ion-schema/src/result.rs
popematt, line 1: ❗ If you look at only the first commit of this PR, you will see that this file was moved to `result/mod.rs`.

ion-schema/src/result/mod.rs
```diff
@@ -3,62 +3,92 @@
...
+mod invalid_schema_error;
+mod isl_source_location;
+
+pub use invalid_schema_error::InvalidSchemaError;
+pub(crate) use isl_source_location::*;
+
 use crate::violation::Violation;
 use ion_rs::IonError;
+use std::convert::Infallible;
+use std::fmt::{Debug, Write};
```
popematt, line 15: I'll fix this before merging.
```suggestion
use std::fmt::Debug;
```

ion-schema/src/result/mod.rs
```diff
@@ -94,18 +108,29 @@ pub fn unresolvable_schema_error_raw<S: AsRef<str>>(description: S) -> IonSchema
...
+                .build()
+        ))
+    };
+}
+pub(crate) use invalid_schema;
+
 /// A macro that checks some condition required to be valid ISL.
 ///
 /// If invalid, returns an InvalidSchemaErr with the given error message.
-#[macro_export]
```
popematt: ❗ This internal-only macro was inappropriately being exported by the crate. I've replaced the `#[macro_export]` with a `pub(crate) use` instead.

ion-schema/src/result/invalid_schema_error.rs
```diff
@@ -0,0 +1,165 @@
...
+///         }
+///         // Do something with `ok_value`
+///     }
+///     // Transform the error collector into a result.
+///     errors.into_result_with(())
+/// }
+/// ```
+#[derive(Debug, Default)]
+pub(crate) struct InvalidSchemaErrorCollector {
+    errors: BTreeSet<InvalidSchemaDetails>,
```
desaikd, line 31: (Question) Why are we using a `BTreeSet` here. I don't think the errors collected has to be sorted unless I am missing something here.

ion-schema/src/result/invalid_schema_error.rs
desaikd: Love this new `Display` for error reporting!

desaikd (approved):

ion-schema/src/result/invalid_schema_error.rs
popematt: There are two different parts to this question:

> I don't think the errors collected has to be sorted

In order to have a better user experience, they are sorted by `schema_id`, `row`, and `column`. Even though reading the ISL may occur in order and errors would already be sorted, resolving the imports and type references is a separate step (and itself includes multiple steps) so errors are not sorted by default.

> Why are we using a BTreeSet here

Because I didn't think it was appropriate to allocate a vec to hold a sorted copy of the error details in the `Display` impl. The other option is to sort them as they are aggregated, for which the `BTreeSet` is the natural choice. This, however, is an implementation detail that can change if necessary.

PR #247, opened 2025-04-17
Adds support for multiple errors from the resolver

popematt:
### Issue #, if available:

#228 

### Description of changes:

Builds on the work in #246, making the `resolve` function able to report multiple problems in the schemas that are being resolved.

As an example, here are the errors reported in `multiple_errs_can_be_reported` test case:
```
my_schema.isl:5:9 > target type not found for type reference 'bar'
my_schema.isl:7:1 > type 'foo' is defined multiple times
my_schema.isl:9:9 > target type not found for type reference 'baz'
```

----

### DIFF
--- a/ion-schema/src/model/constraints/annotations/v2_simple.rs
+++ b/ion-schema/src/model/constraints/annotations/v2_simple.rs
@@ -111,9 +111,9 @@ impl TypeDefinitionBuilder<ISL_2_0> {
     ///
     /// # Arguments
     /// * `modifier` - An [`AnnotationsV2Modifier`] that specifies how the annotations should be applied
-    ///               (e.g., required, closed, or both)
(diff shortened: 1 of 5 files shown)


ion-schema/src/model/constraints/annotations/v2_simple.rs
```diff
@@ -111,9 +111,9 @@ impl TypeDefinitionBuilder<ISL_2_0> {
     ///
     /// # Arguments
     /// * `modifier` - An [`AnnotationsV2Modifier`] that specifies how the annotations should be applied
-    ///               (e.g., required, closed, or both)
```
popematt, line 114: 🪧 This was a clippy suggestion.

nirosys (approved):
Nice. I'm not super familiar with the codebase yet, but changes look good to me.

